Issue #37, opened 2013-07-23
Selenide could log every step in debug mode


asolntsev:
It is useful in case of any mysterious behavior.
For example, for line
```java
$(languageDrpdwn).shouldHave(text("Java"))
```

it should write a log message `Element <option>Javascript</option> contains text "Java"`.

### Option 1: from command line
` -Dselenide.debug=true`

### Option 2: by test annotation
```java
@Test @Debug
public void myTest() {...} 
```
both JUnit and TestNG could be supported

asolntsev:
A draft version of this functionality was released in Selenide 2.16

See usage example in [IntegrationTest](https://github.com/codeborne/selenide/blob/master/src/test/java/integration/IntegrationTest.java):

```java
  @Rule
  public TestRule prettyReportCreator = new PrettyReportCreator();
```

PR #141, opened 2014-10-22
Possibility to log important Selenide Actions/Subjects

kumarunster:
Hi, 

what would you say about this? I have added possibility to listen and process important selenide actions in form like event logs. Main motive is, that, sometimes we need to show to our "business" guys some reports about our UI-Tests, what they really do. Plain code is indeed very readable with Selenide, but, if for example the tests fails in Jenkins we need to show, some pretty reports, which steps are passed and wich not.

So, I think, instead of something like this:

``` java
logger.info("element by name 'password' should not be readonly");
$(By.name("password")).shouldNotBe(readonly);
```

it would be great, to bring selenides great DSL-like syntax to be self-explanatory for some kinds of reports.

With this extension its possible to get following results (please run LogEventTest to see it on the console):
### Pretty report for Some Test Case

| Element | Subject | Status |
| --- | --- | --- |
| Browser | open http://localhost:16968/page_with_selects_without_jquery.html | PASSED |
| By.name: username | should be attribute readonly | PASSED |
| By.name: password | should not be attribute readonly | PASSED |
| By.name: password | should be empty | PASSED |
| By.name: password | set value '123‘ | PASSED |
| By.selector: #login | click | PASSED |

To get such report we can specify a Selenide Event Listener, register it and with help of a Junit Test Rule we can printout the report in each format we like, because all the events are in our hands. Also we can make it as a inherent part of each CI test job, wich shows the test in a human readable format.

Unfortunately I have polluted AbstractSelenideElement with some ugly try catch blocks, but I don't really know how to add such kind of feature without AOP with byte code manipulation into such a framework.

So what do you think about it? Maybe its correlated with your issue #37?

Best regards!

### DIFF
--- a/src/test/java/com/codeborne/selenide/ex/ErrorMessagesTest.java
+++ b/src/test/java/com/codeborne/selenide/ex/ErrorMessagesTest.java
@@ -1,16 +1,18 @@
 package com.codeborne.selenide.ex;
 
-import com.codeborne.selenide.Configuration;
-import com.codeborne.selenide.impl.ScreenShotLaboratory;
+import static com.codeborne.selenide.Screenshots.*;
+import static java.io.File.*;
+import static org.junit.Assert.*;
+import static org.mockito.Mockito.*;
+
+import java.util.Locale;
+
 import org.junit.AfterClass;
 import org.junit.BeforeClass;
 import org.junit.Test;
 
-import static com.codeborne.selenide.Screenshots.screenshots;
-import static java.io.File.separatorChar;
-import static org.junit.Assert.assertEquals;
-import static org.mockito.Mockito.doReturn;
-import static org.mockito.Mockito.mock;
+import com.codeborne.selenide.Configuration;
+import com.codeborne.selenide.impl.ScreenShotLaboratory;
 
 public class ErrorMessagesTest {
 
@@ -29,6 +31,7 @@ public class ErrorMessagesTest {
 
   @Test
   public void formatsTimeoutToReadable() {
+    Locale.setDefault(Locale.UK);
     assertEquals("\nTimeout: 0 ms.", ErrorMessages.timeout(0));
     assertEquals("\nTimeout: 1 ms.", ErrorMessages.timeout(1));
     assertEquals("\nTimeout: 999 ms.", ErrorMessages.timeout(999));


asolntsev:
@kumarunster You pull request has been merged and released in Selenide 2.16

See usage example in [IntegrationTest](https://github.com/codeborne/selenide/blob/master/src/test/java/integration/IntegrationTest.java):

```java
  @Rule
  public TestRule prettyReportCreator = new PrettyReportCreator();
```

Waiting for your feedback and suggestions...

stlecho:
Where is this PrettyReport being created ?

kumarunster:
Hi stlecho,

look into [PrettyReportCreator](https://github.com/codeborne/selenide/blob/a7a6cc9203cac5ca42d4771919d1b9ac7b19fd2e/src/main/java/com/codeborne/selenide/logevents/PrettyReportCreator.java)

you can create your own listener (implements LogEventListener), and attach them to your tests via Junit Rule. than on finish you can write the logged events into the file with the format you need.

deepakarorawins:
Hi,
Can we have live logs on console with testng for each step like...
Entered user name as <abc>
Entered password as <1234>
Clicked on Login button


Could you please suggest?

asolntsev:
We already have `TextReport` that prints out such logs: https://selenide.org/documentation/reports.html#1-text-report

Though, it prints out only in the end of the test.
Did you mean to print them out during the test, not in the end?
But why?
No one should watch running tests by his eyes. Tests are meant to be run on CI server.

deepakarorawins:
Our management want to see live logs, one reason is our tests are too much
flaky and and long running. So we want to see some times live logs on
jenkins console.

kumarunster:
agree with @asolntsev.  But sometimes its hard to argue with "managers". if you really need that "eye tracking" feature then implement it. Check the Rule class pointed in the documentation, register your own report listener and do what you want with the events you get via the listener.

deepakarorawins:
@kumarunster, @asolntsev: Sorry, I forgot to mention that I mean to use live logs with selenide appium tests with **TestNG** and not Junit. I think we can utilize listener, however it would be great if we can get example code.

One dirty way I can think of using locators with Selenide and use our logger in our wrapped action class.

deepakarorawins:
Hi even at the end of test when we get Text Report with Selenide-TestNG, I see repeated log any idea on this that why "isDisplyed" printed multiple times for "btnClose", and "switchToggle" ?


**btnClose**
```
public AndroidInventoryScreen clickCloseButton() throws Exception {
        click($(AppiumBy.id("btnClose")), "X Button to close Edu. screen");
        return this;
    }
```

```
    public void click(WebElement me, String elementName) {
        String msg = MessageFormat.format("Clicked {0}", elementName);
        try {
            waitForVisibility(me);
            me.click();
            utils.log().info(msg);
            ExtentReport.getTest().log(Status.INFO, msg);
        } catch (StaleElementReferenceException sere) {
            sere.printStackTrace();
            utils.log().debug("*StaleElementReferenceException occurred re-trying action.");
            waitForVisibility(me);
            me.click();
            utils.log().info(msg);
            ExtentReport.getTest().log(Status.INFO, msg);
        } catch (Exception e) {
            utils.log().debug("Below is the stake trace: - "+ e);
        }
    }
```


**switchToggle**
```
public AndroidToolMgmtScreen clickSerialNumberBy(String serialNumber) {
        By byDisconnect = AppiumBy.xpath("//android.widget.TextView[@text='Disconnect']");
        $(byAttribute("text",   serialNumber)).scroll(AppiumScrollOptions.up()).click();
        waitUntilExpectedConditionBy(byDisconnect, "displayed", "true", 30);
        return AndroidToolMgmtScreen.getInstance();
    }
```

```
public boolean waitUntilExpectedConditionBy(By by, String attribute, String value, long maxTimeOut) {
        boolean success = false;
        String msg = MessageFormat.format("Waiting until {0} = {1} for element located {2}", attribute, value, by);
        String errMsg = MessageFormat.format("Failed while waiting Until expected condition: {0} = {1} for element located by {2}", attribute, value, by);
        WebDriverWait wait = new WebDriverWait(getDriver(), Duration.ofSeconds(maxTimeOut));
        try {
            utils.log().info(msg);
            ExtentReport.getTest().log(Status.INFO, msg);
            wait.until(ExpectedConditions.attributeToBe(by, attribute, value));
            success = true;
        } catch (StaleElementReferenceException sere) {
            sere.printStackTrace();
            utils.log().debug("*StaleElementReferenceException occurred re-trying action.");
            wait.until(ExpectedConditions.attributeToBe(by, attribute, value));
            success = true;
        } catch (TimeoutException te) {
            utils.log().info(errMsg);
            ExtentReport.getTest().log(Status.INFO, errMsg);
        } catch (Exception e) {
            e.printStackTrace();
        }
        return success;
    }
```

image

kumarunster:
sorry for my opinion here, but pls get familiar with Selenide's domain specific language and advantages using it.
in my opinion your code could be reduced to
$(whatever).shouldBe(visible).click()
$(somethingElse).should(appear|disappear)
thats all. Try to get to that minimalistic, understandable expressions and use Selenide's should expressions, which already have implicit waits inside. In your code you are heavy using Selenium native waits, which in general should be avoided. That waits are just loops.

https://selenide.org/documentation.html

deepakarorawins:
@kumarunster I 100% agree, and know that my code is clumsy and I am trying to fit Selenide into existing FW.
Why actually I tried "$(whatever).shouldBe(visible).click()" first but again softAssert was not working with this (and test terminates here only), then I experimented to just use up to $(whatever).shouldBe(visible) and soft assert worked.

Sorry for the messed experiments. However I appreciate Selenide's features and way of doing things, still first I have to use it in existing FW due to management /work pressure to keep running existing stuff, then gradually I can completely move to Selenide

So, $(whatever).shouldBe(visible).**click()** should this work with SoftAssertion or I have some mess which is failing it right away.

asolntsev:
@deepakarorawins 

### About soft asserts:
1. Yes, line `$(whatever).shouldBe(visible);` works with soft asserts. 
2. Yes, line `$(whatever).shouldBe(visible).click()` doesn't work with soft asserts. Because if this `click` fails, the following actions don't make any sense. All the following checks will be meaningless anyway, even if they don't fail.

### About multiple `isDisplay` in your reports
Yes, it's expected that you see `is displayed()` multiple times in the report. Because you call method `waitForVisibility()`, which obviously calls `element.isDisplayed()` many times in a loop. That's exactly what you see in the report. 

You can at least replace `waitForVisibility(me);` by `me.shouldBe(visible)`.

deepakarorawins:
Thank you very much @asolntsev  I will work on your suggestions.

PR #156, opened 2015-01-21
Add PdfTest and pdfbox library

previousdeveloper:
I have added pdfContentTest besides I have included additional library which is https://pdfbox.apache.org/


asolntsev:
@previousdeveloper Cool!
But actually this test does **not** test the Selenide code. It's more like an example of PDF test. Probably we should move it to http://github.com/selenide-examples/ project. Or we should create a function in Selenide that would test PDF. For example,

```java
$("#downloadSalaryPdf").shouldDownload(pdfWithText(
            "John Smith", "Salary", "November", "150 USD"));
```

What do you think?

previousdeveloper:
Yes I know ;) Probably you can move it to http://github.com/selenide-examples/ project. Thanks a lot .

PR #169, opened 2015-03-25
Check if Sizzle is defined

### LINKED ISSUE #168: Using Sizzle breaks javascript on page
If Sizzle is not included in page, window.onerror is triggered by `return Sizzle() != null` in com.codeborne.selenide.impl.WebElementSelector#sizzleLoaded

Resistance:
Fixes #168


asolntsev:
Thank you!

PR #172, opened 2015-03-31
Add method for WebDriver instance checking via WebDriverRunner

dimand58:
Check of existance of WebDriver instance is needed for some purposes:
-  screenshot creation in my case (if WD alive screenshot will be captured from browser window, else from system via Robot framework)

Without of this check browser is 'blinking', because Selenide trying to restore closed WebDriver (WD can be unexpectable closed after some error in browser/IEDriverServer/tests and etc)

### DIFF
--- a/src/main/java/com/codeborne/selenide/WebDriverRunner.java
+++ b/src/main/java/com/codeborne/selenide/WebDriverRunner.java
@@ -119,6 +119,14 @@ public class WebDriverRunner {
   public static void closeWebDriver() {
     webdriverContainer.closeWebDriver();
   }
+  
+  /**
+   * Is instance of Selenium WebDriver exist
+   * @return true if instance of Selenium WebDriver exist
+   */
+  public static boolean isDriverExist() {
+    return webdriverContainer.isDriverExist();
+  }
 
   /**
    * Is Selenide configured to use Firefox browser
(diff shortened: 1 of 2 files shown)


asolntsev:
Thank you for the pull request! Will merge it soon. Need to resolve few merge conflicts...

How are you going to use the new method `isDriverExist`? Probably it would be simpler just check screenshot for `null` and use Robot in this case?

dimand58:
Hi,
my method for screenshot capturing uses getWebDriver() method for getting of browser window position and width / heigtht.

If WD is not exist, Selenide will try to re-create driver in getWebDriver() (and getWebDriverAndCheck())

I want to use check before of my "screenshot" method to avoid re-creation of WD when it not exist (and do capture of system screenshot in that case)

dimand58:
I guess this check will be helpfull in the future and for another Selenide users =)

asolntsev:
Can't you use something like this: 
String scr = takeScreenshot();
If (scr == null) {
  // use robot to take screenshot
}

dimand58:
takeScreenshot () call getWebDriver() for getting of browser window size and that leading to re-creation of WD if it is not exist ...

dimand58:
I'm sorry, I mean that takeScreenshot() is my realisation of screenshot capturing. (not method in Screenshots class) 

I'm using system screenshot capturing via Robot in any case and crop it to browser window size if WD alive.
 (it necessary for capturing of <select> content in IE browser, cause that content invisible for WebDriver screenshot capturing mechanism)

asolntsev:
I have merged the pull request, thank you!
Please note that I renamed method `isBrowserExist()` to `hasWebDriverStarted()` because this method already existed in `WebDriverThreadLocalContainer`. Please give a feedback if it's ok for you.

dimand58:
It's OK, thanks!

PR #174, opened 2015-04-07
A new method (and property) for clicking on the element via JavaScript

dimand58:
In some case clicking on the element in IE may be instable. I found a lot of discussions about this situation with testing in IE in global network.

1) This new feature provide new method of clicking on the element - via JavaScript. At now we can use standart .click() or .clickViaJs() (for our purposes).
2) If we want to migrate all .click() in our code - we can just use "selenide.click-via-js=true" property (default is false) for some projects with testing in IE.


asolntsev:
Thank you!
I guess old method `$.followLink()` was created a long time ago for the same reason...
I have merged your pull request. You can now manage `$.click()` implementation using `selenide.click-via-js=true` property.

dimand58:
Glad to help!

PR #175, opened 2015-04-10
Switch to window except some handles. Switch to last frame.

dimand58:
In some case need to switch to window with same title, but with some another handle. For this situation new method may be helpful.

In some case need to switch to frame which is included in another "parent" frames. For this purpose you can use new switchToLastFrame method


asolntsev:
@dimand58 Method `switchToLastFrame()` actually switches to **every** frame, not only to the last one.

asolntsev:
@dimand58 What is the real reason for using `switchToWindowExceptHandles()`?  Is it really useful to use both **title** and **handles** in the same method? 

My initial intention was free user from all these handles that are absolutely unusable. Title is much more convenient and human-readable.

dimand58:
@asolntsev 
Af course! Switching to the "child" frame impossible without switching to it "parent". (Frame not found exception)

switchToWindowExceptHandles is needed for situations when target window has same title as current window (in this case we can use handle for exception)

asolntsev:
Can you add an automated test for the new function? Please look at FramesTest for example .

dimand58:
OK.

What about switchToWindowExceptHandles  ?

asolntsev:
Yes, please this test too.

Sent from my HTC

----- Reply message -----

dimand58:
@asolntsev , hi !
Please check new commit with tests implementation.

P.S. Can you release minor version of Selenide with selenium 2.46 ?

asolntsev:
@dimand58 Yes, I am going to release Selenide 2.19 with Selenium 2.46 soon. Probably tonight.

dimand58:
Thanks! I look forward to =)

dimand58:
This comment in Selenium has led to performance degradation when we switch frames often (switching by name/id method in 2.46):
https://github.com/SeleniumHQ/selenium/commit/98161bc5c743918318e81be04f62cf37e49cd971

I created another solution for switchToLastFrame method, because of issue fixing in Selenium is very long process.

dimand58:
Method was enhanced, please review last comment. This version still fast, even on IE8/IE11.

asolntsev:
dimand58 
Well, now I've got the point. Thank you for the tests!
I have merged your pull request. Now i suggest the following improvements that seem reasonable for me:

1. Method `switchToLastFrame()` without parameters should switch to default content (now it just does nothing that may be confusing).
2. The name `switchToLastFrame()`is confusing. At least for me. It's not obvious that it actually switches to **all** frames one-by-one. And intention of this method (switching into **inner** frame) is not obvious. <br>Let's rename it:<br> `Selenide.switchTo().innerFrame("parentFrame", "childFrame_2", "childFrame_2_1");`
3. Use better API for `switchToWindowExceptHandles()`: <br>
`switchTo().windowExceptHandles()`. <br>
I am still in doubt. This method `switchTo().windowExceptHandles()` still seems quite inconvenient and unclear for me. It's just not convenient to user to store some previous handle in variable and use it later. Trying to invent some better api... How will you actually use this method? Can you bring a real example?

asolntsev:
@dimand58 I have released Selenide 2.19 (with Selenium 2.46.0) and your pull request.

PR #178, opened 2015-04-27
Describe.appendAllAttributes() NPE check

shimkiv:
In some cases "executeJavaScript()" may return NULL what can lead to NPE at "new TreeMap<String, String>(map);"


asolntsev:
I found that NPE happens in IE. Merged it.

PR #215, opened 2015-09-03
Take into account element's "maxlength" attr for the JS value setter

shimkiv:
Otherwise additional issues could appear in case if there isn't server-side value length validation.


asolntsev:
@shimkiv Thanks, will merge it!

Two questions:
1. Why "Math.abs" is needed?  I guess "maxlength" cannot be negative.
2. Why this should be optional? I suggest to enable this behaviour by default.

I also suggest to add another feature to "setValue": if element is readonly or disable, setValue() should do nothing, or even throw an exception.

shimkiv:
@asolntsev thank you !

> 1. Why "Math.abs" is needed? I guess "maxlength" cannot be negative.

just in case :)

> 2. Why this should be optional? I suggest to enable this behaviour by default.

to maintain backward compatibility (for those who are already use JS set val feature)

PR #230, opened 2015-10-07
TODO done

Ariel-Isaacm:


asolntsev:
@Ariel-Isaacm Thank you!
But, I you sure the screenshot will get good name? Currently it's a name of current test-method. It could not be initialized when using testing frameworks other than JUnit or TestNG.

Ariel-Isaacm:
You're welcome :dancer: 
I will look for another way to get the method's name and implement it :)

PR #254, opened 2015-12-16
Add updateHash method (for ajax apps)

fabienbancharel:
Hello.
We start using Selenide to test GWT apps, and it is very pleasant !

Here is a proposition of a method to ease navigation in ajax applications like GWT, since 
open( "#GwtPlace1:");
open( "#GwtPlace2:"); => makes a full page reload

Inspired by a solution proposed here : https://code.google.com/p/selenium/issues/detail?id=5165

updateHash("GwtPlace2:"); would just navigate to the new Place, without reloading the whole page from the server.

### DIFF
--- a/src/main/java/com/codeborne/selenide/Selenide.java
+++ b/src/main/java/com/codeborne/selenide/Selenide.java
@@ -58,6 +58,17 @@ public class Selenide {
     navigator.open(absoluteUrl);
     mockModalDialogs();
   }
+  
+  /**
+   * Update the hash of the window location (usefull to navigate in ajax apps, since open(url) makes a full page reload).
+   * @param hash value for window.location.hash
+   */
+  public static void updateHash(String hash) {
+    if (hash.startsWith("#")) {
+        throw new IllegalArgumentException("hash should not start with #");
+    }
+    executeJavaScript("window.location.hash='" + hash + "'");
+  }
 
   private static boolean doDismissModalDialogs() {
     return !supportsModalDialogs() || dismissModalDialogs;


asolntsev:
@fabienbancharel
Cool idea! Let's do it. But, what about implementing this logic in method open("#GwtPlace2")? Then we don't need to introduce new method.

fabienbancharel:
> what about implementing this logic in method open("#GwtPlace2")
I was more comfortable to add a new method than to modify an existing one, but yes, it is a good idea.
Do you close this pull request ?
Do you want me to make another one, or do you prefer to implement it ?

fabienbancharel:
In fact, there could be a problem if it is implemented in open() :
At the beginning of a test, open("#HomePlace:") is expected to make a first request to the server with the start url + "#HomePlace:".
If open("#HomePlace:") just updates the location.hash, it misses the first server call needed for Single Page Applications.
You would have to do :
open(""); open("#HomePlace:");
which is not exactly the same as open(startUrl + "#HomePlace:");

Maybe a specific method is preferable ?
(updateHash, openHash ...)
What do you think ?

asolntsev:
It's up to you.
Probably the easiest way would be to create new pull request.

Andrei Solntsev

2015-12-16 18:27 GMT+02:00 fabienbancharel <notifications@github.com>:

asolntsev:
Well, good point.
Ok, let's use the method updateHash

Andrei Solntsev

2015-12-16 20:18 GMT+02:00 fabienbancharel <notifications@github.com>:

asolntsev:
Just for your information, I have also added tests for the new method `updateHash()`. See https://github.com/codeborne/selenide/commit/b4c75a1b8da980c73e36b89f940062b982836edc

It will be released soon in Selenide 3.0

Issue #196, opened 2015-07-12
Add file downloading mechanism without "href"


asolntsev:
Sometimes file download starts without clicking link with "href" attribute. Selenide could provide way to download those files too.

I guess it's only possible to do using proxy server (like browsermob-proxy).

leslyarun:
@asolntsev  Can u pls provide me with an example of using browsermob-proxy in Selenide ???

leslyarun:
@asolntsev I'm using a corporate proxy. How to make browsermob to use the same proxy server ? please provide me with an example

asolntsev:
I don't know exactly.
I guess BrowserMobProxy could be able to intercept all http requests and
forward them to your corporate proxy. But I don't know if it actually can
do this.

If it cannot, we need to find or create such a proxy server.

Andrei Solntsev

2015-08-05 14:31 GMT+03:00 leslyarun <notifications@github.com>:

leslyarun:
I'm using a proxy address and port no. for accessing the internet. But when I use BrowserMob, it creates it's own proxy server. 
If there is an way to make browsermob to use a particular proxy address and port no. , it would be very helpful. Do u have any idea ? @asolntsev

asolntsev:
@leslyarun I don't know exactly if there is a way. This requires some investigation. If you get it sooner, please inform me.

dimand58:
Hi all. It's simple!
You can tune Browser Mob for auto downloading of files based on Content-Type header.
Test case:
1) Navigate to some URL
2) Click on button which lead to response with Content-Type = application/pdf (downloading of PDF file after click)
3) Browser Mob Proxy catch Content-Type=application/pdf, save file to temp folder and replace response body with full path to PDF file 

First of all you must add&start BMob (You can do it in @BeforeSuite in TestNG as example):
```java
        // proxyServer  - field in class with type BrowserMobProxyServer
        proxyServer = new BrowserMobProxyServer(port);
        proxyServer.start();
        WebDriverRunner.setProxy(ClientUtil.createSeleniumProxy(proxyServer));
```

Then you must to create response filter (thanks StackOverflow and Google)
```java
public class FileDownloader implements ResponseFilter {

    private Set<String> contentTypes = new HashSet<String>();
    // ExConfiguration.CONFIG_DEFAULT_DOWNLOAD_PATH - my own field with absolute path to temporary folder
    private File tempDir = new File(ExConfiguration.CONFIG_DEFAULT_DOWNLOAD_PATH);
[...]

PR #267, opened 2016-01-09
Implement #196 Add file downloading mechanism without "href"

dimand58:
@asolntsev please review first version of implementation #196 issue.

How to use in end-user projects:
```java
        Selenide.prepareDownload("C:\\SomeFolder", "", "application/x-sdlc");
        WebDriverRunner.getWebDriver().get("http://javadl.sun.com/webapps/download/AutoDL?BundleId=113227");
        // Downloaded file with name chromeinstall-8u66.exe
        File tmp = new File(Selenide.$(By.tagName("body")).getText());
```
or
```java
        Selenide.prepareDownload("C:\\SomeFolder","SomeFile.exe","application/x-sdlc");
        WebDriverRunner.getWebDriver().get("http://javadl.sun.com/webapps/download/AutoDL?BundleId=113227");
        // Downloaded file with name SomeFile.exe
        File tmp = new File(Selenide.$(By.tagName("body")).getText());
```
or
```java
        Selenide.$(By.id("button_for_file_download_after_click"))
                .prepareDownload("C:\\SomeFolder", "", "application/pdf")
                .click();
        File tmp = new File(Selenide.$(By.tagName("body")).getText());
```

TODO:
1) Review
2) Some tests
3) Merge conflicts

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -55,7 +55,7 @@ dependencies {
   provided group: 'com.codeborne', name: 'phantomjsdriver', version: '1.2.1', transitive: false
   
   // For BrowserMob Proxy:
-  testCompile group: 'net.lightbody.bmp', name: 'browsermob-proxy', version: '2.0.0', transitive: false
+  compile group: 'net.lightbody.bmp', name: 'browsermob-core-littleproxy', version: '2.1.0-beta-4'
   testRuntime group: 'org.slf4j', name: 'slf4j-jdk14', version: '1.7.13'
   testRuntime group: 'net.sf.uadetector', name: 'uadetector-resources', version: '2014.10'
   testRuntime 'com.fasterxml.jackson.core:jackson-databind:2.6.4'
--- a/src/main/java/com/codeborne/selenide/Selenide.java
+++ b/src/main/java/com/codeborne/selenide/Selenide.java
@@ -12,6 +12,7 @@ import org.openqa.selenium.support.ui.FluentWait;
 
 import java.net.URL;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collection;
 import java.util.List;
 import java.util.logging.Level;
@@ -58,6 +59,14 @@ public class Selenide {
   public static void open(URL absoluteUrl) {
     navigator.open(absoluteUrl);
     mockModalDialogs();
+  }
+
+  /**
+   * @see SelenideElement#prepareDownload(String, String, String...)
+   */
+  public static void prepareDownload(String toFolder, String fileName, String ... expectedContentTypes) {
+    FileDownloader.instance
+                  .prepareDownloadViaBrowserMob(toFolder, fileName, Arrays.asList(expectedContentTypes));
   }
   
   /**
--- a/src/main/java/com/codeborne/selenide/SelenideElement.java
+++ b/src/main/java/com/codeborne/selenide/SelenideElement.java
@@ -438,6 +438,19 @@ public interface SelenideElement extends WebElement, FindsByLinkText, FindsById,
   File download() throws FileNotFoundException;
 
   /**
+   * Prepare BrowserMobProxy for catch file (via Content-Type) from next http responses.<br/>
+   * When response with selected Content-Type will be received you will see page with full path to downloaded file.<br/>
+   * Then you can use: $(By.tagName("body")).getText();<br/>
+   * For manual clearing of filters: WebDriverRunner.getBrowserMobProxy().getFilterFactories().clear();
+   * @param toFolder downloads folder absolute path
+   * @param fileName name for output file. Can be empty (name from response will be used).
+   * @param expectedContentTypes array with content types like application/pdf and etc
+   * @throws IllegalStateException if BrowserMobProxy server is null or stopped for current thread
+   * @throws IllegalArgumentException if toFolder or expectedContentTypes are empty
+   */
+  SelenideElement prepareDownload(String toFolder, String fileName, String ... expectedContentTypes);
+
+  /**
    * @return the original Selenium WebElement wrapped by this object
    */
   WebElement toWebElement();
--- a/src/main/java/com/codeborne/selenide/WebDriverRunner.java
+++ b/src/main/java/com/codeborne/selenide/WebDriverRunner.java
@@ -2,6 +2,7 @@ package com.codeborne.selenide;
 
 import com.codeborne.selenide.impl.WebDriverContainer;
 import com.codeborne.selenide.impl.WebDriverThreadLocalContainer;
+import net.lightbody.bmp.BrowserMobProxyServer;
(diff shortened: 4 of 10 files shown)


dimand58:
Fix

asolntsev:
@dimand58 Thank you! Great! It's almost what we need.

The only problem is the API. The new downloading mechanism is particularly needed in situation when we don't know exact URL. You typically fill some form with 100500 fields, click "submit", and file starts downloading. 

Try for example:
* http://idemo.bspb.ru/#corporate
* Cards -> Acquiring -> "View terminals"
* Put some checkboxes and click "Report"
* Expected result: an XLS file starts downloading

I suggest to make existing method `$.download()` more powerful without adding any new methods.

This method:
```java
  File report = $("button").download();
```

currently only works with elements `<a href=***>`.
What I suggest is to make it work with any downloads. It could work like this:

1. Test calls `$("button").download();`
2. Selenide sets some ThreadLocal flag like `Selenide.browserMobProxyActive.set(true);`
3. Selenide clicks the `button`
4. Because of flag, BMP stores all received files to some folder.
5. Selenide method `download` returns first of these files.
6. Selenide method `download` resets flag back: `Selenide.browserMobProxyActive.set(false);`
7. While this flag is `false`, BMP does not store any files (to avoid degrading performance).

dimand58:
@asolntsev 2 problem exists:
1) BrowserMob must be activated BEFORE of WebDriver instance
2) Web must know ALL mime types for files

asolntsev:
@dimand58 
1) BrowserMob must be activated BEFORE of WebDriver instance
Yes, but it's not a problem. Selenide will activate BrowserMob in the very beginning (before creating webdriver). But this BrowserMob instance will not save any files. It will just proxy requests from webdriver to server. Then, method `$.download()` will set some internal flag to TRUE, and only then BrowserMob will save all responses with header like "attachment" to local files.

2) Web must know ALL mime types for files
I suggest that Selenide can return the first saved file with header like "attachment".

With your implementation, use needs to know the exact URL for downloading file. But user cannot always know it. For example, file can be downloaded as a result of form submission.

dimand58:
OK.
Response filter will be save file if HTTP response header have "Content-Disposition: attachment;". In this case no need to know MIME type of downloaded file.

But what about redirections?
Test case:
1) Go to http://www.java.com/ru/download/chrome.jsp
2) Click on BIG RED BUTTON (java dowload)

Actual:
1) No Content-Disposition in response header
2) 1 redirect occured (to http://sdlc-esd.oracle.com/ESD6/JSCDL/jdk/8u66-b18/chromeinstall-8u66.exe)
3) AGAIN - no Content-Disposition in response header

mplushnikov:
+1 for downloading files from elements without href attribute, like buttons.

asolntsev:
@mplushnikov agree. We need this. I will work on it soon.

asolntsev:
@dimand58 Sorry for the long delay. I am almost ready to merge the pull request, but I have one problem.

BMP doesn't work for me when I run tests locally (against http://127.0.0.1:9000). But it works for remote sites like google.com. Don't you know a solution?

dimand58:
@asolntsev , hi!
Give me a few days, I will try to solve this problem.

dimand58:
@asolntsev 
Are you tried to use DNS name of your PC  instead of 127.0.0.1 ? I suppose that this trick can help

asolntsev:
How can I use name of computer? I cannot hardcode it in tests - tests
should be easily runnable on any machine.

dimand58:
You can get current pc name in code, via some method maybe

asolntsev:
Yes, something like InternetAddress.getLocal().getHost() should return name
of current computer. But it can be very slow, depending on network
configuration.

Why couldn't BMP just work with IP? Both Jetty an Netty can.

dimand58:
@asolntsev What does happen when your test opens http://127.0.0.1:9000 via proxy?
I tried on my PC and all works great.

asolntsev:
This is my case that does not work:
1. open http://127.0.0.1:9000 via proxy
2. download a file

Result: proxy intercepts all requests, but does not intercept server
response. So I cannot get the downloaded file.
Expected result: I want to get server response.


Andrei Solntsev

2016-04-22 19:03 GMT+03:00 Demin Dmitrii <notifications@github.com>:

dimand58:
@asolntsev strange...
My steps:
1) Init BMP before first WD creation
 ```
BrowserMobProxy proxy = new BrowserMobProxyServer();
 {
     proxy.start(1234);
     Proxy seleniumProxy = ClientUtil.createSeleniumProxy(proxy);
     WebDriverRunner.setProxy(seleniumProxy);
 }
```
2) Add filter before URL opening:
`proxy.addResponseFilter(new ResponseFilterImpl());`
3) Open URL:
`Selenide.open("http://127.0.0.1");`
4) In debug mode I can see call of filterResponse method in ResponseFilterImpl

**Comment by dimand58** [comment]
Browser - Google Chrome
BMP:
```
<dependency>
        <groupId>net.lightbody.bmp</groupId>
        <artifactId>browsermob-core-littleproxy</artifactId>
        <version>2.1.0-beta-5</version>
</dependency>
```

asolntsev:
@dimand58 Please review my failing test. In Selenide master branch, there are 2 tests:
* `integration.proxy.BrowserMobProxyTest` - it uses deprecated `net.lightbody.bmp.proxy.ProxyServer` - and it works
* `integration.proxy.BrowserMobProxyServerTest` - it uses newer `net.lightbody.bmp.BrowserMobProxyServer` - and it doesn't work for localhost. Please take a look at test  `local_https()` that is currently ignored. 

@dimand58 Can you say why this test fails?

dimand58:
@asolntsev , got it. I will investigate this test at evening today

dimand58:
@asolntsev can you update your certificate? It can be one of reason of current problem. 
image

asolntsev:
Do you mean the SSL certificate of https://localhost:xxxx?

I deliberately use self-signed (untrusted) SSL certificate there because I
want to test how Selenide works with untrusted certs. It's because many
companies use untrusted certs on their test-servers.

Typically webdrivers have method like "setTrustCerts(true)", I guess BMP
also has such a method.

Andrei Solntsev

2016-04-25 17:57 GMT+03:00 Dmitriy Demin <notifications@github.com>:

dimand58:
I created new self-signed cert (with valid expiration date and from/to fields 127.0.0.1) but problem still exists. 
**However, in IE11 page opens via proxy!** I'm totaly confused

dimand58:
@asolntsev 
I've found and checked the solution - we should use latest version of BPM (beta-7-SNAPSHOT which not realised yet). With that BMP version local web works good.
I guess we should to wait a little bit

asolntsev:
@dimand58 I implemented this idea a little bit differently in Selenide 3.9

Huge thanks for your ideas and support!

PR #274, opened 2016-01-25
Comparison operator with an expected size for collections

vasilevichra:
Yo Andrey!

Please review my commit with comparison operator for collections of WebElements.
I have added new feature for the method shouldHaveSize(int):

.shouldHaveSize(GREATER, 1);
.shouldHaveSize(EQUALS, 1); works like good old .shouldHaveSize(1);
.shouldHaveSize(GREATER, 1).shouldHaveSize(LESS_OR_EQUALS, 5);

### DIFF
--- a/src/main/java/com/codeborne/selenide/CollectionCondition.java
+++ b/src/main/java/com/codeborne/selenide/CollectionCondition.java
@@ -2,6 +2,7 @@ package com.codeborne.selenide;
 
 import com.codeborne.selenide.collections.ExactTexts;
 import com.codeborne.selenide.collections.ListSize;
+import com.codeborne.selenide.collections.ListSize.Operator;
 import com.codeborne.selenide.collections.Texts;
 import com.codeborne.selenide.impl.WebElementsCollection;
 import com.google.common.base.Predicate;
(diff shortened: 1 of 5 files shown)


vinogradoff:
We should probably wait, until the selenide collection implementation will be updated. The problem is that any of shouldHaveSize() - can not work 100% reliable with ajax apps, as we don't know when collection is completely loaded yet. More methods that are not 100% reliable - more confusion.

vasilevichra:
I understand what you mean, but my case is not related to Ajax.
I have a elastic markup which may contain from 4 to 7 elements (like Youtube for example) depending on a screen resolution. I need to check it as $(".videos").shouldHaveSize(GREATER_OR_EQUALS, 1).shouldHaveSize(LESS_OR_EQUALS, 7) but Selenide don't offers way to do that.

asolntsev:
@vasilevichra Thank you for the pull request! Let's discuss nuances.

1. If number of elements depends on browser size. Isn't it more correct to set browser size explicitly and then test exact number of elements?
2. If we still decide to have these methods, I would prefer to have one method `$$.shouldHave()` and multiple `CollectionCondition` subclasses. Something like that:

* `$$("li").shouldHave(sizeGreaterThan(3));`
* `$$("li").shouldHave(sizeLessThan(7));`
* `$$("li").shouldHave(sizeBetween(3, 7));`

asolntsev:
@vasilevichra I did it! Thought, a little bit differently. See https://github.com/codeborne/selenide/commit/1c3094a1e28525c7efef97257a86fe9a5b5c5432

Anyway, thank you for the commitment!

PR #276, opened 2016-01-27
Fix exception message typo

dumbdev:
I believe the NoSuchWindowException is meant to read "Browser window is not found"


vinogradoff:
Thank you for PR!

PR #298, opened 2016-02-27
Possibiltiy to grab report results of @Listeners(TextReport.class) 

thasherwin:
I changed  the instance of SimpleReport that is being initialized in TextReport class to static. I have doubts whether this is the best solution because it's static. I did not know any other way to retrieve the report data because TextReport class is being called using listeners. I have no idea how to call methods from a class that is part of a Listener. 

This solution seems to work when I run a test.

In afterMethod I do see the results being printed

### DIFF
--- a/src/main/java/com/codeborne/selenide/logevents/SimpleReport.java
+++ b/src/main/java/com/codeborne/selenide/logevents/SimpleReport.java
@@ -1,5 +1,6 @@
 package com.codeborne.selenide.logevents;
 
+import com.codeborne.selenide.testng.HistoryReportData;
 import com.google.common.base.Joiner;
 
 import java.util.Collections;
@@ -10,7 +11,9 @@ import java.util.logging.Logger;
  */
 public class SimpleReport {
   private static final Logger log = Logger.getLogger(SimpleReport.class.getName());
-  private String reportLog = "";//added variable
+  private String reportLog = "";
+  private String testName = "";
+  private String methodName = "";
   private EventsCollector logEventListener;
 
   public void start() {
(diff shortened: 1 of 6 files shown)


asolntsev:
@thasherwin I guess SimpleReport cannot be static. In this case, parallel tests will not work correctly (they will share the same SimpleReport instance).

thasherwin:
Okay. I will see whether i can find another solution.

thasherwin:
@asolntsev 
I did it now differently. I created a new class with Hashmap to save the report data. That class is static. But the data that is saved is being done using key: testName and method name. So to request the report data, you need to mention in  after method the correct key to get the correct data. Key is: testName and MethodeName
By default no data is saved in the Hashmap class. It’s turned off. To turn it on, the user needs to call the method; HistoryReportData.SaveHistoryData(); in for example @BeforeTest annotation.

thasherwin:
Any update on this code? Is it good enough?

asolntsev:
@thasherwin Sorry, I haven't looked at it yet. 
Will do it during the weekend.

thasherwin:
OK. no problem and thanks

asolntsev:
Hi @thasherwin!
Sorry for the long delay. I have finally reviewed the pull request.

There are few problems with it.
1. package "com.codeborne.selenide.logevents" refers to package "com.codeborne.selenide.testng". 
No core package should depend on junit, testng or any other specific implementation.

2. Test cannot be uniquely identified by class name + method name. In case of parameterized tests, every test method can be executed multiple times (with different arguments). 

Can you remind, why do we need this?
Do you want to attach the TextReport to Allure report? How are actually going to do that? How are you going to use the new method `TextReport.getReportLog()`?

thasherwin:
Yes i would like to attach the TextReport in Allure report. To do that i need to attach it using this Allure  method:
```java
@Attachment(value = "Selenide Test details")
  public String createSelenideTestResultDetails(String screenShot) {
    return screenShot;
  }
```
I wanted to call the report saved using the newly hashmap created function

```java
@AfterMethod
	public void saveSelenideReportInAllure(ITestResult _result,ITestContext context){
		createSelenideTestResultDetails(HistoryReportData.getReportData(context.getCurrentXmlTest().getName(),_result.getMethod().getMethodName());
	}
```
I totally forgat about parameterized tests. 

Uhm i will than go for another solution. 
I will make a copy of your TextReport and SimpleReport class and add the allure attachment function mentioned above in the SimpleReport class so that i can capture the TextReport of Selenide in Allure. I tested this solution  and it works. 
It will be a personal solution as of now because i don't know any generic way. 

Thanks for reviewing my code. I hope that I will be able to do a commitment in the future.

thasherwin:
I am closing this pull request. For people who would like to do this, i suggest they just make a clone of the Textreport class and add Allure Annotations.

asolntsev:
@thasherwin thank you!

PR #310, opened 2016-04-03
Added utility for taking screenshot of single page element as image

Akkuzin:
Added utility for taking screenshot of single page element as in-memory image without using of temporary files


asolntsev:
@Akkuzin thank you for the pull request!

PR #314, opened 2016-04-14
An ability to use FindAll annotation

msdjl:
I believe it's good idea to allow usage of FindAll annotation in addition to FindBy to save "native" way to annotate lists of page elements.


msdjl:
Oh, sorry. It's misunderstanding.

asolntsev:
@msdjl No problems! Feel free to send new pull requests!

PR #319, opened 2016-04-28
Warning if implicitlyWait property is not equal 0

dimand58:


asolntsev:
@dimand58 Thanks for the pull request! Thought, I have some questions.

1. It's not sufficient to just check system property `webdriver.timeouts.implicitlyWait`. Users also can set implicit wait programmatically: `driver.manage().timeouts().implicitlyWait(30, TimeUnit.SECONDS);`

2. It's not sufficient to check implicit wait in method `WebDriverThreadLocalContainer.create()`. Users also can create WebDriver by their own and pass it via method `setWebDriver()`. In this case method `create()` is not used at all.

dimand58:
@asolntsev 
1) I have no idea how to do this (implicitlyWait just call WD command execution)
2) Fixed, see last commit

asolntsev:
@dimand58 we could check for webdriver.manage().timeouts() right inside shouldHave method?

PR #320, opened 2016-04-28
Implementation of basic auth. for many browsers

dimand58:
Now we can use 
```java
Selenide.open("http://somepage.org", "SomeDomain", "JohnLogin", "JohnPassword");
```
if we want to pass basic authentication.
Supported (and tested) browsers:
1) IE11
2) Chrome
3) Firefox
4) PhantomJs
5) HtmlUnit

Additional:
- UnsupportedOperationException added to catch() clause in getJavascriptErrors() method (specially for HtmlUnit)
- Tests for Chrome, IE, PhantomJs disabled because executables are needed

### DIFF
--- a/src/main/java/com/codeborne/selenide/impl/Navigator.java
+++ b/src/main/java/com/codeborne/selenide/impl/Navigator.java
@@ -57,13 +57,12 @@ public class Navigator {
       if (!domain.isEmpty()) domain += "%5C";
       if (!login.isEmpty()) login += ":";
       if (!password.isEmpty()) password += "@";
-      int idx1 = url.indexOf("//") + 2;
-      url = url.substring(0, idx1 - 2)
-              + "//"
+      int idx1 = url.indexOf("://") + 3;
+      url = (idx1 < 3 ? "" : (url.substring(0, idx1 - 3) + "://"))
               + domain
               + login
               + password
-              + url.substring(idx1);
+              + (idx1 < 3 ? url : url.substring(idx1));
     }
 
     SelenideLog log = SelenideLogger.beginStep("open", url);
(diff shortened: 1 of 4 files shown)


asolntsev:
@dimand58 Do we really need to disable CSP? 
For example, we configure CSP policy, why not to test in via selenium tests?

`myProfile.setPreference("security.csp.enable", false);`

dimand58:
@asolntsev
I checked security.csp.enable in corporate network - it is useless proprety which have no influence on basic auth

asolntsev:
@dimand58 yes, it's useless for basic auth, but probably it's useful for something else? Do we really need to disable it?

src/main/java/com/codeborne/selenide/impl/Navigator.java
```diff
@@ -35,14 +45,32 @@ protected String absoluteUrl(String relativeUrl) {
...
+              + login
+              + password
+              + url.substring(idx1);
     }
 
     SelenideLog log = SelenideLogger.beginStep("open", url);
     try {
       WebDriver webdriver = getAndCheckWebDriver();
       webdriver.navigate().to(url);
+      if (isIE()) Selenide.switchTo().alert().authenticateUsing(new UserAndPassword(domain + login, password));
```
shimkiv: Hey Guys,
in case if BasicAuth is not required then exception will be thrown since we will not be able to switch to alert.
We should handle this case inside an additional try-catch block (without exception throwing) or introduce the new Configuration property

simple-elf: Bug confirmed in #366

PR #324, opened 2016-05-07
Implement of extended criteria class with new features

ddemin:
### New features for Selectors class:

1) **User-friendly description for element**:
`By loginButton = byId("123").as("Some super login btn");`
in some exceptions you will see "Element not found {Some super login btn}"
 instead of "Element not found {By.id: 123}" 

2) **Automatically switching to the frame before searching of element**:
`By loginButton = byId("123").inFrame("frameTop", "frameChild");`
or
`By loginButton = byId("123").inFrame(frameTopLocator, frameChildLocator);`
in this case switching to defaultContent and frameChild will be executed before searching of element, just use:
`$(loginButton).click();`

**In you have a lot of nested frames you can init locators in your Page Object (as example):**

```
By frameTop = byId("frameTopId").inRoot();
By frameChild1 = byId("frameChild1").inFrame(frameTop);
By frameChild11 = byId("frameChild11").inFrame(frameChild1);
By btnInChild11 = byId("someBtn").inFrame(frameChild11);
```

and then just
`$(btnInChild11).click();`

3) **Automatically waiting of some conditions before searching of element**:
in any place of code:
`Selectors.setExpectant(new ClassWhichImplementsExpectant());`
ClassWhichImplementsExpectant should implements ExBy.IExpectant interface. Any code in ClassWhichImplementsExpectant.actionsForWait() will be executed before searching of element (created via Selectors class of course). _As example - waiting of jQuery.active == 0 condition_
You can manage of waiting via .withWait and .withNoWait methods:
`By loginButton = byId("123").withWait();`

4) **Parametrized locators:**
`By btn = byXpath("//button[@onclick='{0}']");`
than
`$(btn.with("some value of onclick")).click();`
### Default values:

1) Expectant - null
2) Waiting - no
3) Frame - undefined (no switching)
You can set default values if you want:

```
Selectors.setDefaultFrame("frameTop", "frameChild");
Selectors.setWaitByDefault(true);
```

### DIFF
--- a/src/main/java/com/codeborne/selenide/Selectors.java
+++ b/src/main/java/com/codeborne/selenide/Selectors.java
@@ -211,7 +211,7 @@ public abstract class Selectors {
 
     @Override
     public String toString() {
-      return description + " (by text: " + elementText + ")";
+      return description.isEmpty() ? "by text: " + elementText : description;
     }
 
     String getXPath() {
@@ -229,7 +229,7 @@ public abstract class Selectors {
 
     @Override
     public String toString() {
-      return description + " (with text: " + elementText + ")";
+      return description.isEmpty() ? "with text: " + elementText : description;
     }
 
     String getXPath() {
--- a/src/test/java/com/codeborne/selenide/SelectorsTest.java
+++ b/src/test/java/com/codeborne/selenide/SelectorsTest.java
@@ -12,7 +12,7 @@ public class SelectorsTest {
   @Test
   public void withTextUsesXPath() {
     By selector = Selectors.withText("john");
-    assertTrue(selector instanceof By.ByXPath);
+    assertTrue(selector instanceof ExBy.ExByXPath);
     assertEquals("with text: john", selector.toString());
     assertEquals(".//*/text()[contains(normalize-space(.), \"john\")]/parent::*", ((WithText) selector).getXPath());
   }
@@ -21,7 +21,7 @@ public class SelectorsTest {
   public void withTextEscapesQuotes() {
     By selector = Selectors.withText("Ludvig'van\"Beethoven");
     assertEquals("with text: Ludvig'van\"Beethoven", selector.toString());
-    assertTrue(selector instanceof By.ByXPath);
+    assertTrue(selector instanceof ExBy.ExByXPath);
     assertEquals(".//*/text()[contains(normalize-space(.), concat(\"Ludvig'van\", '\"', \"Beethoven\"))]/parent::*", 
         ((WithText) selector).getXPath());
   }
@@ -30,7 +30,7 @@ public class SelectorsTest {
   public void byTextUsesXPath() {
     By selector = Selectors.byText("john");
     assertEquals("by text: john", selector.toString());
-    assertTrue(selector instanceof By.ByXPath);
+    assertTrue(selector instanceof ExBy.ExByXPath);
     assertEquals(".//*/text()[normalize-space(.) = \"john\"]/parent::*", ((ByText) selector).getXPath());
   }
 
@@ -38,7 +38,7 @@ public class SelectorsTest {
   public void byTextEscapesQuotes() {
     By selector = Selectors.byText("Ludvig'van\"Beethoven");
     assertEquals("by text: Ludvig'van\"Beethoven", selector.toString());
-    assertTrue(selector instanceof By.ByXPath);
+    assertTrue(selector instanceof ExBy.ExByXPath);
     assertEquals(".//*/text()[normalize-space(.) = concat(\"Ludvig'van\", '\"', \"Beethoven\")]/parent::*", 
         ((ByText) selector).getXPath());
   }


admizh:
:thumbsup: 
@dimand58 pls Please describe the point three. This can give a problems when running parallel tests

dimand58:
@admizh , @asolntsev 
Is no problem with parallel tests will be if you call
`Selectors.setExpectant(new ClassWhichImplementsExpectant());`
at once in your code like
```
// Some base page object as example
static {
  Selectors.setExpectant(new ClassWhichImplementsExpectant());
}
```
Example:
- you are testing some web with jQuery and want to wait jQuery.active==0 before searching of element (of course you must set default waiting flag or use .withWait() method for locators)

vinogradoff:
I am not quite sure, I understand your changes well enough…


Let’s go it over step by step.

1) User-friendly description for element:
By loginButton = byId("123").as("Some super login btn");
in some exceptions you will see "Element not found {Some super login btn}"
instead of "Element not found {By.id: 123}"


Can you give more examples? Does it work for CSS Selector, or XPath too?
What is log output before and after the change?

$(„.someclass“).click();
$(„.someclass“).as(„fancy button“).click();

What are the logs, if $(„.someclass“) is not visible?

Does it work with assertions too?

like $(„.someclass“).as(„fancy button“).shouldHave(text(„Login“));

——————-
I almost think, it’s better to split that in 4 PullRequests because then we could merge a few, a let the other outside more easily.

@asolntsev: what do you think?

dimand58:
@vinogradoff 
It will work for any type of selectors which created via Selectors class. I just modified the toString method:
```
@Override
public String toString() {
  return description.isEmpty() ? "By.linkText: " + linkText : description;
}
```
You will see description instead 'By.typeOfLocator: locatorValue' text in any exception which calls toString method of criteria

vinogradoff:
I would probably vote for both description and selector texts then.  It helps to understand errors better I suppose. 



Viele Grüße
Alexei Vinogradov 
--
Sent from mobile phone

asolntsev:
@dimand58
I suggest creating a separate pull request for every feature.

### 1. User-friendly description for element
Ok, Let's do it! 
It's not useful for me, but probably can be useful for others. 

### 2. Automatically switching to the frame before searching of element
I am not sure that we really need it. 
  * The logic of locating elements would be hidden and hard to debug
  * What if this element can be located in different frames?

Actually it can be done simpler. 
Let's create a method `runInFrame` for running given code block inside frame. 
Then your problem can be solved with just one line:
`runInFrame("frame-name", () -> $("yyy").click());`

### 3. Automatically waiting of some conditions before searching of element
It seems to be overcomplicated. 
And as @admizh said, it can create problems with parallel tests. I mean, `expectant` is  static field. What if one thread set it, but another thread doesn't expect this? Behaviour of tests may vary depending on order and speed of tests. 

Again, it can be done simpler: just add your waiting code right after `open` method. Then it's explicit. Straightforward. Easy debuggable. 

### 4. Parametrized locators
I really don't see why it can be useful. 
You can just use `String.format` for creating `By` instances? 

Like: 

```java
String buttonLocator = "//button[@onclick='%s']";
By btn1 = byXpath(String.format(buttonLocator, "some value of onclick"));
By btn2 = byXpath(String.format(buttonLocator, "another value of onclick"));
```

yashaka:
First of all, as I already mentioned in #447, to put all proposed logic into By - is not Selenide way. I believe Bys - are classic Selenium locators and they should be left like this. Maybe they should be enhanced in Selenium, but not in Selenide, to not confuse users. 

All "magic" lives in SelenideElement in Selenide, so all proposed features should be considered to be added to SelenideElement, not to By. 

1. 
I don't understand the benefits. Want to have good name for your button? - put it into method with the correspondent name - and it will appear in stacktrace, or add the allure @Step annotation to it - and it will appear in the report. 

But to blow the code with a bunch of additional comment-like strings - is not a good way... 
I would expect more arguments for this feature before giving it a green light.

By the way, we have already discussed something similar for Selene: https://github.com/yashaka/selene/issues/114

2.
Good idea. But again - should be implemented in SelenideElement, not By.
@asolntsev 
I believe that the argument about "the logic will be hidden" is not relevant.
First of all the majority of Selenide users use PageObjects. That means that the "frame logic" will nevertheless be hidden - and that is exactly what the majority wants.

And it does not matter, where the logic will be hidden - in the method with a direct switch to a frame, or in a variable with "more concise way to indicate frame switch". 

Even your code: `runInFrame("frame-name", () -> $("yyy").click());` is not much better in context of "not hiding logic" than `$("yyy").inFrame("frame-name").click();`. I would even say that the proposed version is much more concise, readable and userfriendly - exactly what the Selenide should be about, not? :)

3. 
IMHO still a good idea but overcomplicated implementation.
And this is still a "Selenide way" idea - to be implicit as much as possible. 

Usually, Selenide waits implicitly for visibility of elements. But what is sometimes the user should wait for the element to be enabled, not just visible?

Then it would be pretty easy to write once:

SelenideElement submit = $("#submit").when(enabled);

("when" can be changed to another alias)

This implementation will be threadsafe, and not complicated much. 

Of course, we still have a workaround for this:
```java
public SelenideElement submitBtn(){
    return $("#submit").shouldBe(enabled);
}
```

which is, by the way, a bit "stronger"...

here is why...
according to what I remember. 
If somebody called should method on some SelenideElement - then the implicit waiting are disabled, and only explicit waiting for the "condition in should" will be performed.

This means that the following code can unexpectedly fail for the user:

SelenideElement submit = $("#submit").when(enabled);
...
submit.shouldHave(text("Ok")).click()
because the text may be "Ok", but the button is still "disabled", and the "implicit waiting for custom condition" will be disabled by the explicit should.

To fix this, we can make "custom implicit conditions" - being not disabled by "should" waiting, but this may overcomplicate the implementation a bit... Do we need this then? I am not sure... This is the point to think about... Nevertheless, the implementation should be simpler than originally proposed by the author.


4. 
Just create a method with parameter and that's all. I don't see any valuable reason to overcomplicate implementation for this case.

-----
**About Default Values**
First of all, all such configurations already live in Configuration, not in Selectors.
In addition - default waiting - is already not "no" in selenide. Selenide waits by default for visibility and this is good.

**P.S.**

I really do not understand, why to use 

```
By submit = By.id("submit");
...
$(submit).
$(submit).
$(submit).
```

over

```
SelenideElement submit = $("#submit");
...
submit.
submit.
submit.
```

Do you like writing more letters with your fingers (bodyfingerbuilding?) ? creating more objects instead of one? Be less readable and a bit more bulky for the reader of your code?

Sometimes the one would use one locator in different contexts, like 
```
By item = ...
...
$(item)
$$(item)
```
But the following is much more readable:
```
ElementsCollection items = ...
...
items.first()
items
```
no?
:)

yashaka:
@ddemin 
Yeah, and nevertheless, this should be all different issues and all different Pull Requests.

asolntsev:
Sorry for the long delay.

@ddemin I agree with @yashaka. Can you prepare several independent PRs, one per feature?

PR #333, opened 2016-05-11
Prevent NPE throwing during LogEventListener removing attempt.

shimkiv:


asolntsev:
@shimkiv  If I am not mistaken, the original NPE can happen if `listeners.get()` returns null. 
But in this case the finally block will also throw NPE. Am I right?

vinogradoff:
I don't like catch(NPE..){donothing - bad style}

shimkiv:
@vinogradoff really ? me too :)
I might be wrong but as I remember in previous Selenide versions there were a number of places where different exceptions were not processed so I assumed that it was kind of requirements or something like that.
Anyway whether it will be enough just to log the exception ?

asolntsev:
@shimkiv I don't like catch (NPE) at all. 
Can you provide a real case for reproducing the problem? In which cases it happens?

shimkiv:
@asolntsev yeah, definitely, I've tried to figure out the reason of this rare case with no luck so far. I'll try to provide it later.

PR #349, opened 2016-06-14
This commit will provide initial Marionette support

ridiekel:
This pull request will add initial support for MarionetteDriver.

For more info on this driver: https://developer.mozilla.org/en-US/docs/Mozilla/QA/Marionette/WebDriver

I tested this in my project since the latest selenium version and Firefox 47 are having issues. Switching to the marionette driver fixes this.


vinogradoff:
Thank you for you commit!

Issue #358, opened 2016-07-22
Disable creating .html file, while creating screenshot


kirdeleg:
If it possible, it would be nice to create feature, that can turn off creating .html file, while creating screenshot, because sometimes .html files are not needed, but they are takes a lot of place. For our project it is really important, because of lack of memory on VM.

JustAlexOne:
+1
Having the same issue. Would really appreciate to have a possibility to "switch off" html files creation

asolntsev:
@kirdeleg @JustAlexOne We will do it, no problems.
But let me suggest you a workaround for this problem. 

Screenshots and *.html files are saved only for failed tests.
So, just **fix your tests**!

Why should one have a lot of failing tests? Is it normal? 
P.S. We typically have 0 failing tests, in some bad days we have 3-5 of them.

JustAlexOne:
In my case we do take screenshots before each test step and also for passed
tests as well. I understand this is useless (if not to say in other words),
but this is a customer request to have test reports with screenshots for
each step...
That's we have a lot of screenshots even for not failing tests.
Again, I understand, that's a senseless stuff

asolntsev:
Closing the feature request: it has been implemented in #359

PR #359, opened 2016-07-23
Disable creating .html file, while creating screenshot

BorisOsipov:
#358 Disable creating .html file, while creating screenshot

### DIFF
--- a/src/main/java/com/codeborne/selenide/Configuration.java
+++ b/src/main/java/com/codeborne/selenide/Configuration.java
@@ -161,6 +161,13 @@ public class Configuration {
   public static boolean screenshots = Boolean.parseBoolean(System.getProperty("selenide.screenshots", "true"));
 
   /**
+   * Does Selenide need to save page source on failing tests.
+   *
+   * Default value: true
+   */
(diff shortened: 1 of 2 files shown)


asolntsev:
Thanks for the PR!
However, there is one problem. If failed to take screenshot, Selenide returned html (with default settings). Now it does not return html. :(

BorisOsipov:
Hi, Andrei!

Yes indeed, now method takeScreenShot doesn't return html.
But it's a rare case when screenshoting fails and we need return smth != null. And I believe end-user should setup its enviroment in these cases, if they want screenshots.
Also, it looks strange: method takeScreenShot() in fact could return path to non-screenshot file.

Nevertheless, if you don't think so, let me know and I change fix.

Thanks,
Boris

asolntsev:
@BorisOsipov I don't like losing old functionality.
However, it was not well-designed. Error message should show path to both ".html" and ".png" files - see issue [210](https://github.com/codeborne/selenide/issues/210). So, let's merge current PR as is and implement issue [210](https://github.com/codeborne/selenide/issues/210).

Thanks for the PR! Merged.

PR #362, opened 2016-07-28
marked hasClass condition method as deprecated

biomaks:
Deprecated hasClass(String) method as it distracts users and does exactly the same as cssClass(String) method


asolntsev:
@biomaks thanks, merged!

Issue #366, opened 2016-08-04
Getting following error when selenide was updated to 3.7


anilreddy:
Hi @asolntsev,

While running our test suites by upgrading the selenide version to 3.7. I am facing following exception. It is working fine in 3.5 version.

```java
org.openqa.selenium.TimeoutException: Timed out after 120 seconds waiting for alert to be present
Build info: version: '2.53.1', revision: 'a36b8b1cd5757287168e54b817830adce9b0158d', time: '2016-06-30 19:26:09'
System info: host: 'KMI-DSERT956', ip: '193.225.171.68', os.name: 'Windows 7', os.arch: 'amd64', os.version: '6.1', java.version: '1.8.0_65'
Driver info: driver.version: unknown
selenide.url: https://xdev.basic.com/facility/secure/jas/facilityVal?timestamp=866939626117330
selenide.baseUrl: https://xdev.basic.com/facility/secure/jas/facilityVal
	at org.openqa.selenium.support.ui.FluentWait.timeoutException(FluentWait.java:292)
	at org.openqa.selenium.support.ui.FluentWait.until(FluentWait.java:261)
	at com.codeborne.selenide.SelenideTargetLocator.alert(SelenideTargetLocator.java:57)
	at com.codeborne.selenide.impl.Navigator.navigateToAbsoluteUrl(Navigator.java:72)
	at com.codeborne.selenide.impl.Navigator.open(Navigator.java:33)
	at com.codeborne.selenide.Selenide.open(Selenide.java:79)
	at com.codeborne.selenide.Selenide.open(Selenide.java:148)
	at com.codeborne.selenide.Selenide.open(Selenide.java:129)
	at com.basic.facility.regression.tests.LoginTest.setUpConfig(LoginTest.java:155)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:497)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:50)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:47)
[...]

PR #369, opened 2016-08-09
Fixed bug with IE and basic auth

simple-elf:
Problem mentioned in #366 and highlighted in #320.
So here is quick fix of that. I think if there is empty login parameter we don't need to handling alert for basic auth.

### DIFF
--- a/src/main/java/com/codeborne/selenide/impl/Navigator.java
+++ b/src/main/java/com/codeborne/selenide/impl/Navigator.java
@@ -69,7 +69,7 @@ public class Navigator {
     try {
       WebDriver webdriver = getAndCheckWebDriver();
       webdriver.navigate().to(url);
-      if (isIE()) Selenide.switchTo().alert().authenticateUsing(new UserAndPassword(domain + login, password));
+      if (isIE() && !"".equals(login)) Selenide.switchTo().alert().authenticateUsing(new UserAndPassword(domain + login, password));
       collectJavascriptErrors((JavascriptExecutor) webdriver);
       SelenideLogger.commitStep(log, PASS);
     } catch (WebDriverException e) {


anilreddy:
@simple-elf Thanks for the quick fix :-)

simple-elf:
Sorry for commit spam, I'm new on github =)

anilreddy:
@simple-elf also can you add a test-case for it.

asolntsev:
@simple-elf Thank you for the fix.

NB! Why do you think that this is "quick fix" Do you think that it's a "dirty temporary solution"? Do you have an idea of f better solution?

asolntsev:
@simple-elf @anilreddy I have merged the fix and released selenide 3.8.1
It will appear in maven central repo in few hours.

simple-elf:
We need to test not only this fix, but all code in IE too, but it's hard to find such service for IE. Maybe SauceLabs? I need to investigate it.

simple-elf:
Test shoul be smth like this
```
  @Test
  public void canPassWithoutBasicAuthInIe() {
    assumeTrue(isIE());
    Selenide.open("http://httpbin.org/get");
    assertThat(source(), containsString("\"url\": \"http://httpbin.org/get\"")); // "": ""
  }
```

asolntsev:
+1
Actually we need a VPS with Windows. Because we need to 
1) run Selenide tests in IE, and
2) build NSelenide - selenide port to .NET

Don't you know a good and cheap Windows hosting?

simple-elf:
For running tests in IE we probably can use SauceLabs, it has free VM's for open source projects with IE

simple-elf:
@anilreddy 3.8.1 already in maven
https://mvnrepository.com/artifact/com.codeborne/selenide/3.8.1

anilreddy:
@simple-elf yeah am already using it.

PR #379, opened 2016-08-22
Turn on temporary disabled test for firefox

BorisOsipov:
Bug http://bit.ly/1JKT4AE has been fixed. Turn on related tests.


BorisOsipov:
Corresponding test failed on Travis CI, but reason as I can see that is for tests used 
"firefox 31.0esr" when bug aren't fixed yet.
On my local machine with Firefox 47.0.1 all tests passed.

asolntsev:
@BorisOsipov Really? Are you sure? I still cannot run any tests on FireFox browser...

PR #435, opened 2016-12-06
Making Element column size in Text Report dynamic

ostap-oleksyn:
By default the size of Element column in text report is 20 characters. If there is an element which length is bigger than 20 chars(e.g. By.xpath: ./[@class='rc']/h3/a) the text report gets messed up:
<a href="http://www.imagebam.com/image/760ad3518848859" target="_blank">imagebam.com</a>

My changes making the Element column size dynamic, based on the element length. If the length is less than 20, the column will have the default size of 20. If the length is greater or equals than 20, then the Element column size will be equal to element length + 1 (the +1 is needed to separate it from the | delimiter). 
So the report will look like this:
<a href="http://www.imagebam.com/image/f9b2a5518848851" target="_blank">imagebam.com</a>

### DIFF
--- a/src/main/java/com/codeborne/selenide/logevents/SimpleReport.java
+++ b/src/main/java/com/codeborne/selenide/logevents/SimpleReport.java
@@ -3,6 +3,7 @@ package com.codeborne.selenide.logevents;
 import com.google.common.base.Joiner;
 
 import java.util.Collections;
+import java.util.OptionalInt;
 import java.util.logging.Logger;
 
 /**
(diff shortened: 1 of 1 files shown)


vinogradoff:
nice PR! You seem to have some checkstyle issues with a change, could you please fix them?
run

    gradle check

to see the problems.

ostap-oleksyn:
I fixed the warning caused by my changes, but there are still a lot of warnings(159 to be exact) saying "Missing a Javadoc comment." in classes that I haven't changed. Any thoughts? I've never used gradle before this.

vinogradoff:
missing javadocs is okay for now :)  warnings are eternal „to do“-s :)

PR #442, opened 2016-12-20
CollectionConditions now accept a descriptive message (similar to Conditions)

jaagupkymmel:
See https://github.com/codeborne/selenide/issues/440

### DIFF
--- a/src/main/java/com/codeborne/selenide/CollectionCondition.java
+++ b/src/main/java/com/codeborne/selenide/CollectionCondition.java
@@ -1,10 +1,13 @@
 package com.codeborne.selenide;
 
 import com.codeborne.selenide.collections.*;
+import com.codeborne.selenide.ex.ExplainedUIAssertionError;
+import com.codeborne.selenide.ex.UIAssertionError;
 import com.codeborne.selenide.impl.WebElementsCollection;
 import com.google.common.base.Predicate;
 import org.openqa.selenium.WebElement;
 
+import javax.annotation.Nullable;
 import java.util.List;
 
 public abstract class CollectionCondition implements Predicate<List<WebElement>> {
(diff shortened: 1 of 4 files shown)


BorisOsipov:
Hi @jaagupkymmel
Thanks for making the PR.
Could you please fix checkstyle issues:
https://s3.amazonaws.com/archive.travis-ci.org/jobs/185392668/log.txt

asolntsev:
Hi @jaagupkymmel!
Thank you for the pull request.
What do you think about a little simplification? 
What if we don't add a new class `ExplainedCollectionCondition`, but add a field `message` to the existing class `CollectionCondition`? It seems that we will have less code in this case...

jaagupkymmel:
Hey guys!
@BorisOsipov, I'll fix the warning shortly. The commit I based my changes on already had warnings at build, so I didn't spot them.
@asolntsev I'll need to go over the code again, and I'll update when I get to it, but from what I can remember, I wanted to use the {{message}} field (like {{Condition}} does}}, but IIRC, the error message was build differently for {{CollectionCondition}} from {{Condition}} and {{CollectionCondition::toString}} wasn't used in the process.

jaagupkymmel:
Hi!

I fixed checkstyle errors and a failing test. For some reason, there's still a test that fails on my machine and when I run checkstyle task locally, itdoesn't seem to pick up the config files.

The reason I chose to create `ExplainedCollectionCondition` is simple, I wanted `CollectionCondition` to be as close to `Condition` as possible and `Condition` has an inner class called `ExplainedCondition`.

asolntsev:
@jaagupkymmel Thank you. Now your reasoning for `ExplainedCollectionCondition` is clear.
And you are right, we need to create a separate class `ExplainedCollectionCondition`, and we cannot add field `message` to `CollectionCondition` because some `CollectionCondition` instances can be shared (for example, `CollectionCondition.empty`). 

Now I continue my analysis. 
The next thing that I don't like is class `ExplainedUIAssertionError`. Do we really need one more exception class? It seems to be too much of code just to show message. Can we simplify it?

asolntsev:
@jaagupkymmel ping!

jaagupkymmel:
Sorry for the late reply, I've been fairly busy lately.

I don't have the time to look into it right now, but IIRC, when a `Condition` fails, the generated `Exception` message uses `Condition::toString` behind the scenes, which is not the case with `CollectionCondition` failing, which means the solution without a `ExplainedUIAssertionException` class would require changes to each `UIAssertionException` that can be thrown whan a  `CollectionCondition` fails.

sidelnikovmike:
Hi guys! Any update on this PR?

BorisOsipov:
hi @sidelnikovmike

feel free to pick up this and address issues from https://github.com/codeborne/selenide/pull/442#issuecomment-275548005

sidelnikovmike:
@BorisOsipov Sounds good! I'll try! Thanks!

sidelnikovmike:
Guys, I move code to [another PR](https://github.com/codeborne/selenide/pull/749).

rosolko:
New reference https://github.com/codeborne/selenide/pull/749

PR #449, opened 2017-01-05
Added ability to set Capabilities and custom Firefox profile using Configuration.

ostap-oleksyn:
As per this [request](https://github.com/codeborne/selenide/issues/444) I added the ability to set Desired Capabilities using Configuration class:
```
DesiredCapabilities myCapabilities = new DesiredCapabilities();
myCapabilities.setCapability("someCapability", "value");
Configuration.capabilities.set(myCapabilities); 
```
<br>
You can also add a custom profile to your capabilities:

```
FirefoxProfile profile = new FirefoxProfile();
profile.setPreference("intl.accept_languages", "de, en, en-us");
myCapabilities.setCapability(FirefoxDriver.PROFILE, profile);

```
<br>
This way your custom profile will override the default profile Selenide passes to the FirefoxDriver. If no profile is passed to the custom capabilities, the default Selenide profile will be used.

### DIFF
--- a/src/main/java/com/codeborne/selenide/Configuration.java
+++ b/src/main/java/com/codeborne/selenide/Configuration.java
@@ -309,7 +309,7 @@ public class Configuration {
    * <b>Usage:</b><p>
    * DesiredCapabilities myCapabilities = new DesiredCapabilities();<p>
    * myCapabilities.setCapability("someCapability", "value");<p>
-   * Configuration.capabilities.set(myCapabilities)
+   * Configuration.capabilities.set(myCapabilities);
    */
   public static ThreadLocal<DesiredCapabilities> capabilities = new ThreadLocal<>();
 }
--- a/src/main/java/com/codeborne/selenide/Configuration.java
+++ b/src/main/java/com/codeborne/selenide/Configuration.java
@@ -1,5 +1,7 @@
 package com.codeborne.selenide;
 
+import org.openqa.selenium.remote.DesiredCapabilities;
+
 import java.util.logging.Logger;
 
 import static com.codeborne.selenide.Configuration.AssertionMode.STRICT;
@@ -300,4 +302,8 @@ public class Configuration {
 
   public static FileDownloadMode fileDownload = FileDownloadMode.valueOf(
       System.getProperty("selenide.fileDownload", HTTPGET.name()));
+
+
+
+  public static ThreadLocal<DesiredCapabilities> capabilities = new ThreadLocal<>();
 }
--- a/src/main/java/com/codeborne/selenide/Configuration.java
+++ b/src/main/java/com/codeborne/selenide/Configuration.java
@@ -304,6 +304,12 @@ public class Configuration {
       System.getProperty("selenide.fileDownload", HTTPGET.name()));
 
 
-
+  /**
+   * Sets additional WebDriver capabilities.<p>
+   * <b>Usage:</b><p>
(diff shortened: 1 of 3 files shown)


asolntsev:
@ostap-oleksyn why `Configuration.capabilities` is ThreadLocal, thought all other fields in `Configuration` are not? Shouldn't we keep consistency?..

vinogradoff:
I don’t we need this any more, as since 4.2 the SystemProperties „capabilites.*“ are transformed in DesiredCapability - automatically.

There is some point to include also FirefoxProfile and ChromeOptions to have complete configuration in hand. I will do it in 4.3 or perhaps even in 4.2.1 as it is a small fix.

There are too many arbitary capabilites nowadays, that is why I don’t want to put some of the in Configuration.
And in the most cases, I want independent tests, that I would tweek with SystemProperties more than with Configuration Parameters.

ostap-oleksyn:
@asolntsev I just decided to make it thread safe in case someone would need to set different capabilities in different threads while running tests in parallel(I think it would be nice to make all configurations thread safe in the future). This can be changed if that's an issue.

@vinogradoff I know about SystemProperties „capabilites.*“, but using SystemProperties you can only set capabilities that have String as a value. There are capabilities that accept boolean as a value, for example acceptSslCerts. So if you try to set that capability like this:
```
@Test
public void test() {
    System.setProperty("capabilities.acceptSslCerts", "false");
    open("https://google.com.ua");
}
```
you will get an exception:
```
java.lang.ClassCastException: java.lang.String cannot be cast to java.lang.Boolean

	at org.openqa.selenium.firefox.FirefoxDriver.populateProfile(FirefoxDriver.java:183)
	at org.openqa.selenium.firefox.FirefoxDriver.extractProfile(FirefoxDriver.java:168)
	at org.openqa.selenium.firefox.FirefoxDriver.<init>(FirefoxDriver.java:135)
	at com.codeborne.selenide.webdriver.WebDriverFactory.createFirefoxDriver(WebDriverFactory.java:127)
	at com.codeborne.selenide.webdriver.WebDriverFactory.createWebDriver(WebDriverFactory.java:47)
	at com.codeborne.selenide.impl.WebDriverThreadLocalContainer.createDriver(WebDriverThreadLocalContainer.java:240)
	at com.codeborne.selenide.impl.WebDriverThreadLocalContainer.getAndCheckWebDriver(WebDriverThreadLocalContainer.java:113)
	at com.codeborne.selenide.WebDriverRunner.getAndCheckWebDriver(WebDriverRunner.java:136)
	at com.codeborne.selenide.impl.Navigator.navigateToAbsoluteUrl(Navigator.java:68)
	at com.codeborne.selenide.impl.Navigator.open(Navigator.java:31)
	at com.codeborne.selenide.Selenide.open(Selenide.java:81)
	at com.codeborne.selenide.Selenide.open(Selenide.java:55)
	at com.codeborne.selenide.webdriver.Tst.test(Tst.java:12)
	...
```

vinogradoff:
Thanks for the showing the issue with boolean capabliity! 

The design of Capability sucks :)

I will look if it is typical for Capability true/false to be booelan (I am afraid - not).

We can always solve it with like capabilites.boolean.acceptSslCerts - but it looks ugly.

ostap-oleksyn:
@vinogradoff You can see the list of Firefox capabilities and their values types [here](https://github.com/SeleniumHQ/selenium/wiki/DesiredCapabilities).

PR #456, opened 2017-01-20
Presentation mode

kkochetov:
Instead of https://github.com/codeborne/selenide/pull/392

After conversation with @vinogradoff

Presentation mode can be activated by `selenide.PresentationMode.active` config option.
It allow make site presentation from tests.
Other config options:
```
 /**
  * Flash element before click
  * Default: true
  */
 public static boolean flashElements = Boolean.parseBoolean(System.getProperty("selenide.PresentationMode.flashElements", "true"));
 /**
  * Mark element before click
  * Default: false
  */
 public static boolean markElements = Boolean.parseBoolean(System.getProperty("selenide.PresentationMode.markElements", "false"));
 /**
  * Delay before click in ms.
  * Default: 200
  */
 public static Integer delayBeforeCommand = Integer.parseInt(System.getProperty("selenide.PresentationMode.delayBeforeCommand", "200"));
 /**
  * Element flasher color
  * Default: #00ff00
  */
 public static String flashColor = System.getProperty("selenide.PresentationMode.flashColor", "#00ff00");
 /**
  * Element marker color
  * Default: #00ff00
  */
 public static String markColor = System.getProperty("selenide.PresentationMode.markColor", "#00ff00");
```
In presentation mode selenide try flash element before click, doubleClick, contextClick, and followLink.
Also wait some ms before every action. 

Also here impements some kind of before/after action hooks(https://github.com/codeborne/selenide/issues/447).
For this just implement HookAction interface.
Then register yuor action with 
`HookPerformer.getInstance().addBeforeAction(String name, HookAction action)`
or
`HookPerformer.getInstance().addAfterAction(String name, HookAction action)`
Examples in MarkAction and FlashAction.

Sorry for my bad English.


BorisOsipov:
Hi folks.

Actually the idea is nice.
I think we can achieve the same aims implementing `WebDriverEventListener` and adding implementation by `WebDriverRunner#addListener` if it need.

kkochetov:
Hi.
I think, better use hook functional on selenide layer, not on selenium layer.
Selenide can do more than one selenium action for one $().click, for example.
With this case `WebDriverEventListener` will be called several times.
And, with current implementatiom, we can do something before/after should commands.
Please, correct me if I'm wrong.

BorisOsipov:
Ok. Got it

src/main/java/com/codeborne/selenide/hookperformers/HookAction.java
```diff
@@ -0,0 +1,23 @@
...
+public interface HookAction {
+  /**
+   * Condition for activation
+   *
+   * @param element WebElement
+   * @param methodName  Name of method, who called hook
+   * @return boolean
+   *
+   */
+  boolean conditionForAction(WebElement element, String methodName);
```
BorisOsipov: Could you please add command args here? It will be helpful for implementing other 'Hooks' e.g. some kind of loggers and so one.

src/main/java/com/codeborne/selenide/hookperformers/HookAction.java
```diff
@@ -0,0 +1,23 @@
...
+
+  /**
+   * Action when activated
+   *
+   * @param element WebElement
+   * @param methodName  name of method, who called hook
+   */
+  void action(WebElement element, String methodName);
```
BorisOsipov: Could you please add command args here? It will be helpful for implementing other 'Hooks' e.g. some kind of loggers and so one.

kkochetov: Of course.
In last commit.

kkochetov:
summon @vinogradoff @asolntsev

asolntsev:
@kkochetov I will review the last changes, thank you. Sorry for the delay.

By the way, have you noticed that the build fails? https://travis-ci.org/codeborne/selenide/builds/206495943
Could you research why it fails?

kkochetov:
@asolntsev 
As far as i understand travis logs, build was terminater with "No output has been received in the last 10m0s" 
Is there a way to restart build without pushing?
Also i fing one failed test integration.ConfirmTest > canCancelConfirmDialog. Trying to fail it locally.

SergeyPirogov:
It's useless. Presentation mode in library that should make life of people who is doing a lot of testing. Are you sure? I even can't imagine how we can use this for debugging.

asolntsev:
@kkochetov @vinogradoff I tend to agree with @SergeyPirogov: I am not sure if it's really useful feature. Tests should be automated. Why anybody should need to view how tests run?

kkochetov:
@asolntsev @vinogradoff OK. I understand. But what about actions before/after selenide commands? It is useless too? 
Is there any reason to move actions functional without presentations to another pull request? Or just edit this?

asolntsev:
@kkochetov Sorry for the long delay. It seems that actions before/after Selenide commands could be very useful. Somebody already asked it to improve stability of Angular tests.

asolntsev:
I found this old PR and started thinking... :)
Looking back at this PR, it seems to be quite useful in development/debugging/investigating flaky tests. 

@kkochetov Probably we could extract this functionality to a separate Selenide plugin? What do you think?

kkochetov:
Yep. But I need some time to remember what is all about. :-)
I will return with new PR and leave a link to it here.

PR #458, opened 2017-01-22
Add java.net.URI based method on Selenide class.

kazuki-ma:


asolntsev:
@kazuki-ma Thank you for the pull request.
Can you comment why is it useful?
Why `open(URL)` does not suite for you?

kazuki-ma:
Sorry for lack of description and thanks comment.

`java.net.URI` is newer than URL and have some useful method.

I use URI for my test code and `toURL()` in every open is a little wired...

But i don't want increase complexity as test said.. Let me close this PR.

PR #464, opened 2017-01-31
Adding customCommands support.

gmaslowski:
customCommands support
=====================
Basically supporting chaining of custom commands mixed in within SelenideElement. Example:
```java
Commands.getInstance().addCustomCommand("myCommand", (proxy, locator, args) -> {
        Integer i = (Integer) args[0];
        // do whatever you want
        return proxy;
      });

$("#valid-image").customCommand("myCommand", 1).click();
```

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,4 +1,8 @@
 == Changelog
+=== 4.2.1 (Release in January 2017)
+* fixed problems with int and boolean capability values, now converting automatically
+* added support for FirefoxProfile over commandline (-Dfirefoxprofile.<option>=<value>,
+* added support for ChromeOptions (args only) over commandline (-Dchromeoptions.args=<value1>,<value2>...)
 
 === 4.2 (Released 13.12.2016)
 * #431 browser=firefox uses legacy driver (works for <=47), browser=marionette - gecko driver (any Firefox)
--- a/src/main/java/com/codeborne/selenide/SelenideElement.java
+++ b/src/main/java/com/codeborne/selenide/SelenideElement.java
@@ -689,4 +689,15 @@ public interface SelenideElement extends WebElement, FindsByLinkText, FindsById,
    * @see com.codeborne.selenide.commands.TakeScreenshotAsImage
    */
   BufferedImage screenshotAsImage();
+
+  /**
+   * Execute registered custom command
+   *
+   * @param command name of registered custom command
+   * @param params parameters to pass to the registered custom command
+   * @return this element
+   *
+   * @see com.codeborne.selenide.commands.Commands
+   */
+  SelenideElement customCommand(String command, Object... params);
 }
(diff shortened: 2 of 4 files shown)


gmaslowski:
customCommands support
=====================
Basically supporting chaining of custom commands mixed in within SelenideElement. Example:
```java
Commands.getInstance().addCustomCommand("myCommand", (proxy, locator, args) -> {
        Integer i = ((Integer) args[0]);
        // do whatever you want
        return proxy;
      });

$("#valid-image").customCommand("myCommand",1).click();
```

gmaslowski:
@vinogradoff Could we merge this functionality?

asolntsev:
@gmaslowski This is probably a nice idea. Let me think a little bit...

But did you see [this implementation](https://github.com/codeborne/selenide/tree/master/src/test/java/integration/customcommands)? 

Isn't it more readable to use `$.tripleClick()` instead of `$.customCommand("tripleClick")`?

gmaslowski:
Yes. I saw that but it doesn't allow chaining the rest of defined methods
on SelenideElement.

7 lut 2017 21:29 "Andrei Solntsev" <notifications@github.com> napisał(a):

asolntsev:
@gmaslowski I asked people in Slack chat. Their reaction was that this feature doesn't seem really useful... They say that it's as simple s using own method `myCommand()`, and you don't really often need to chain it with other own methods. What do you think?

maciekh:
@asolntsev Consider such situation:
We have angular application with loading bar. When loading bar is active all elements on page are clickable but click doesn't trigger any action.
I need a method which can, for example, click element only when loading bar is not active.
Ofcourse i can make method waitForLoadingBarDesapear() but without Greg fix i can't write element().waitForLoadingBarDesapear().click()
Now I can only define my own click() method like in this example: https://github.com/codeborne/selenide/blob/master/src/test/java/integration/customcommands/MyFramework.java
But it will work only for element found by simple $_ method.
E.g  $$.find(...).tripleClick() 
$_(byAttribute(...)).tripleClick() 
$_(byText(...)).tripleClick() 
are not possible

BorisOsipov:
@maciekh I understand your case. And I ran into the same issues.
I guess in this  case it will be nice to have feature implemented in https://github.com/codeborne/selenide/pull/456
With such feature you will implement "before hook" that handle "loading bar" for all actions and after that you can dont worry about any progress bars :)
Please, have look at this pr. Maybe it keeps your needs.

maciekh:
@BorisOsipov Yes, maybe it can be usefull.
But will it work for all actions or only for actions listed in MarkAction.presentationMethods?
What with eg. selectOption, setValue etc?

BorisOsipov:
@maciekh 
It will work that you would like :)
I propose to implement you an `HookAction `interface and add to "Hook performer".
And handle all(or not all) selenide <->browser operations that you would like.

gmaslowski:
Closing because of different solution. Thanks for input.

PR #475, opened 2017-02-15
Driver manager

SergeyPirogov:
The second attempt to integrate automatic driver manager with Selenide.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -40,6 +40,7 @@ configurations {
 }
 
 dependencies {
+  compile group: 'io.github.bonigarcia', name: 'webdrivermanager', version: '1.6.0'
   compile('org.apache.httpcomponents:httpcore:4.4.6')
   compile('org.seleniumhq.selenium:selenium-java:3.0.1') {
     exclude group: 'org.seleniumhq.selenium', module: 'selenium-android-driver'
--- a/src/main/java/com/codeborne/selenide/webdriver/WebDriverBinaryManager.java
+++ b/src/main/java/com/codeborne/selenide/webdriver/WebDriverBinaryManager.java
@@ -1,0 +1,51 @@
+package com.codeborne.selenide.webdriver;
+
+import com.codeborne.selenide.Configuration;
+import io.github.bonigarcia.wdm.*;
+import org.apache.commons.lang3.StringUtils;
+
+import java.util.logging.Logger;
+
+import static com.codeborne.selenide.WebDriverRunner.*;
+
+/**
+ * Created by sepi on 15.02.17.
+ */
+public class WebDriverBinaryManager {
+
+  private static final Logger log = Logger.getLogger(WebDriverFactory.class.getName());
+
+  private static final String CHROME_PROP = "webdriver.chrome.driver";
+  private static final String MARIONETTE_PROP = "webdriver.gecko.driver";
+  private static final String OPERA_PROP = "webdriver.opera.driver";
+  private static final String EDGE_PROP = "webdriver.edge.driver";
+  private static final String IE_PROP = "webdriver.ie.driver";
+  private static final String PHANTOM_PROP = "phantomjs.binary.path";
+
+  static void setupBinaryPath() {
+    try {
+      if (isChrome() && isEmptyProperty(CHROME_PROP)) {
+        ChromeDriverManager.getInstance().setup();
+      } else if (isEdge() && isEmptyProperty(EDGE_PROP)) {
+        EdgeDriverManager.getInstance().setup();
+      } else if (isIE() && isEmptyProperty(IE_PROP)) {
+        InternetExplorerDriverManager.getInstance().setup();
+      } else if (isOpera() && isEmptyProperty(OPERA_PROP)) {
+        OperaDriverManager.getInstance().setup();
+      } else if (isPhantomjs() && isEmptyProperty(PHANTOM_PROP)) {
+        PhantomJsDriverManager.getInstance().setup();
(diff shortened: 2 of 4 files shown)


build.gradle
```diff
@@ -40,6 +40,7 @@ configurations {
 }
 
 dependencies {
+  compile group: 'io.github.bonigarcia', name: 'webdrivermanager', version: '1.6.0'
```
BorisOsipov: It will be nice to exclude transitive dependencies that already exists in project dependencies.
e.g. com.google.code.gson
Some projects doesn't not accept to have different jar's version in classpath.

SergeyPirogov: Ok, how can I know what dependency to exclude?

BorisOsipov: Hi! Sorry for delay. It's an minor issue.
It looks like for gradle it is hard to check it.
Let's skip it now. I will have look at this later.

BorisOsipov:
It's an awesome feature.
Thank you for it @SergeyPirogov !

src/main/java/com/codeborne/selenide/webdriver/WebDriverBinaryManager.java
```diff
@@ -0,0 +1,51 @@
...
+import com.codeborne.selenide.Configuration;
+import io.github.bonigarcia.wdm.*;
+import org.apache.commons.lang3.StringUtils;
+
+import java.util.logging.Logger;
+
+import static com.codeborne.selenide.WebDriverRunner.*;
+
+/**
+ * Created by sepi on 15.02.17.
```
BorisOsipov: Minor issue. It's an template generated by intellij idea ;)

build.gradle
SergeyPirogov: I will try to enable transitive for webdrivermanager and see what will happen

SergeyPirogov:
@asolntsev @vinogradoff what if we add automatic driver manager as experimental feature? So by default automatic driver management will be disabled,  but users will  be able to enable it by setting some flag like **Configuration.automatic_driver_manager = true**

What do you think?

rmarinsky:
@SergeyPirogov why not Configuration.automaticDriverManager or Configuration.automaticWebDriverManager?

SergeyPirogov:
@rmarinsky oh yes, Configuration.automaticDriverManager is good for me

JustAlexOne:
What if someone would like to use this driver manager, but also needs to keep with some particular version of the driver, instead of latest?
I think the possibility to set the driver version would also be a nice thing to have.

SergeyPirogov:
@JustAlexOne Yes, this is possible via property file.

mariaklimenko:
We use [WebDriverManager]( https://github.com/bonigarcia/webdrivermanager) with Selenide in our project and this is perfect match. Does it make sense to create new code when its possible to leverage existing tool?

asolntsev:
@SergeyPirogov @BorisOsipov @rmarinsky @JustAlexOne @mariaklimenko 
Hi all!
I am sorry for the long delay with this PR. :(

Let me explain why I am in doubt. 
1. I tried to use WebdriverManager in my project, and it failed. Because my project occasionally contains file `application.conf`, and WebdriverManager detects it and tries to parse. And fails (because our `application.conf` has a bit different format). I don't like tool that fails. :(

2. WebdriverManager only downloads webdriver binary, thought users still need to install browser. So, it only solves half of problem. At the same time, there is a solution that solves the whole problem: [TestContainers](https://github.com/testcontainers/testcontainers-java). It runs both browser and webdriver (using Docker). Probable we should integrate this into Selenide instead of WebdriverManager?

SergeyPirogov:
To use TestContainers user should have docker installed as well.

rmarinsky:
1. I'm not sure that allot of selenide's users have the application.conf file in projects
2. We can describe it in article for this functional and in javadoc - it's not a problem. If user want to use any of browser for testing it's clear that he should have installed binary. I think that it not a problem.
3. This feature will simplify first start of use selenide for any user and cool feature for speakers.
4. TestContainer on Mac and Windows use only legacy Docker Toolbox - and it have some problems with start and kill containers

SergeyPirogov:
Take a look at this commits 

https://github.com/bonigarcia/webdrivermanager/commit/a615372135425a325fbb9538165bceaf9ab8d052

There is no more application.properties

BorisOsipov:
I agree with @SergeyPirogov and @rmarinsky.
It is nice to have such a feature in Selenide even if it will be disabled by default.
TestContainers needs docker.Docker is less popular than browsers are and we need to install it:)

@asolntsev I realized you're worried about future support effort as well, but I guess we don't need to stuck on your single unsuccessful attempts to use DriverManager.

SergeyPirogov:
Also README mentions that wd manager uses webdrivermanager.properties now

https://github.com/bonigarcia/webdrivermanager#configuration

SergeyPirogov:
Moreover, anybody can set path by passing System property

```
protected WdmConfig() {
        conf = ConfigFactory.load(WdmConfig.class.getClassLoader(), System
                .getProperty("wdm.properties", "webdrivermanager.properties"));
    }
```

asolntsev:
@SergeyPirogov @BorisOsipov @rmarinsky @JustAlexOne @mariaklimenko

Good news: PR #610 has been merged to master. Finally!

SergeyPirogov:
Wow, it's a huge step forward. Waiting for the new release

mariaklimenko:
Thank you!

2017-09-29 7:51 GMT+03:00 Serhii Pirohov <notifications@github.com>:

Issue #486, opened 2017-02-25
getJavascriptErrors causes exceptions when element is not found in native app on Android


dkorobtsov:
Researching how Selenide and Appium are working together.
In general, everything works just fine, except for error handling.
Every single time the following line of code is called:

selenide/src/main/java/com/codeborne/selenide/ex/UIAssertionError.java

  protected String uiDetails() {
    return screenshot(screenshot) + jsErrors(jsErrors) + timeout(timeoutMs) + causedBy(getCause());
  }

Appium shows an exception:

Feb 25, 2017 10:58:02 PM com.codeborne.selenide.Selenide getJavascriptErrors
SEVERE: org.openqa.selenium.WebDriverException: Method has not yet been implemented (WARNING: The server did not provide any stacktrace information)

Wonder if there are any option to turn getJavascriptErrors feature off?

asolntsev:
@dkorobtsov Hi! Actually it's just a log message. Everything still works. As a quick workaround, you can just ignore this log (add a line to log4j configuration).

As a long-term solution, we need to disable getJavascriptErrors feature for Appium. Can you prepare a pull request?

dkorobtsov:
It's not just a log message :) There are more then 100 lines long stacktrace in console and log for every single "element not found" event. Even if it's expected. Not only it's very annoying but also significantly complicates debugging process.

As for pull request - still not sure how to fix this... Suppose something like this would solve the problem for Android native apps:
```
protected String uiDetails() {
     if (WebDriverRunner.getWebDriver() instanceof AndroidDriver) {
        return screenshot(screenshot) + timeout(timeoutMs) + causedBy(getCause());
     } else {
        return screenshot(screenshot) + jsErrors(jsErrors) + timeout(timeoutMs) + causedBy(getCause());
     }
}
```

asolntsev:
@dkorobtsov I have committed the fix. Please review: https://github.com/codeborne/selenide/commit/b4d2cf900bd1c403fbe4f4a600873fbca8bade10

dkorobtsov:
Should work as well, thanks a lot! I'll test changes on my environment and post an update.

PR #487, opened 2017-02-26
Fixes 'method not implemented' exceptions in Appium

dkorobtsov:
Fix addresses following issue;
getJavascriptErrors causes exceptions when element is not found in native app on Android #486
https://github.com/codeborne/selenide/issues/486
Added additional check to getJavascriptErrors() method.

### DIFF
--- a/src/main/java/com/codeborne/selenide/Selenide.java
+++ b/src/main/java/com/codeborne/selenide/Selenide.java
@@ -635,6 +635,9 @@ public class Selenide {
     if (!hasWebDriverStarted()) {
       return emptyList();
     }
+    if (WebDriverRunner.getWebDriver() instanceof AndroidDriver) {
+      return emptyList();
+    }
     else if (!supportsJavascript()) {
       return emptyList();
     }


asolntsev:
@dkorobtsov Thank you for the pull request, but it does not compile :(

See Travis CI output: 

```java
compileJava/home/travis/build/codeborne/selenide/src/main/java/com/codeborne/selenide/Selenide.java:638: error: cannot find symbol
    if (WebDriverRunner.getWebDriver() instanceof AndroidDriver) {
                                                  ^
  symbol:   class AndroidDriver
  location: class Selenide
```

It's because Selenide doesn't depend on Appium. 
Can we just use `try/catch(MethodNotImplementedException)`? Can you provide exact stack trace of the error?

asolntsev:
I am cancelling this PR.
I have committed another fix: https://github.com/codeborne/selenide/commit/b4d2cf900bd1c403fbe4f4a600873fbca8bade10

dkorobtsov:
Seems like fix does not work. Severity is now a warning, but it does not help against "Method not implemented exception".

Or just could user some instruction how system property could be set :) 

tried all possible combinations - writing to pom, settng along with driver:
System.setProperty("selenide.captureJavascriptErrors)", "false");

no effect...

dkorobtsov:
my pull request was actually working ;-)

just forgot to add a dependency to pull request:
import io.appium.java_client.android.AndroidDriver;

dkorobtsov:
Correction:
-Dselenide.captureJavascriptErrors=false actually works, thanks!

Btw, meanwhile found another workaround:
It's possible to turn off javascript during adroid driver setup as well:
        capabilities.setCapability("javascriptEnabled", false);

It does not solve problem completely, but errors are less annoying:)

PR #497, opened 2017-03-17
Implement methods $x(), $$x() witch can find elements via By.xpath 

rmarinsky:
I think I need to add some more tests
And what about interface SelenideElement? Need to add $x and $$x methtods to interface SelenideElement?

PR #503, opened 2017-03-25
Extension for PR #497 - new $x() and $$x) inside SelenideElement

ddemin:
Addition for https://github.com/codeborne/selenide/pull/497
Now we can use $x and $$x in methods call chains

### DIFF
--- a/src/main/java/com/codeborne/selenide/SelenideElement.java
+++ b/src/main/java/com/codeborne/selenide/SelenideElement.java
@@ -434,6 +434,12 @@ public interface SelenideElement extends WebElement, FindsByLinkText, FindsById,
 
   /**
    * ATTENTION! This method doesn't start any search yet!
+   * @see com.codeborne.selenide.commands.Find
+   */
+  SelenideElement $x(String xpathSelector);
+
+  /**
+   * ATTENTION! This method doesn't start any search yet!
    * Same as {@link #find(String, int)}
    * @see com.codeborne.selenide.commands.Find
    */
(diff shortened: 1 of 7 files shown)


SergeyPirogov:
@ddemin As I understand this PR bring possibility to write such code?

```
$x("//div").$x("//a")
```

What is the main value?

ddemin:
XPath - not only a // construction...
You can use something like $x(someByOrCssLocator).$x("./somexpath_locator_from_parent_node")

SergeyPirogov:
First of all $x() accepts only String parameter, so `$x(someByOrCssLocator)` is incorrect.
Second why just not use **$x("/div/a")** ?

ddemin:
It was a typo and it fixed already.
Second - as I know we need point to current context by '.'

JustAlexOne:
Oh man, - I've just implemented the same feature in my fork. My bad not to check already existing pull requests ).
The only thing I want to ask - wouldn't it be better to add 2 dedicated `Command`type classes 
1) `com.codeborne.selenide.commands.FindByXpath` 
2) `com.codeborne.selenide.commands.FindAllByXpath` 
Instead of adding a parameter to already existing class `com.codeborne.selenide.commands.Find`?
Having dedicated classes looks more like following the project architecture.
Moreover, in this case, no need to alter any other, already existing classes/methods.
Take a look at this [commit](https://github.com/JustAlexOne/selenide/commit/f843e5069a876ce1e9430cd99b1f40c09ae68352) for details.

JustAlexOne:
@SergeyPirogov about main value, let me give a code example:
Before:
```
SelenideElement form = $x("//div[@id='login-form']");
SelenideElement userName = form.$(By.xpath(".//input[@id='email']"));
SelenideElement password = form.$(By.xpath(".//input[@id='password']"));
```
After
```
SelenideElement form = $x("//div[@id='login-form']");
SelenideElement userName = form.$x(".//input[@id='email']");
SelenideElement password = form.$x(".//input[@id='password']");
```
Sure we can use other locators here instead of xpath, but let's assume we can't.

asolntsev:
@JustAlexOne I agree, adding 2 new classes `FindByXpath`/`FindAllByXpath` would be better.
@ddemin Can you do that?

asolntsev:
Closing this PR because another one did the same job: #533 
Anyway, thank you @ddemin!

PR #518, opened 2017-04-23
setValue refactoring

AleksanderPopov:
pr for issue https://github.com/codeborne/selenide/issues/508

### DIFF
--- a/src/main/java/com/codeborne/selenide/Configuration.java
+++ b/src/main/java/com/codeborne/selenide/Configuration.java
@@ -269,6 +269,20 @@ public class Configuration {
   public static boolean fastSetValue = Boolean.parseBoolean(System.getProperty("selenide.fastSetValue", "false"));
 
   /**
+   * If set to true, 'setValue' and 'val' methods of SelenideElement can work as 'selectOptionByValue', 'selectRadio'
+   * depending on the real control type, defined by element's tag.
+   *
+   * Will decrease performance of setValue, make it slower, but will also make tests implementation more "business oriented".
+   * With this property being set to true, tests may no longer be dependent on actual control implementation in html and
+   * be more abstract.
+   *
+   * https://github.com/codeborne/selenide/issues/508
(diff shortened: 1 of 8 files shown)


src/main/java/com/codeborne/selenide/Configuration.java
```diff
@@ -269,6 +269,20 @@ private static String getJenkinsReportsUrl() {
   public static boolean fastSetValue = Boolean.parseBoolean(System.getProperty("selenide.fastSetValue", "false"));
 
   /**
+   * If set to true, 'setValue' and 'val' methods of SelenideElement can work as 'selectOptionByValue', 'selectRadio'
+   * depending on the real control type, defined by element's tag.
+   *
+   * Will decrease performance of setValue, make it slower, but will also make tests implementation more "business oriented".
+   * With this property being set to true, tests may no longer be dependent on actual control implementation in html and
+   * be more abstract. The same effect can be achieved with PageObject pattern though.
```
asolntsev: > The same effect can be achieved with PageObject pattern though.

@AleksanderPopov I am not agree that PageObject is the rescue for this problem. It's just not related. 

1. You always can extract any code to a reusable method, no matter if you use PageObject or not. 
2. Inside of this method, you still need to think which html element it deals with. So the problems remains unresolved in this sense.

src/main/java/com/codeborne/selenide/commands/SetValue.java
```diff
@@ -11,49 +12,57 @@
...
+        setValueForTextInput(element, text);
+        return proxy;
+    }
+
+    private void setValueForTextInput(WebElement element, String text) {
+        if (text == null || text.isEmpty()) {
+            element.clear();
+        }
+        else if (fastSetValue) {
+            text = truncateMaxLength(element, text);
```
asolntsev: @AleksanderPopov Retrieving attribute `maxlength` is also slow. Can you move this `truncateMaxLength` thing into the JS code?

Issue #517, opened 2017-04-13
Remove unneeded log


asolntsev:
See https://files.gitter.im/AlexanderPoleschuk/CLAL/2017-04-12_15-59-29.png

It seems that we should add `if (proxy != null) ` before logging this:
```
INFO: Close proxy server: 1 -> null
```

asolntsev:
Fixed by https://github.com/codeborne/selenide/pull/524

PR #524, opened 2017-05-05
log proxy details only if it's not null

azakordonets:
Small bug fix that checks if proxy is not null before logging details about it. It's related to issue #517


azakordonets:
Weird, on my machine i just launched same task as on Travis and it works fine, but i cannot restart Travis job. Any ideas ?

BorisOsipov:
Hi @azakordonets
you can easy restart a Travis job just close and reopen pr.

Issue #529, opened 2017-05-11
Cannot find capabilities with browserName=ie when grid hub url specified in -Dremote


baranoff-ivan:
**Preconditions:**
```java
Configuration.remote = "<some hub url>";
Configuration.browser = WebDriverRunner.INTERNET_EXPLORER;
```

Failure when running tests with Selenide:
```
Caused by: org.openqa.selenium.WebDriverException: Error forwarding the new session cannot find : Capabilities [{nativeEvents=true, acceptSslCerts=true, requireWindowFocus=true, pageLoadStrategy=normal, browserName=ie}]
Build info: version: '3.4.0', revision: 'unknown', time: 'unknown'
...
```

**Root cause:**
Selenium node recognizes capabilities with browserName "internet explorer", but not "ie"

Of course, I could just set browser as below:
```java
Configuration.browser = "internet explorer";
```
But Selenide has different workarounds for IE which uses `isIE()` method. And they won't work in such case.

asolntsev:
@baranoff-ivan @BorisOsipov Thank you! Fix with PR #531.

PR #531, opened 2017-05-11
#529 Cannot find capabilities with browserName=ie when grid hub url specified

BorisOsipov:
- Set correct correct IE browser name in capabilities
- Add ability to set IE browser through "internet explorer" string

I reproduced issue and verified fix locally with IE 11 and selenium v.3.4.0


simple-elf:
I fixed IE with grid by
`Configuration.browser = "internet explorer";`

simple-elf:
But if we say "ie" to grid - it can't find browser, so we need to pass "internet explorer" to grid capabilities

PR #544, opened 2017-06-05
Create SelenideElement from Suppliers

nick318:
I suggest an idea of getting element from Supplier<? extends WebElement>. 
It can be helpful in different situations:  
- If we have method that return some div of block and we want to find some element from that block we can write something like that:
WebElement el = getDiv();
$(el).$(By.id("input").setValue("text");
But in this case we have a problem if something goes wrong with first element 'el', I mean if that element updates, we will get StaleElementReferenceException and Selenide cannot re-init that element.

But have a look at that example:
$(this::getDiv).$(input).setValue("text").

In that case, if something goes wrong, Selenide will invoke the given supplier and can re-init first element and others elements as well.

@asolntsev What do you think?


asolntsev:
@iCok It's an interesting idea. Yes, probably we should do it.

But can you describe a real case? How could your `getDiv()` method be implemented?
Can't you just use `$("selector")`?

iCok:
@asolntsev To answering to your question:
-- Can't you just use $("selector")? 
The problem is if someone already implemented methods that return WebElement.
Sometimes yes I can, but I think it is not necessary restriction to return everywhere only 'By' objects.

-- How could your getDiv() method be implemented?
some examples:

```
getDiv() {
return $$("collection").stream()
                .filter(elem -> elem.is(visible))
                .filter(elem -> elem.getText().equals("need text"))
                .findFirst().orElseThrow(() -> new AssertionError("Not found div"));
}
```
```
getDiv() {
   //some actions to click by 3-rd libraries like Robot or Sikuli.
  return driver.switchTo().activeElement(); //It helpful in IE browser, because it does not work by locators sometimes
}
```

asolntsev:
@iCok The build is failing. There are some checkstyle issues. Please execute `./gradle check` in your branch.

iCok:
Done

BorisOsipov:
I am closing it due inactivity. Feel free to start conversation again if you would mind to add such changes.

PR #545, opened 2017-06-06
Improve events logging

admizh:


asolntsev:
Hi @admizh !
1. Can you describe what problem you are trying to solve?
2. There are some Checkstyle issues in your PR - the build is red. Can you fix it?

admizh:
Hi @asolntsev
1. Selenide does not make calls to logger before any action in Proxy side. 
I just adding logger calling, now Selenide should call LogEventListener.onEvent(LogEvent currentLog) before and after.
It comfortable for reading console in real-time. And some other cases.

2. The second fix should solve the problem with the long text in the console table.
image
Probably not an elegant solution, because if the characters lengths exceeds MAX allowed value then i truncate it. I'll try implement transferring to the next next line in table - if the current implementation does not suitable for our community

3. For check style issue, this issue not related with may changes - i just touch this file in PR - i guess this issue migrated before checkstyle implementation for pull request. I will fix.

BorisOsipov:
I am closing this due no answers...

PR #552, opened 2017-06-17
Closes #551 Method `$.setValue()` should not fail if it could not tri…

### LINKED ISSUE #551: Method `$.setValue()` should not fail
.. if it could not trigger `change` event (for whatever reason).

Currently method `$.setValue()` does the following:
1. `element.clear();`
2. `element.sendKeys(text);`
3. trigger JS event `change`

If the latter fails, the whole `$.setValue()` fails. But it's not always reasonable. For example, step 3 could fail if webdriver doesn't support javascript.

asolntsev:
…gger change event (for whatever reason).

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -2,6 +2,7 @@
 
 === 4.5 (planned to 18.06.2017)
 * #549 Selenide waits too much for collections
+* #551 Method `$.setValue()` should not fail if it could not trigger change event (for whatever reason).
 
 === 4.4.3 (released 25.04.2017)
 * upgrade to selenium-java 3.4.0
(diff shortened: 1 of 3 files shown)


SergeyPirogov:
Have never faced such problem

asolntsev:
Me too. :)

SergeyPirogov:
So what is the reason? Just fail fast and not wait for the timeout?

asolntsev:
No, it's not "fail fast". Method will **not** fail in this case.

The reason is rather theoretical: it's **possible** that it happens, and method `setValue()` **should not fail** in this case.

PR #575, opened 2017-07-24
findAndAssertElementExists before performing scrollTo and getText

richmischler:
Have noticed that scrollTo and getText methods need to have a built-in wait before performing the action.
For example: 
`$(selenideElement).should(exist).scrollTo()` is necessary instead of just `$(selenideElement).scrollTo()`

### DIFF
--- a/src/main/java/com/codeborne/selenide/commands/GetText.java
+++ b/src/main/java/com/codeborne/selenide/commands/GetText.java
@@ -10,7 +10,7 @@ public class GetText implements Command<String> {
   
   @Override
   public String execute(SelenideElement proxy, WebElementSource locator, Object[] args) {
-    WebElement element = locator.findAndAssertElementIsVisible();
+    WebElement element = locator.findAndAssertElementExists();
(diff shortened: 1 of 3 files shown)


BorisOsipov:
Why only these methods?) 
why not click, setValue and so on?

richmischler:
Some of the methods already check to see if element is visible like click and setValue, but for these two I just need to see that it exists. I can definitely add to other methods. Just let me know which ones.

BorisOsipov:
Sorry, but I still don't understand why we need such checks(
_Current behavior is:_

- If element exists we will get text from them.
- If it isn't we will wait some time and get the NotFoundException.

_With your changes behavior is:_

- If element exists we will get text from it.
- If it isn't we will wait some time and get the NotFoundException.

what are the differences?

asolntsev:
@richmischler All Selenide method works like this:
1. Try to execute the operation - for example, `$.scrollTo()`
2. If failed, wait 100ms and go to step 1.

So, what will change with your suggestion?

BorisOsipov:
It  is already implemented in Selenide. Close PR.

PR #581, opened 2017-08-11
Refactoring

czerminski:
Code cleanup -> using Chain of Respobility pattern

### DIFF
--- a/src/main/java/com/codeborne/selenide/webdriver/ChromeDriverProcessor.java
+++ b/src/main/java/com/codeborne/selenide/webdriver/ChromeDriverProcessor.java
@@ -1,0 +1,84 @@
+package com.codeborne.selenide.webdriver;
+
+import org.openqa.selenium.Proxy;
+import org.openqa.selenium.WebDriver;
+import org.openqa.selenium.chrome.ChromeDriver;
+import org.openqa.selenium.chrome.ChromeOptions;
+import org.openqa.selenium.remote.DesiredCapabilities;
+
+import java.io.IOException;
+import java.util.Arrays;
+import java.util.List;
+import java.util.function.Supplier;
+import java.util.logging.Logger;
+
+import static com.codeborne.selenide.Configuration.browser;
+import static com.codeborne.selenide.Configuration.chromeSwitches;
+import static com.codeborne.selenide.WebDriverRunner.CHROME;
+
+class ChromeDriverProcessor extends DriverProcessor {
+
+  private static final Logger log = Logger.getLogger(ChromeDriverProcessor.class.getName());
+
+  private final Supplier<Boolean> condition = () -> CHROME.equalsIgnoreCase(browser);
+  private final DriverProcessor nextDriverProcessor;
+
+  ChromeDriverProcessor() {
+    this.nextDriverProcessor = new MarionetteDriverProcessor();
+  }
+
+  @Override
+  public WebDriver process(final Proxy proxy) {
+    return condition.get() ?
+            createChromeDriver(proxy) : nextDriverProcessor.process(proxy);
+  }
+
+  private WebDriver createChromeDriver(final Proxy proxy) {
+    DesiredCapabilities capabilities = createCommonCapabilities(proxy);
+    ChromeOptions options = createChromeOptions();
+    capabilities.setCapability(ChromeOptions.CAPABILITY, options);
+    return new ChromeDriver(capabilities);
+  }
+
+  private ChromeOptions createChromeOptions() {
+    ChromeOptions options = new ChromeOptions();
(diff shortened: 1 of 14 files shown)


czerminski:
I'll try to fix the errors, but to be honest I don't get it - I only modified 1 class and created several other but Checkstyle fails on different ones.

BorisOsipov:
Hi @czerminski 

Thank you for changes in this 'ugly' part of selenide codebase.
Some suggestions.
1) Please add unit tests to your code. It will helps us to verify it and change it in future.
2) I think it is not good to have dependencies in SomeDriverProcessor to AnotherDriverProcessor. Maybe it will be better if you set such dependencies in WebDriverFactory

Best regards,
Boris

czerminski:
Thanks for quick replay. I think that the whole idea behind Chain of Responsibility pattern is to have such dependencies, but I know that that pattern is behavioral not creational. I wondered if it is better to use just a simple factory or this one, but if you prefer factory then so be it

BorisOsipov:
@czerminski Sure, I realized what you would like to achieve. Let's wait sometime for feedback for other guys.

czerminski:
@BorisOsipov Sure

SergeyPirogov:
I think it's a huge and awesome change to resolve this issue https://github.com/codeborne/selenide/issues/555

asolntsev:
@czerminski Hi! Thank you for the pull request. Unfortunately it doesn't compile: see class `com.codeborne.selenide.webdriver.WebDriverFactoryTest`.

czerminski:
@asolntsev Yea, I'm aware of that. I just wanted to know whther my changes are gonna to be accepted or not. If you think that they are then I'll fix the unit tests. By the way, do you have any idea why Checkstyle fails on the classes that are not included in my commits? I cannot build the project locally because of that.

asolntsev:
@czerminski Basically I like the idea to split `WebDriverFactory` to multiple independent class. But I don't like two things in your current implementation:
1. The name `Processor`. It doesn't say anything. Every class can be named "processor". 
2. Those processor know about each other. It break the "single responsibility principle" - every processor has two responsibilities: to create a webdriver and to forward task to next processor. 
3. As a sequence, it's hard to get the full picture: list of all available "processors" and their order. It's also hard to understand where to add a new processor when it comes. 

I would suggest a little bit different design.
* Let's call them factories: `ChromeDriverFactory`, `GeckodriverFactory` etc.
* Let's them implement interface, e.g. `WebDriverFactory` with 2 methods:
    * public boolean supports()
    * public WebDriver create()
* (and probable they could extend a base class like `AbstractWebDriverFactory`)
* Let's create a list of all available factories and iterate them. Something like this:

```java
  List<WebDriverFactory> factories = asList(new ChromeDriverFactory(), new GeckodriverFactory(), ...);

  public WebDriver create() {
    for (WebDriverFactory factory in factories) {
       if (factory.supports()) {
         return factory.create();
       }
    }
  }
```

czerminski:
@asolntsev Please, correct me if I'm wrong, but there is no Dependency Container in Selenide? That's the main issue here because normally I would create a bean in some kind of a container module (like in Guice) and that way it would be easy to manage dependencies between processor's - which by the way is kinda generic name and I'd agree that it can be make less generic. At least, as far as I understand you don't want the Chain of Responsibilities pattern to be used in its raw form, but rather use some extra class to manage dependencies between the Processor's objects (here I would use DI container)?

asolntsev:
@czerminski You are right, Selenide doesn't currently use any dependency injection mechanism. And yes, probably we should start using one. 

But the question is not in DI. The question is: if "processors" should know about each other, or there should be one central class that holds all "processors". I believe the second option is better.

czerminski:
@asolntsev I absolutely agree in that matter, what I'm saying is that the DI module should be responsible for managing dependencies between 'Processors', but as you said - currently there is no DI container that cam be used, I'll use one more class to manage processors. Thanks

czerminski:
@asolntsev Please, just tell me if it's any closer to what would you like to achieve.

asolntsev:
@czerminski Yes, it's much better! Thank you.

One thing I have noticed: you are using lot of `static` method. This is typically not a good pattern. Can you make most of methods non-static? For example, `createInstanceOf`.

czerminski:
@asolntsev Any better? I replaced all static methods with instance methods using additional class - DefaultDriverFactory.

BorisOsipov:
I am closing this. PR author account is removed or banned :(

SergeyPirogov:
But we can fix comments ourselves and apply. This is a good change

BorisOsipov:
@SergeyPirogov I've managed this. 
I moved fixes to  https://github.com/codeborne/selenide/pull/606

Issue #520, opened 2017-04-26
setSelected method should check value after .click() action


AlexanderPoleschuk:
setSelected() method should care about the result of .click() action. 
If we need to tick the checkbox, then setSelected() should check that checkbox was selected and vice versa. This will avoid the situation when .click() is sent to wrong item(e.g. element locator is not checkbox).

The most useful way for this would be to check element for selected condition:
element.shouldBe(selected);
element.shouldNotBe(selected);

```
public class SetSelected implements Command<WebElement> {
  Click click = new Click();
  
  @Override
  public WebElement execute(SelenideElement proxy, WebElementSource locator, Object[] args) {
    boolean selected = (Boolean) args[0];
    WebElement element = locator.getWebElement();
    if (element.isSelected() ^ selected) {
      if (element.getAttribute("readonly") != null)
        throw new InvalidStateException("Cannot change value of readonly element");
      click.execute(proxy, locator, NO_ARGS);
    }
    return proxy;
  }
}
```

rosolko:
Good point to check actual behavior on tests level not on framework feature level.

PR #594, opened 2017-09-01
setSelected method should check value after .click() action #520

jane-ryabchenko:
Added SetSelected.isSelectable and SetSelected.isDeselectable methods.
If user tries to select not selectable element InvalidStateException will be thrown.
If user tries to de-select radio or option element InvalidStateException will be thrown.
Added Unit and Integration tests to cover every case.

### DIFF
--- a/.idea/codeStyleSettings.xml
+++ b/.idea/codeStyleSettings.xml
@@ -1,0 +1,9 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project version="4">
+  <component name="ProjectCodeStyleSettingsManager">
+    <option name="PER_PROJECT_SETTINGS">
+      <value />
+    </option>
+    <option name="PREFERRED_PROJECT_CODE_STYLE" value="Default (1)" />
+  </component>
+</project>
--- a/.idea/compiler.xml
+++ b/.idea/compiler.xml
@@ -1,0 +1,26 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project version="4">
+  <component name="CompilerConfiguration">
+    <resourceExtensions />
+    <wildcardResourcePatterns>
+      <entry name="!?*.java" />
+      <entry name="!?*.form" />
+      <entry name="!?*.class" />
+      <entry name="!?*.groovy" />
+      <entry name="!?*.scala" />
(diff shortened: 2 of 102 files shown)


SergeyPirogov:
Well, guys with modern UI frameworks it's all useless!

jane-ryabchenko:
@SergeyPirogov did you mean that this pull request will never be reviewed/merged ?

asolntsev:
@jane-ryabchenko Not exactly :)

But yes, it's true: I am in doubt. I am afraid that this change will make things slower, but will not always solve the problem. I don't claim that your PR is definitely wrong, I am just in doubt. 


Anyway, let's start with a simple question: 
1. What problem are you trying to solve? What is the real case?
2. Why your change doesn't correspond to the title? Title says "setSelected method should check value AFTER .click() action", but actually it checks something else (type, not value) and BEFORE action.

jane-ryabchenko:
@asolntsev 
I'll start with second question.

2) Required change described in #520 doesn't makes much sense to me.
What is the point of reckless clicking of the element and checking if it was right element in right state after click?

1) For example, if I call checkbox.setSelected(true) I expect that  checkbox.isSelected() will return true after the call. But actual code in setSelected method does not checks the state of the element to determine if it's already selected and shouldn't be clicked. Definitely I don't want to be in a situation when checkbox is already selected and call checkbox.setSelected(true) sets selection to false.

rosolko:
@jane-ryabchenko, @asolntsev, @BorisOsipov I my point of view this case mix functionality with verification.
I think the real use case will we:
1. Set selected.
2. Make assert that check condition.
3. You decide check or not.
4. No performance issues.
Are you agree?
So If yes - we can close this PR and check such cases on your side.

BorisOsipov:
First, I don't understand what changes are proposed to this PR - diff is so big and full of garbage. I guess it makes sense to create a new PR with corresponding commits.

Second, I agree with @SergeyPirogov "Well, guys with modern UI frameworks it's all useless!" - it is the sad truth. In modern UI frameworks you hardly ever deal with standard html select.

rosolko:
Close it for now, if really needed submit a new PR cleared from not necessary files.

PR #602, opened 2017-09-18
Fix checkstyle suppressions on Windows

BorisOsipov:
On windows` ./gradle check` doesn't work correctly, we always get errors "Missing a Javadoc comment" 
It confuses people.


asolntsev:
@BorisOsipov Thank you! Merged.

PR #604, opened 2017-09-20
Update README about WD version compatibility

Scorpibear:
Update README to show actual information about compatibility with WebDriver version - add 3.0 instead of 2.0


BorisOsipov:
@Scorpibear Thank you!

PR #611, opened 2017-10-01
Update the project to selenium 3.6.0

ostap-oleksyn:
## Proposed changes
1. Updated project to use selenium 3.6.0. In this version the htmlunit dependency was removed from selenium-java. I removed the transitive property from htmlunit in order for the project to compile(I'm really not sure if this is the right way to do it, as I haven't worked with gradle before).
2. Since version 3.6.0 the WebDriver constructor that accepts Capabilities object is deprecated and will be removed in the future. The recommended way of creating the webdriver is using browser specific options(e.g. FirefoxOptions, ChromeOptions etc.). I replaced the Capabilities with browser specific option classes where possible.
4. Replaced the deprecated Locatable interface with recommended org.openqa.selenium.interactions.internal.Locatable
5. Fixed EdgeDriverFactory. For some reason it was creating a InternetExplorerDriver instead of EdgeDriver.
6. Removed killBrowser() from WebDriverThreadLocalContainer. The Killable interface is removed from selenium 3.6. Not really sure what this method did before, as the Killable interface wasn't implemented by any class.
7.  Since version 3.6 there is a possibility to start chrome and firefox in a headless mode using browser options(which makes htmlunit kinda obsolete). I added headless property to Configuration class to be able to use that in selenide.
8. Fixed some tests accordingly with he new changes.

## Checklist
- [-] I have added tests that prove my fix is effective or that my feature works(The only feature that I added is the headless configuration property. Not really sure how to test it.)

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -42,7 +42,7 @@ configurations {
 dependencies {
   compile group: 'io.github.bonigarcia', name: 'webdrivermanager', version: '1.7.2'
   compile('org.apache.httpcomponents:httpcore:4.4.6')
-  compile('org.seleniumhq.selenium:selenium-java:3.5.3') {
+  compile('org.seleniumhq.selenium:selenium-java:3.6.0') {
     exclude group: 'org.seleniumhq.selenium', module: 'selenium-android-driver'
     exclude group: 'org.seleniumhq.selenium', module: 'selenium-iphone-driver'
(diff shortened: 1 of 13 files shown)


asolntsev:
@ostap-oleksyn Thank you!

Next time please make independent changes in a separate pull request (like headless mode in this case). Small PRs are easier to review and merge.

PR #621, opened 2017-10-14
Make marionette the default firefox driver implementation.

ostap-oleksyn:
## Proposed changes
Since Selenium 3.0, marionette is the default firefox driver implementation. I think we should also move into that direction. What is done:
1. Now, in order to use Firefox 48+, you need to specify "firefox" as browser parameter, instead of "marionette"(which I never really liked, since it can be a bit confusing for newcomers).
2. To use the legacy Firefox (47 and lower) you need to pass "legacy_firefox" as browser parameter.
3. Marionette Firefox is now the default browser, if you don't specify other browser parameter.
4. Additionally I changed the opera driver creation, not sure why it used createInstanceOf() method instead of just returning new OperaDriver object.

## Checklist


BorisOsipov:
Hi @ostap-oleksyn 
Thank you for changes.

Have look
https://github.com/codeborne/selenide/blob/01774b9cc8cbed18ffa4292f169cdf6718961802/src/main/java/com/codeborne/selenide/webdriver/RemoteDriverFactory.java#L31-L33

In this place remote driver creates and capabilities.setBrowserName sets from Configuration.browser.
We can't pass "legacy_firefox" as broser name here. Correct me if I'm wrong.
Could you please enhance your changes?

Also may be it makes sense find other usages of Configuration.browser and go through them carefully.

ostap-oleksyn:
@BorisOsipov thanks for pointing that out. 
Fixed that and made the code a bit safer for other browsers as well. Selenide supports following browser parameters:  ie, edge, opera. But those are not valid selenium browser names when using grid, they should be: internet explorer(was already fixed in code), MicrosoftEdge and operablink. So I added a little if block that replaces those parameters with valid ones based on passed browser parameter. Checked other Configuration.browser usages, and everything seems fine.
Also added safari, edge and jbrowser to the browser parameter description.

asolntsev:
@ostap-oleksyn Merged, thank you!

Issue #623, opened 2017-10-16
ElementsCollection#first(int) & last(int) methods to get specified number of collection elements


yashaka:
Sometimes at some applications where we can't control/seed the database, we have to create new objects in a list and check them only in the head or in the tail of the list, not the full list. 

GIVEN
```java
$$(“.item”).shouldHave(texts(“a”, “b”, “c”, “d”, “e”, “f”));
```

THEN the following methods would be helpful:

```java
$$(“.item”).first(3).shouldHave(texts(“a”, “b”, “c”));
$$(“.item”).last(3).shouldHave(texts(“d”, “e”, “f”));
```

asolntsev:
Implemented in https://github.com/codeborne/selenide/pull/624, will be released in Selenide 4.9

yashaka:
The feature is not implemented correctly. It also lacks test coverage in context of "waiting behaviour" and "error messages". 

@asolntsev, please reopen.

asolntsev:
@yashaka Can you remind what exactly was implemented incorrectly? 

P.S. I added missing tests in a separate PR: https://github.com/codeborne/selenide/pull/779

asolntsev:
@yashaka Can you remind what exactly was implemented incorrectly?

yashaka:
oh... was so far in the past... I can hardly recall the point... 

Also, seems like I was not detailed enough in my last message... Sorry for that:(

PR #624, opened 2017-10-22
#623: Add methods to get first and last n elements of the collection

ostap-oleksyn:
## Proposed changes
As per this [request](https://github.com/codeborne/selenide/issues/623), I added first() and last() methods to get the first and last n elements of the elements list. 

## Checklist
- [x ] Checkstyle and unit tests pass locally with my changes by running `gradle check chrome htmlunit` command
- [x ] I have added tests that prove my fix is effective or that my feature works
- [x ] I have added necessary documentation (if appropriate)


src/test/java/integration/CollectionMethodsTest.java
```diff
@@ -209,19 +212,53 @@ public void canIterateCollection_withIterator() {
...
+            .collect(Collectors.toList());
+
+    Assert.assertEquals(regularSublist, selenideSublist);
+  }
+
+  @Test
+  public void canGetLastNElements() {
+    ElementsCollection collection =  $$x("//select[@name='domain']/option");
+    collection.last(2).shouldHaveSize(2);
+    collection.last(10).shouldHaveSize(collection.size());
```
yashaka: The expected result as a part of test logic should be explicit. 
I would change the corresponding line to:
collection.last(10).shouldHaveSize(10);

src/test/java/integration/CollectionMethodsTest.java
```diff
@@ -209,19 +212,53 @@ public void canIterateCollection_withIterator() {
...
+
+    List<String> regularSublist = $$x("//select[@name='domain']/option").stream()
+            .map(SelenideElement::getText)
+            .collect(Collectors.toList()).subList(2, collection.size());
+
+    List<String> selenideSublist = collection.last(2).stream()
+            .map(SelenideElement::getText)
+            .collect(Collectors.toList());
+
+    Assert.assertEquals(regularSublist, selenideSublist);
```
yashaka: Yet the test logic is not explicit in the test... I recommend to explicitly verify what texts should be in the sub-collection.

yashaka: In addition - just a few tests are not enough. It does make sense to check error messages for such "sub-collections"

src/main/java/com/codeborne/selenide/ElementsCollection.java
```diff
@@ -261,6 +261,26 @@ public SelenideElement last() {
...
   }
 
+  /**
+   * returns the first n elements of the collection
+   * @param elements - number of elements
+   * @return
+   */
+  public ElementsCollection first(int elements) {
+    List<WebElement> sublist = getActualElements().subList(0, Math.min(elements, size()));
+    return new ElementsCollection(new WebElementsCollectionWrapper(sublist));
```
yashaka: This implementation is not correct. It's not "reactive" enough.

GIVEN "1", "2", "c", "d" texts in the '.item' list during first 1 second
WHEN texts change to "a", "b", "c", "d" during next 5 seconds
THEN the code `$$(".item").first(2).shouldHave(texts("a", "b"))` still will fail on timeout of 6 seconds
BUT it should pass. 

This test by the way should also be added to the test suite. 

The correct way to implement these methods is the same as for "filterBy" method.

src/test/java/integration/CollectionMethodsTest.java
ostap-oleksyn: The size of the collection in this case is 4. If you try to get elements that are out of array boundaries the method will just return the full collection instead of throwing an error. That's what that line is testing.

src/test/java/integration/CollectionMethodsTest.java
ostap-oleksyn: What error messages do you have in mind?

src/main/java/com/codeborne/selenide/ElementsCollection.java
ostap-oleksyn: I'm not sure this error is caused by my changes. It looks like the .shouldHave(texts()) is not working properly for collections. Using your example, lets say we have one element with text "1" which changes text to "a" after 5 seconds. If you call $$(".item").shouldHave(texts("a")) without my first() method it will still fail after 6 seconds. Even if you set the Configuration.timeout to any other value that line still fails after 6 seconds timeout.

src/test/java/integration/CollectionMethodsTest.java
yashaka: ah, good, then you should put 4 explicitly in the assert, not collection.size()

src/test/java/integration/CollectionMethodsTest.java
yashaka: In Selenide we have kind of "informative error messages" in cases when the "should" method failed, or some operation on collection element failed. In such cases the message should be really informative. The user should understand from the error message what was the collection that he tried to deal with. 

Just run:

$$(".item").last(3).shouldHaveSize(4)
And check the error message. It should be informative. But probably it will not:)
And we should have tests for such cases too.

src/main/java/com/codeborne/selenide/ElementsCollection.java
yashaka: It was not about waiting 6 seconds. It's about waiting at most 6 seconds and succeed to find elements if they appear or change. Your implementation is not dynamic, it will not succeed. It will remember the exact number of elements at the very beginning - and will not update them after they change during the period of 6 seconds or whatever is set in the Configuration.collectionTimeout

Nevertheless - write a test and check it:)

Just a few tests is not enough for good test coverage of a new feature ;)

You checked only the "subListing" in context of counting size. 
But selenide is much more - it is also:
- dynamic elements that are updated when needed
- waiting elements
- informative error messages

I believe we should ensure that all those features are supported by the implementation of new first/last methods, and so we should have tests for them

PR #625, opened 2017-10-24
Actions before and after selenide commands, also, when command was executed with error.

kkochetov:
Here impemened some kind of before/after action hooks(requested https://github.com/codeborne/selenide/issues/447)

For this just implement HookAction interface. Then register your action with
`HookActions.getInstance().addBeforeAction(String name, HookAction action)`
This actions will be executed before executing selenide command.

or
`HookActions.getInstance().addAfterAction(String name, HookAction action)`
This actions will be executed after successfully finished selenide command i.e without any exceptions.

or
`HookActions.getInstance().addErrorAction(String name, HookAction action)`
This actions will be executed after falied selenide command right before throwing exception.

Also i added demo and tests for new feature.
And add empty line in the end of src/test/java/com/codeborne/selenide/webdriver/RemoteDriverFactoryTest.java 
About that code style checker was complained.

### DIFF
--- a/src/main/java/com/codeborne/selenide/hookactions/HookAction.java
+++ b/src/main/java/com/codeborne/selenide/hookactions/HookAction.java
@@ -1,0 +1,25 @@
+package com.codeborne.selenide.hookactions;
+
+import org.openqa.selenium.WebElement;
+
+public interface HookAction {
+  /**
+   * Condition for activation
+   *
+   * @param element     WebElement
+   * @param methodName  Name of method, who called hook
+   * @param args        Called command arguments
+   * @return boolean
+   *
+   */
+  boolean conditionForAction(WebElement element, String methodName, Object... args);
+
+  /**
+   * Action, if conditionForAction is true
+   *
+   * @param element     WebElement
+   * @param methodName  Name of method, who called hook
+   * @param args        Called command arguments
+   */
+  void action(WebElement element, String methodName, Object... args);
+}
--- a/src/main/java/com/codeborne/selenide/hookactions/HookAction.java
+++ b/src/main/java/com/codeborne/selenide/hookactions/HookAction.java
@@ -12,10 +12,10 @@ public interface HookAction {
    * @return boolean
    *
    */
-  boolean conditionForAction(WebElement element, String methodName, Object... args);
+  boolean isActive(WebElement element, String methodName, Object... args);
 
   /**
-   * Action, if conditionForAction is true
+   * Action, if isActive is true
    *
    * @param element     WebElement
    * @param methodName  Name of method, who called hook
--- a/src/main/java/com/codeborne/selenide/hookactions/HookActions.java
+++ b/src/main/java/com/codeborne/selenide/hookactions/HookActions.java
@@ -1,0 +1,116 @@
+package com.codeborne.selenide.hookactions;
+
+import org.openqa.selenium.WebElement;
+
+import java.util.Collection;
+import java.util.HashMap;
+
+public class HookActions {
+  private static ThreadLocal<HookActions> instance = new ThreadLocal<>();
+  private HashMap<String, HookAction> beforeActions = new HashMap<>();
+  private HashMap<String, HookAction> afterActions = new HashMap<>();
+  private HashMap<String, HookAction> errorActions = new HashMap<>();
+
+  private HookActions() {
+  }
+
+  /**
+   *  Return instance of HookActions for current thread.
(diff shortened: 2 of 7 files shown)


src/main/java/com/codeborne/selenide/hookactions/HookActions.java
```diff
@@ -0,0 +1,116 @@
...
+      instance.get().afterActions.remove(name);
+    if (instance.get().errorActions.containsKey(name))
+      instance.get().errorActions.remove(name);
+  }
+
+  private static HookActions instance() {
+    return instance.get();
+  }
+
+  private void preform(WebElement element, String methodName, Collection<HookAction> actions, Object... args) {
```
BorisOsipov: @kkochetov preform => perform?

BorisOsipov: Why Collection<HookAction> actions is a parameter?
In all methods `after\beforePreform(... )` you need to pass additional parameter in such a case...

src/main/java/com/codeborne/selenide/hookactions/HookActions.java
```diff
@@ -0,0 +1,116 @@
...
+    instance.get().errorActions.put(name, action);
+  }
+
+  /**
+   * Remove action from HookActions
+   * Try to remove from before, after and error actions lists.
+   *
+   * @param name    Name of hook to remove
+   */
+  public void removeAction(String name) {
```
BorisOsipov: Maybe better to have three methods: removeBefore\after\error action?

BorisOsipov: Instead of if's

src/main/java/com/codeborne/selenide/hookactions/HookActions.java
```diff
@@ -0,0 +1,116 @@
+package com.codeborne.selenide.hookactions;
+
+import org.openqa.selenium.WebElement;
+
+import java.util.Collection;
+import java.util.HashMap;
+
+public class HookActions {
+  private static ThreadLocal<HookActions> instance = new ThreadLocal<>();
+  private HashMap<String, HookAction> beforeActions = new HashMap<>();
```
BorisOsipov: ConcurrentHashMap?

src/main/java/com/codeborne/selenide/hookactions/HookAction.java
```diff
@@ -0,0 +1,25 @@
...
+  /**
+   * Condition for activation
+   *
+   * @param element     WebElement
+   * @param methodName  Name of method, who called hook
+   * @param args        Called command arguments
+   * @return boolean
+   *
+   */
+  boolean conditionForAction(WebElement element, String methodName, Object... args);
```
BorisOsipov: Maybe instead of conditionForAction => isEnabled() or isActive() will be better?

src/main/java/com/codeborne/selenide/hookactions/HookActions.java
kkochetov: I just checked this case with
```
HookActions.getInstance().addBeforeAction("testHook", new TestAction("thread1", true));
new Thread(() -> {
  HookActions.getInstance().addBeforeAction("testHook", new TestAction("thread2", true));
  HookActions.getInstance().beforePerform(null, "command");
}).start();
Thread.sleep(500);
HookActions.getInstance().beforePerform(null, "command");
```
everything was fine.
ActionHooks with the same names was added and executed correctly.
one time per thread.

src/main/java/com/codeborne/selenide/hookactions/HookAction.java
kkochetov: fixed in last commit

src/main/java/com/codeborne/selenide/hookactions/HookActions.java
kkochetov: made three different methods. 
also i desided to remove ifs. I think, it will be correct to throw exception when we trying to remove nonexistent element of collection.

BorisOsipov:
I am not sure that is possible to merge in current implementation.
I would like to see such interceptors in kind of extension of [LogEventListener](https://github.com/codeborne/selenide/blob/master/src/main/java/com/codeborne/selenide/logevents/LogEventListener.java) extension to support custom event listeners.
Also first please desribe a way what you are going to extend it to discuss a corresponding API

PR #626, opened 2017-10-24
Check size of collection before get()

kkochetov:
Small fix, only adds checking of size before trying to get element.
If the collection is empty, selenide no longer break the test with
`java.lang.IndexOutOfBoundsException: Index: 0, Size: 0` 
but will be correctly fail it with
```
Method threw 'com.codeborne.selenide.ex.ListSizeMismatch' exception.
: expected: = 1, actual: 0, collection: div.zxcv
Elements: []
```


asolntsev:
@kkochetov Please look at failing tests. 
For example, `MethodCalledOnElementFailsOnTest. actionWithVisibilityWaiting_WhenCollectionElementByIndex_WithIndexOutOfRange`. 

For collection of 2 elements, the call `$$("ul li").get(10).click();` caused 

```java
Element not found {ul li[10]}
Caused by: java.lang.IndexOutOfBoundsException: Index: 10, Size: 2
```

and after your change, it causes 

```java
ListSizeMismatch : expected: > 10, actual: 2, collection: ul li
```

Probably the new error message is even better, but you need to update tests.

kkochetov:
tests fixed.

CaBocuk:
This request conflicts with my one - #653 where I re-evaluate actual elements if index is out of bounds to achieve better support of slowly-loading collections. 
BTW should*(...) and waitUntil(...) methods don't do any re-evaluation of actual elements. And in this case test will fail immediately.
So do we need to make such assertion here?

kkochetov:
@CaBocuk I will check it as soon as possible.
But, for the first look your PR solves my problem.
Maybe I need to close this pr.

asolntsev:
@kkochetov @CaBocuk So, can we close this PR?

kkochetov:
closed because duplicate https://github.com/codeborne/selenide/pull/653

PR #646, opened 2017-12-18
Method to check if page is scrolled to the bottom

pavelpp:
## Proposed changes
I found it quite useful in my tests with lazy loading of elements to check if the end of the page is reached. 

## Checklist

### DIFF
--- a/src/main/java/com/codeborne/selenide/Selenide.java
+++ b/src/main/java/com/codeborne/selenide/Selenide.java
@@ -837,7 +837,7 @@ public class Selenide {
    * Useful if you need to scroll down by x pixels unknown number of times.
    */
   public static boolean atBottom() {
-    return Selenide.executeJavaScript("return window.scrollY + window.innerHeight >= document.body.scrollHeight");
+    return Selenide.executeJavaScript("return window.pageYOffset + window.innerHeight >= document.body.scrollHeight");
   }
 
   private static <T> List<String> listToString(List<T> objects) {
--- a/src/main/java/com/codeborne/selenide/Selenide.java
+++ b/src/main/java/com/codeborne/selenide/Selenide.java
@@ -831,14 +831,14 @@ public class Selenide {
     }
   }
 
-    /**
-     * Return true if bottom of the page is reached
-     *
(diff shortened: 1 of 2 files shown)


asolntsev:
@pavelpp Can you describe how are going to use this method? What is the real case?
Probably it's better to add method `Selenide.shouldBe(atBottom)` or `Selenide.scrollTo(bottom)`?

pavelpp:
@asolntsev shouldBe methods would also be helpful for some cases, but my use case is completely different. One thing is expecting that after a certain action you are at the bottom of the page, and another checking state to decide what to do next. My PR is about the latter. I'm using it a lot with lazy loading when I need to perform some action on the elements which are being loaded into dom while I scroll the page, so essentially it helps with looping 
```
while(!atBottom()){
scrollPixels(200,0);
//do something
}
```

BorisOsipov:
I am not sure about such changes..
I think it isn't good idea to place such js scripts to `Selenide.java`.
Yeah it is good script(especially if it works for chrome\firefox and other browsers as well ) and probable I will use it, but in very specific cases...

pavelpp:
@BorisOsipov , I was looking for a better place for this method, and couldn't find anything better. Perhaps we could add a new utility class, something like ScrollActions or similar?

BorisOsipov:
@pavelpp  Yeah it make sense. @asolntsev what do yo think?

asolntsev:
@pavelpp @BorisOsipov Yes, probably `ScrollActions` is a good name. 

I was also thinking that some of existing static methods in class `Selenide` are also too specific, and (in theory) should also be moved to dedicated classes:

* `confirm()`
* `dismiss()`
* `prompt()`
* `onConfirmReturn()`
* `getJavascriptErrors()`
* `getWebDriverLogs()`
* `getSelectedRadio()`
* `getUserAgent()`
* `zoom()`

Should we have a single class for all of them or many different small classes?

BorisOsipov:
>Should we have a single class for all of them or many different small classes?

I think yes, it will be better to separate them.

pavelpp:
@asolntsev yes, it makes sense, though, I think it shouldn't be part of this PR

PR #652, opened 2017-12-24
fixed problem, when fastSetValue didn't trigger blur on elements

vinogradoff:


vinogradoff:
How the heck is coverage decreased, if the file where it happened is not the part of the commit...

PR #659, opened 2018-01-02
Fix #655"Listeners SoftAsserts: Screenshot: null for FAIL test"

### LINKED ISSUE #655: Listeners  SoftAsserts:   Screenshot: null   for FAIL test
Listeners  SoftAsserts:   Screenshot: null   for FAIL
Selenide: 4.8
TestNG: 6.11

image

BorisOsipov:
## Proposed changes
Fix https://github.com/codeborne/selenide/issues/655 issue

## Checklist

### DIFF
--- a/src/main/java/com/codeborne/selenide/ElementsCollection.java
+++ b/src/main/java/com/codeborne/selenide/ElementsCollection.java
@@ -62,7 +62,7 @@ public class ElementsCollection extends AbstractList<SelenideElement> {
       return this;
     }
     catch (Error error) {
-      SelenideLogger.commitStep(log, error);
+      SelenideLogger.commitStep(log, UIAssertionError.wrap(error, collectionsTimeout));
       switch (assertionMode) {
         case SOFT:
           return this;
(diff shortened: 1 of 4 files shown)


BorisOsipov:
It again complains about `WebDriverThreadLocalContainer.java` coverage decrease but it was not affected in the changes..

BorisOsipov:
Yes it is. But some times it reports unexpected coverage changes here https://github.com/codeborne/selenide/blob/0f2e4502f85b3b46184373536309b2da0db956cb/src/main/java/com/codeborne/selenide/impl/WebDriverThreadLocalContainer.java#L41

and we can't fix it :)
Also I checked Codeconv documentation. It cannot exclude specific lines from coverage reports.

rosolko:
@BorisOsipov Can you please try to replace foreach with stream if reduce this `if`:
```
ALL_WEB_DRIVERS_THREADS.stream()
  .filter(thread -> !thread.isAlive())
  .forEach(thread -> {
    log.info("Thread " + thread.getId() + " is dead. Let's close webdriver " + THREAD_WEB_DRIVER.get(thread.getId()));
    closeWebDriver(thread);
  });
``` 
Possible it can help avoid false codecov trigger.

BorisOsipov:
@rosolko Let's make a separate issue about code coverage flashing and test your proposal?
I don't think such changes directly reference to this PR changes.

Issue #675, opened 2018-01-16
Chrome does not start maximized


tbgdn:
## The problem
Chrome browser does not start maximized on Windows with latest Selenide and Selenium.

## Details
Even if Configuration.startMaximized is enabled by default, Chrome does not start maximized and the logs show the following exception:
`Cannot maximize ChromeDriver: java.awt.HeadlessException`
It seems that Chrome browser is adjusted by using the screen resolution provided by Java toolkit, which fails and no retry mechanism is in place. Apparently, the adjusting method used on other browsers also works on Chrome and maybe this can be used in the catch clause.

## Tell us about your environment
* **Selenide Version:** 4.9.1
* **Chrome Version:** 55 and 63
* **Browser Driver Version:** 2.35
* **Selenium Version:** 3.8.1
* **OS Version:** Windows 7 and 10

## Code To Reproduce Issue [ Good To Have ]
Any code will do.
```
Configuration.browser = "chrome";
Selenide.open("http://www.google.com");
```

BorisOsipov:
Need enhance logic this https://github.com/codeborne/selenide/blob/master/src/main/java/com/codeborne/selenide/webdriver/WebDriverFactory.java#L71
catch exception, log them and don't change browser size.

BorisOsipov:
@tbgdn Thanks for reporting this issue

SergeyPirogov:
As I remember Andrey wanted to delete maximize. The motivation is that users should implicitly set browser window

jpratt2:
I cannot duplicate this issue. In Windows 10 and Windows 7, the Chrome browser maximizes to fill the screen. It appears this issue can be marked complete.

asolntsev:
@jpratt2 did you try it in headless mode?

leysanka:
Crashes for me as well with the following error: "Chrome Automation extension has crashed. Click this balloon to reload the extension"

asolntsev:
Yes, `--start-maximized` never worked for me. Instead, you can use Selenide option `Configuration.startMaximized = true;`

But I personally recommend not to start browser maximized, but instead set the browser size explicitly.
It helps you to avoid flaky tests.

Neboyarsky:
> @leysanka Yes, `--start-maximized` never worked for me. Instead, you can use Selenide option `Configuration.startMaximized = true;`
[...]

PR #556, opened 2017-06-18
Browser capabilities configuration section added

SergeyPirogov:
Need to add merge for chrome and firefox capabilities !!!

PR #680, opened 2018-01-18
#675: Chrome browser gets maximized

tbgdn:
## Proposed changes
Fix for #675 issue.

## Checklist

### DIFF
--- a/src/main/java/com/codeborne/selenide/webdriver/WebDriverFactory.java
+++ b/src/main/java/com/codeborne/selenide/webdriver/WebDriverFactory.java
@@ -68,16 +68,24 @@ public class WebDriverFactory {
       driver.manage().window().setSize(new org.openqa.selenium.Dimension(width, height));
     } else if (startMaximized) {
       try {
-        if (isChrome()) {
-          maximizeChromeBrowser(driver.manage().window());
-        } else {
-          driver.manage().window().maximize();
-        }
+        maximizeBrowser(driver);
       } catch (Exception cannotMaximize) {
         log.warning("Cannot maximize " + describe(driver) + ": " + cannotMaximize);
       }
     }
     return driver;
+  }
+
+  private void maximizeBrowser(WebDriver driver) {
+    try {
+      driver.manage().window().maximize();
+    } catch (Exception cannotMaximize) {
+      if (isChrome()) {
+        maximizeChromeBrowser(driver.manage().window());
+      } else {
+        throw cannotMaximize;
+      }
+    }
   }
 
   protected void maximizeChromeBrowser(WebDriver.Window window) {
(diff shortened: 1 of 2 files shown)


user667:
Pretty sure this solution will not solve the issue I'm seeing at the moment:
```bash
WARNING: Cannot maximize RemoteWebDriver: org.openqa.selenium.WebDriverException:
    unknown error: failed to change window state to maximized, current state is normal
```
Seems related to [SeleniumHQ/docker-selenium#559](https://github.com/SeleniumHQ/docker-selenium/issues/559).

Possible solution is using a DesiredCapabilities profile for Chrome.

BorisOsipov:
>Pretty sure this solution will not solve the issue I'm seeing at the moment:

Unquestionably! And I know why - this PR for issue https://github.com/codeborne/selenide/issues/675 and it doesn't try to solve problem you are ran into.

user667:
Yes indeed. I only quickly scanned through the issues - I'm new to Selenide. Would it make sense to create an issue for that problem?

BorisOsipov:
@user667 I think if your run chrome in docker you have to specify window size explicitly.

asolntsev:
@BorisOsipov @user667 I strongly recommend to **always** set browser size explicitly and never use `startMaximized`.

asolntsev:
@tbgdn Thank you!
However, `driver.manage().window().maximize()` doesn't work for Chrome. At least on my MacBook, it doesn't throw exceptions but also doesn't maximize the Chrome window.

user667:
Thanks for the hints @BorisOsipov and @asolntsev and @tbgdn for the PR. I'm sure we can bypass the "issue" for the time being.

However I wanted to try the recommended workaround by passing custom desired capabilities but as far as I understand I'm depending on [Browser capabilities configuration section added #556](https://github.com/codeborne/selenide/pull/556). Therefore I'll try version 4.10 or 4.10.01 asap.

```java
ChromeOptions options = new ChromeOptions();
options.addArguments("--start-maximized");
DesiredCapabilities capabilities = DesiredCapabilities.chrome();
capabilities.setCapability(ChromeOptions.CAPABILITY, options);
```
See https://stackoverflow.com/a/9942198

asolntsev:
@user667 What should we do with this PR?
As I understand, it will not help because `driver.manage().window().maximize()` doesn't maximize the Chrome, and doesn't throw any exceptions. Thus method `maximizeBrowser(WebDriver driver)` will never maximize chrome window. 

Should we close this PR?

BorisOsipov:
@asolntsev @tbgdn yes we should.
See my comment in https://github.com/codeborne/selenide/pull/680#issuecomment-375850395

PR #684, opened 2018-01-27
bugfix for SetSelected (not it doesn't work with invisible, readonly,…

vinogradoff:
… disabled and not checkbox/option/radion)

### DIFF
--- a/src/main/java/com/codeborne/selenide/commands/SetSelected.java
+++ b/src/main/java/com/codeborne/selenide/commands/SetSelected.java
@@ -13,9 +13,23 @@ public class SetSelected implements Command<WebElement> {
   public WebElement execute(SelenideElement proxy, WebElementSource locator, Object[] args) {
     boolean selected = (Boolean) args[0];
     WebElement element = locator.getWebElement();
-    if (element.isSelected() ^ selected) {
-      if (element.getAttribute("readonly") != null)
-        throw new InvalidStateException("Cannot change value of readonly element");
+    if (!element.isDisplayed()) {
+      throw new InvalidStateException("Cannot change invisible element");
+    }
+    String tag = element.getTagName();
+    if (!tag.equals("options")) {
+      if (tag.equals("input")) {
+        String type = element.getAttribute("type");
+        if (!type.equals("checkbox") && !type.equals("radio")) {
+          throw new InvalidStateException("Only use setSelected on checkbox/option/radio");
+        }
+      } else {
+        throw new InvalidStateException("Only use setSelected on checkbox/option/radio");
+      }
+    }
+    if (element.getAttribute("readonly") != null || element.getAttribute("disabled") != null)
+      throw new InvalidStateException("Cannot change value of readonly/disabled element");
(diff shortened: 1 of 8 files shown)


src/main/java/com/codeborne/selenide/commands/SetSelected.java
```diff
@@ -13,9 +13,23 @@
...
     boolean selected = (Boolean) args[0];
     WebElement element = locator.getWebElement();
-    if (element.isSelected() ^ selected) {
-      if (element.getAttribute("readonly") != null)
-        throw new InvalidStateException("Cannot change value of readonly element");
+    if (!element.isDisplayed()) {
+      throw new InvalidStateException("Cannot change invisible element");
+    }
+    String tag = element.getTagName();
+    if (!tag.equals("options")) {
```
asolntsev: why `options`? There is no such html tag.
And you haven't added any tests with `options`.

src/main/java/com/codeborne/selenide/commands/SetSelected.java
```diff
@@ -13,9 +13,23 @@
   public WebElement execute(SelenideElement proxy, WebElementSource locator, Object[] args) {
     boolean selected = (Boolean) args[0];
     WebElement element = locator.getWebElement();
-    if (element.isSelected() ^ selected) {
-      if (element.getAttribute("readonly") != null)
-        throw new InvalidStateException("Cannot change value of readonly element");
+    if (!element.isDisplayed()) {
+      throw new InvalidStateException("Cannot change invisible element");
+    }
```
asolntsev: This is a valid change, but there is 1 thing I don't like. This code will call too many `WebElement` method, which makes it slow. `isDisplayed`, `getTagName`, `getAttribute("type")`, `getAttribute("readonly")`, `element.getAttribute("disabled")`. 

Probably we should implement it with one JS call?

src/main/java/com/codeborne/selenide/commands/SetSelected.java
vinogradoff: right you are, thanx!

src/main/java/com/codeborne/selenide/commands/SetSelected.java
BorisOsipov: Btw why `element.isDisplayed()` check needs here? Doesn't Selenide check it automaticaly whatever?

vinogradoff:
No) Checked it, in this place it would treat invisible successfuly (but do
nothing about it)

Viele Grüße

Alexei Vinogradov

src/main/java/com/codeborne/selenide/commands/SetSelected.java
vinogradoff: 1. options->option fixed, Test added.
2. many getAttribute is actually not slow, because it just doesn't search at every call just once.
3. element.isDisplayed is needed, selenide did not check it at *this* place (not_existing_element.setSelected(false); -> works without error, which is wrong).

vinogradoff:
@asolntsev please review

PR #686, opened 2018-01-29
Bug fix for occasional NPE's in SelenideReport 

dkorobtsov:
## Proposed changes
Bug fix for occasional NPE's in SelenideReport when used with TestNg listeners.
On certain occasions - when SimpleReport.start() method was not called for any reason, SelenideReport.finish() was causing NPE.

## Checklist
- [+] Checkstyle and unit tests pass locally with my changes by running `gradle check chrome htmlunit` command
- [+ ] I have added tests that prove my fix is effective or that my feature works
- [n/a] I have added necessary documentation (if appropriate)

### DIFF
--- a/src/main/java/com/codeborne/selenide/logevents/SimpleReport.java
+++ b/src/main/java/com/codeborne/selenide/logevents/SimpleReport.java
@@ -21,10 +21,10 @@ public class SimpleReport {
   public void finish(String title) {
     EventsCollector logEventListener = SelenideLogger.removeListener("simpleReport");
 
-      if (logEventListener == null) {
-          log.warning("Can not publish report because Selenide logger has not started.");
-          return;
-      }
+    if (logEventListener == null) {
+      log.warning("Can not publish report because Selenide logger has not started.");
+      return;
+    }
 
     OptionalInt maxLineLength = logEventListener.events()
             .stream()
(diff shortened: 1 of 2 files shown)


src/main/java/com/codeborne/selenide/logevents/SimpleReport.java
```diff
@@ -21,6 +21,11 @@ public void start() {
   public void finish(String title) {
     EventsCollector logEventListener = SelenideLogger.removeListener("simpleReport");
 
+    if (logEventListener == null) {
```
vinogradoff: it is actually a hack. It certainly breaks nothing, but actual question is, why/when does removeListener returns null?
Root cause is not addressed with this fix....
is the problem reproducible somehow?

dkorobtsov: Agreed. You are right, it does not solve root cause. Made some research, it feels like in certain circumstances we have conflict between log4j2 and SelenideLogger. At least, I was able to reproduce the issue only in case another logging library was used at the same time.

Here is a gist with steps to reproduce, screenshots and code snippets to reproduce the issue:
https://gist.github.com/dkorobtsov/0d121bcb11a3c421e0dbc7ea93f0c8bc

BorisOsipov: @vinogradoff I don't want to have NPE in Selenide code in any cases of user mistakes.

dkorobtsov: @BorisOsipov @vinogradoff btw, just in case I've updated gist with another code sample and added some comments. 

TLDR: context is null in case test was removed from suite scope for any reason on before invocation event before SelenideLogger.start() was called. Since afterInvocantion is called anyway, finish() is called with null context. In theory cheap and fast solution all similar situations is to add boolean flag like "selenideLoggingStarted" - so finish() event will be called only in case logging has actually started.

PR #687, opened 2018-02-01
Browser position

rosolko:
## Proposed changes

Add ability to set browser window position.

## Checklist

### DIFF
--- a/src/test/java/integration/BrowserPositionTest.java
+++ b/src/test/java/integration/BrowserPositionTest.java
@@ -1,0 +1,47 @@
+package integration;
+
+import com.codeborne.selenide.Configuration;
+import com.codeborne.selenide.WebDriverRunner;
+import org.junit.Before;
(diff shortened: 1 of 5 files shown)


src/test/java/integration/BrowserPositionTest.java
```diff
@@ -0,0 +1,56 @@
...
+import static org.junit.Assert.assertNotEquals;
+import static org.junit.Assume.assumeFalse;
+
+public class BrowserPositionTest extends IntegrationTest {
+  String position;
+  Point point;
+
+  @Before
+  public void prepareTestData() {
+    Configuration.browser = WebDriverRunner.CHROME;
```
BorisOsipov: why only chrome? what if I would run test against firefox?

rosolko: my bad, commit test configuration. fixed.

PR #705, opened 2018-03-22
Make screenshot of SelenideElement/WebElement which is inside iframe

andrejska:
## Proposed changes
New feature: added possibility to make a screenshot of SelenideElement or WebElement which is inside iframe

## Checklist


rosolko:
@andrejska For first why don't you cover this feature with tests?
For second wht don't you reuse exist method for `takeScreenshotAsImage(WebElement element)` with minimal extend for frame switch?

rosolko:
@andrejska ping!

andrejska:
@rosolko 
>For first why don't you cover this feature with tests?
Tests added
>For second wht don't you reuse exist method for takeScreenshotAsImage(WebElement element) with minimal extend for frame switch?
Difference between taking screenshot of element inside iframe and regular element screenshot is big, so reusability of this method will just add more complexity
>Also isn't it work for you just take screenshot of frame?
For my specific use case I need a screenshot of webelement inside iframe
>What behavior do you expect if iframe or element inside iframe doesn't exist?
Behavior implements, warning message and returning null, similar approach as in takeScreenshotAsImage

andrejska:
@rosolko can you please check?

andrejska:
@rosolko did changes according to your latest comment, added missing tests & semantic change in order of parameters, please check

BorisOsipov:
>Also I'm not sure about such complex method takeScreenshotAsImage that return BufferedImage, I'll take a look at them soon.
+1 It looks nasty.

andrejska:
@rosolko, @BorisOsipop what are next steps :)?

rosolko:
@andrejska Sorry for long response. I'll take a look at this PR today and if I can't decrease complexity of method fastly - I'll merge it.

asolntsev:
@andrejska @rosolko Thanks for the PR! Merged to Selenide 4.11.2

PR #715, opened 2018-04-19
Adding static analysis for avoiding start import for main package

andrejska:
## Proposed changes
Adding static analysis for avoiding start import for main package, as in my first contribution to Selenide (https://github.com/codeborne/selenide/pull/705) it was done by maintainer as it was not defined in static analysis tools configuration.
Also in affected files white spaces are trimmed (as defined in .editorconfig).
Test package right now is ignored for this check, if this is the way to go I can fix this issues also there and enable checking for this rule for test package.

## Checklist


andrejska:
@rosolko most probably this is for you to check :)

rosolko:
@asolntsev It's good java code style practice.

andrejska:
Thanks, this is good coding practice and one of Google coding guideline bulletpoints. 

Checked test package - there are much more wild card imports and most of them are static imports for which Intellij doesn't automatically replace them with single class imports. So adjusting code style for tests is not so straight forward :)

PR #717, opened 2018-04-23
Adding static analysis for avoiding wildcard imports for test package

andrejska:
Adding static analysis for avoiding wildcard imports for test package, fixing existing imports, applying .editorconfig style for affected files

## Proposed changes
Follow up of https://github.com/codeborne/selenide/pull/715, making all project fully compatible with google Java coding guidelines for wildcard imports.
These changes are done automatically by IntelliJ IDEA in scope of optimise imports feature, I just checked that it is done correctly and removed from change list optimisations which are out of scope of this change.

## Checklist


andrejska:
@BorisOsipov but how in other way you can apply automatic code style check if not fix existing problems? 
It is not code reformat, it is code style improvements (http://checkstyle.sourceforge.net/config_imports.html). Otherway this coding styles rule should be checked for correctness manually for affected files, so in some cases they will be missed and there will be different code styles rules for main and test package.

src/test/java/integration/SelenideMethodsTest.java
```diff
@@ -11,15 +11,55 @@
 import org.openqa.selenium.WebDriverException;
 import org.openqa.selenium.WebElement;
 
-import static com.codeborne.selenide.Condition.*;
```
BorisOsipov: @andrejska @rosolko 
I don't like such changes. It replace few starts import to at least 50 new lines.
I find it outrageous.

BorisOsipov:
@andrejska ok let's ask other guys

rosolko:
@andrejska @BorisOsipov, 
I agree with point that for test scope it isn't take a lot of profit as for main scope. 
I think we can skip this rule for tests.
So if you agree with feel free to close or merge this PR if not.

andrejska:
If you think that test package can be left without replacing wildcard imports then it is fine for me, I think this is one of most controversial Java coding guideline :)

Just to have both sides presented, will add some why this change is ok:
I think nowadays size of code is not so critical and modern IDE can fold imports, also discussion about Java wildcard imports which is almost 10 years old I still valid:
https://stackoverflow.com/questions/147454/why-is-using-a-wild-card-with-a-java-import-statement-bad 
Also same style for all code base I think is better then different rules for main and test packages :)


So feel free to do as you want for this project :)

rosolko:
So I'm close it for now, if we reconsider to style code w/o splitting to sources sets w/ selft configs we'll use this rule.
Thanks.

PR #721, opened 2018-04-26
Adding new Condition - moving

andrejska:
## Proposed changes
Adding new Condition - moving.
It is useful in cases when there is a moving animation and we want to click on something inside moving element. If element is moving click might not work, so we want to be sure that moving animation has finished.

## Checklist

### DIFF
--- a/src/main/java/com/codeborne/selenide/Condition.java
+++ b/src/main/java/com/codeborne/selenide/Condition.java
@@ -4,14 +4,12 @@ import com.codeborne.selenide.conditions.Text;
 import com.codeborne.selenide.impl.Describe;
 import com.codeborne.selenide.impl.Html;
 import com.google.common.base.Predicate;
-import com.google.common.util.concurrent.Uninterruptibles;
 import org.openqa.selenium.Point;
 import org.openqa.selenium.StaleElementReferenceException;
 import org.openqa.selenium.WebElement;
 import org.openqa.selenium.interactions.internal.Locatable;
 
 import static com.codeborne.selenide.Selenide.getFocusedElement;
-import static java.util.concurrent.TimeUnit.MILLISECONDS;
 
 /**
  * Conditions to match web elements: checks for visibility, text etc.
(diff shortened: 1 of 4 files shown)


src/main/java/com/codeborne/selenide/Condition.java
```diff
@@ -101,6 +105,26 @@ public boolean apply(WebElement element) {
...
+   * <p>Sample: <code>$("#loginLink").shouldBe(moving);</code></p>
+   */
+  public static final Condition moving = new Condition("moving") {
+    @Override
+    public boolean apply(WebElement element) {
+      if (!(element instanceof Locatable)) {
+        throw new RuntimeException("Provided WebElement is not Locatable, cannot understand if it moving or not");
+      }
+      Point initialLocation = ((Locatable) element).getCoordinates().inViewPort();
+      Uninterruptibles.sleepUninterruptibly(200, MILLISECONDS);
```
rosolko: What behavior should be if animation took more that 200 milliseconds?
For example, some portion of test data loaded by 1 second, and it is single page react application, where page loaded before content.
So, what should we expect and do?

andrejska: We are just checking if location of an element has changes during 200ms, if it has - then it means that element is moving, otherwise it is not

andrejska: For example we can check when snackbar (https://www.w3schools.com/howto/howto_js_snackbar.asp) has fully appeared (moving animation is finished), and if we want to click a link on it is a must to do, otherwise sometimes click might not work

rosolko: Btw, 200ms it's magic number that can't be proof of good application work.
Maybe it's good to introduce configuration, something like `animationWait` or something like that to configure this sleep.
I think we need more ideas from @asolntsev and @BorisOsipov

andrejska: yes, is it magic number, but it has some assumptions:
smaller interval of webdriver calls introduces significant performance issues, especially if some any kind of waiter is which calls this condition.
Bigger waiting times makes test slower, so for it looks like 200ms is an optimal solution, but maybe there are some use cases where it should be configured, dunno

BorisOsipov:
I think testing animations through selenium isn't good idea... So flaky...

@andrejska May I ask you show us example of what animation do you test?

src/main/java/com/codeborne/selenide/Condition.java
rosolko: If parameterize this condition I think it should be something like `$("#element").should(movesOver(200));`

MikeShysh:
Hi, 

1. What about adding to your application some test mode with zero animation duration.
2. What about extra attribute for animated element. For example, in my case (angular) elements have  attribute .ng-animation. If there is no such attribute, how about force devs to add it.

src/main/java/com/codeborne/selenide/Condition.java
andrejska: How about `moveAround`/`movingAround`? and what about not parametrised version with default value of 200?
```$("element").should(moveAround(200));
$("element").shouldBe(movingAround(200))```

andrejska:
@MikeShysh 
>What about adding to your application some test mode with zero animation duration.
if there are many animations with different length it might require some refactoring to add such testing feature for product

>What about extra attribute for animated element. For example, in my case (angular) elements have attribute .ng-animation. If there is no such attribute, how about force devs to add it.
Not all js-frameworks are providing such possibility as Angular.
https://www.w3schools.com/angular/tryit.asp?filename=try_ng_animation_css 
^ in angular you can understand state of animation by checking element's class

andrejska:
@rosolko @BorisOsipov I have renamed this Condition to moveAround and added optional parameter of 200, so that we can avoid using some magic numbers, please check :)

rosolko:
@andrejska Can you please also to add some integration tests as proof of feature?
Still not completely sure about this feature 😕

andrejska:
@rosolko Integration tests added

rosolko:
@andrejska I pondered this solution and thinks that it shouldn't be part of standard conditions list.
I propose to create wiki page with custom community conditions and collect them there.
So, basic conditions will live in selenide code base, and super environment specific conditions will live in wiki.
And of course any user can see this condition and apply to their project by copying it from wiki and place somewhere in their framework.
What do you think about this solution?

rosolko:
@andrejska ping!

andrejska:
@rosolko I agree that this is quite specific case and might not be part of Selenide, although approach with wiki will make this solution sooner or later outdated, but lets see how it goes ;)

rosolko:
https://github.com/codeborne/selenide/wiki/Community-conditions#moving

MikeShysh:
Hi,
Eventually, I faced the situation when the proposed solution suits well. The brilliant example is default css feature - keyframes animations. Such kind of animations affects only position of the element from DOM point of view.

PR #724, opened 2018-05-02
Stop making lots of threads

fyrewall77:
## Proposed changes
Every time a webdriver is closed, a new thread is created and quickly destroyed.  This pull request creates a single long-lived Executor (thread) instead.

### DIFF
--- a/src/main/java/com/codeborne/selenide/impl/WebDriverThreadLocalContainer.java
+++ b/src/main/java/com/codeborne/selenide/impl/WebDriverThreadLocalContainer.java
@@ -12,6 +12,7 @@ import org.openqa.selenium.WebDriverException;
 import org.openqa.selenium.remote.UnreachableBrowserException;
 import org.openqa.selenium.support.events.EventFiringWebDriver;
 import org.openqa.selenium.support.events.WebDriverEventListener;
+import org.apache.commons.lang3.concurrent.BasicThreadFactory;
 
 import java.util.ArrayList;
 import java.util.Collection;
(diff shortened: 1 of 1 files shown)


asolntsev:
@fyrewall77 What problem you are trying to solve? Yes, we spawn multiple threads, but they die quickly. I don't see a problem.

By the way, it was done with intention. There is a problem with IE that sometimes hangs on closing. That's  why we started spawning a separate thread for every webdriver: this way one hanging webdriver doesn't affect others.

fyrewall77:
I have a long running, highly multi-threaded application and it's always bothered me when I have huge threadIDs, because I want to ensure that I am handling exceptions on all threads properly.  Executors will typically start a new thread if an exception escapes and there are no exception handler registered.

That said, I now understand why you have this thread starting, and would like to submit another patch that is simply:

if(closeBrowserTimeoutMs <=0)
  //Dont start new thread and run close in the same thread
else
  //Start a new thread to close the Webdriver

Would that make sense to you?  It would solve my problem.

asolntsev:
I still don't understand what problem you are trying to solve. If I
understand correctly, you just feel bad because of many threads, but it's
not a real problem.

AlexanderPoleschuk:
@fyrewall77 when you  say "it has always bothered me when I have huge threadIDs, because **I want to ensure that I am handling exceptions on all threads properly.**" do you realize that narrowing down a number of threads will VICE VERSA confuse you? In case each test is run in a separate thread, you may track it from start till the end('cause Selenide tracks threadID in most cases).

Anyway, if you want to re-use the thread... why don't you want to create your own instance of WebDriver and manage closing it on your own? then it will use the same instance of browser, hence, the same thread.
But beware that for such case you'll have to manage LOTS of cookie/auth things, especially logging out the user before you start a new test, if your new own intends to login('cause it'll use the same browser profile until you initialize new browser instance)

rosolko:
This PR is obsolete.
From 5.0.0 version selenide has no `WebDriverThreadLocalContainer` impl.
Alive only in `v4` branch.

Issue #673, opened 2018-01-16
Selenide logs too much with marionette driver


vinogradoff:
see gist:

https://gist.github.com/barancev/d7d4b5be3f3f55243fb8627f457fe1f1 (thanx @barancev)

PR #732, opened 2018-05-08
Redirect tons of firefox logs to /dev/null

rosolko:
## Proposed changes
Generally this PR addressed to #673 
Linked with firefox [bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1399441) that still open
Selenium bug - https://github.com/SeleniumHQ/selenium/issues/2632
And geckodriver bug - https://github.com/mozilla/geckodriver/issues/619

## Checklist

### DIFF
--- a/src/main/java/com/codeborne/selenide/webdriver/FirefoxDriverFactory.java
+++ b/src/main/java/com/codeborne/selenide/webdriver/FirefoxDriverFactory.java
@@ -23,6 +23,7 @@ class FirefoxDriverFactory extends AbstractDriverFactory {
 
   @Override
   WebDriver create(final Proxy proxy) {
+    System.setProperty(FirefoxDriver.SystemProperty.BROWSER_LOGFILE, "/dev/null");
     return createFirefoxDriver(proxy);
   }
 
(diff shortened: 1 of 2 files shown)


src/main/java/com/codeborne/selenide/webdriver/FirefoxDriverFactory.java
```diff
@@ -23,6 +23,7 @@ boolean supports() {
 
   @Override
   WebDriver create(final Proxy proxy) {
+    System.setProperty(FirefoxDriver.SystemProperty.BROWSER_LOGFILE, "/dev/null");
```
BorisOsipov: System.getProperty("selenide.pollingInterval", "100"))

rosolko: We need to decide - do we really want to provide way to configure this property.
If yes - we have to introduce new configuration field.
If not - just hide this with default value.

Because if use suggested code - it will be undocumented configuration.

vinogradoff:
The number of configurations is growing. Let's consider splitting into - basic configs / adavnced / expert configs?

rosolko:
@vinogradoff Not sure about this. 
From my point of view it will be not intuitive for users.

asolntsev:
@rosolko @vinogradoff By the way, this PR fixes issue https://github.com/codeborne/selenide/issues/673

src/main/java/com/codeborne/selenide/webdriver/FirefoxDriverFactory.java
asolntsev: I vote for not adding new configuration field. Let's just use system property. Yes, it will be an "undocumented" feature, but almost nobody does really need to change it.

PR #735, opened 2018-05-10
Add '?' symbol to filename pattern to avoid url query string parameters.

### LINKED ISSUE #734: Incorrect filename of downloaded file
## Incorrect filename after downloading
Filename of the downloaded file is the same as href. So 
`href="myfileblabla.txt?sessioncookie=12345%2323`
will produce file with name "myfileblabla.txt?sessioncookie=12345%2323"

## Details
Need to extract filename from href until `?`

* **Selenide Version:** 4.11.4
* **Chrome\Firefox\IE Version:** 66
* **OS Version:** Windows

rosolko:
Fix #734 issue.

## Proposed changes

## Checklist

### DIFF
--- a/src/main/java/com/codeborne/selenide/impl/HttpHelper.java
+++ b/src/main/java/com/codeborne/selenide/impl/HttpHelper.java
@@ -11,7 +11,7 @@ import static org.apache.commons.lang3.StringUtils.defaultIfEmpty;
 
 public class HttpHelper {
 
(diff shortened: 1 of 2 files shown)


rosolko:
@asolntsev Not so easy, as long as this value going from http headers. 
So simple adding query string parameters to element `href`attribute ins't working.

BorisOsipov:
@rosolko may be you can extend these tests?

https://github.com/codeborne/selenide/blob/51aef5723f39c2adb1ed70c9a79591e328feb439/src/test/java/com/codeborne/selenide/impl/HttpHelperTest.java#L12

PR #736, opened 2018-05-10
Provide webdrivermanager on api level.

rosolko:
## Proposed changes
If you create custom driver provider and want to download custom binary use provided dependency.

## Checklist


rosolko:
@asolntsev Not from implementation level :)
Webdrivermanager for now marked in classpath as `provided` not `compile` and you can't just type `WebDriverManager.chromedriver().setup();` that's the point.

Issue #695, opened 2018-02-15
Remove Automatic Creation of FireFox Browser When No WebDriver Bound


cmkatarn:
## The problem
Could you please add a check to the property "selenide.reopenBrowserOnFail" in WebDriverThreadLocalContainer.getWebDriver()? Line 95 has a log message that "No webdriver is bound to the current thread", followed by line 96 which creates a new driver.

## Details
We happen to manage our own browser lifecycles (using WebDriverRunner.setWebDriver()) and run all of our tests on an internal hub/nodes, and would prefer no browser to be created if no webdriver is bound to the current thread. Currently, if we declare and initialize Selenide Page Objects outside of our test methods we are experiencing FireFox browsers being opened locally, even when running remotely. These browsers are never utilized as our tests use TestNG annotations (@BeforeClass) to set the WebDriver to a new instance of whatever type we've specified in our configuration files. 

With the existing code, there seems to be no way to prevent the creation of a new FireFox browser, even if our tests handle the creation internally.

## Code To Reproduce Issue
public class SomeClass {

//this will spawn a FF browser - it should not if reopenBrowserOnFail is 'false'
private static SomePageObject somePageObject = page(SomePageObject.class);

@Test
public void someTest() {
    //this will also spawn a FF browser - it should not if reopenBrowserOnFail is 'false'
    $("").click();
}

BorisOsipov:
Hi, @cmkatarn, thank you for feedback.

I don't undestand what we should do in your example

> @test
> public void someTest() {  
> //this will also spawn a FF browser - it should not if reopenBrowserOnFail is 'false' reopenbrowser  
> $("").click();  
> }  

What will happen when I call click() and there is no driver? throw exception? It is ridiculous..

asolntsev:
Can you explain what is expected behavior for you? What should Selenide do
if you try to create a page object, but haven't initialized browser yet?

cmkatarn:
[...]

PR #754, opened 2018-06-05
#695 Do not open a browser if `Configuration.reopenBrowserOnFail` is …

asolntsev:
…`false` and user has not set webdriver manually

## Proposed changes

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 # Changelog
 
 ## 4.12.2
-*
+* #695 Do not open a browser if `Configuration.reopenBrowserOnFail` is `false` and user has not set webdriver manually 
 
 ## 4.12.1 (released 02.06.2018)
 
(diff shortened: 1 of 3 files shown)


BorisOsipov:
@asolntsev Do you take into account such changes will break tests for users who accidentally still use `reopenBrowserOnFail=true`?

asolntsev:
@BorisOsipov Why? `reopenBrowserOnFail` is `true` by default. Nothing will change in this case.

BorisOsipov:
I've misprinted. If someone accidentally have been setted reopenBrowserOnFail=false.

asolntsev:
@BorisOsipov Yes, right, in this case the behaviour will change. The browser will not start. It's ok for me. The test will fail with a clear error message:

```
java.lang.IllegalStateException: No webdriver is bound to current thread: 1, and cannot create a new webdriver because Configuration.reopenBrowserOnFail=false
```

PR #755, opened 2018-06-05
Add ability to actualize element collection elements.

rosolko:
## Proposed changes
Add ability to actualize element collection elements.
Ignore in this case previously cached elements in collection.

## Checklist


asolntsev:
@rosolko I don't understand why it's needed. This is effectively the same as calling $$("...") once more.

rosolko:
@asolntsev I agree, but as discussed in slack some users want to work with single declared field, so this method allow to avoid cached elements in collection.

rosolko:
Close in favor of #762

PR #757, opened 2018-06-27
Migrate tests to Junit5 + AssertJ

rosolko:
## Proposed changes
Introduce `BrowserStrategyExtension`, `ScreenShooterExtension`, `SoftAssertsExtension`, `TextReportExtension` and test `MockWebDriverExtension` extensions as port of Junit4 rules.

Still not sure if we need to save old Junit4 test `SoftAssertJUnitTest` that ignored for a long time.
Also this test requred to save old `LegacyIntegrationTest` base test.
As solution I propose to delete this methods as test doesn't run.

**Note1:** Squash commits until merge.

**Note2:** Also I propose to move Junit4 rules and Junit5 extensions to modules. So no dependencies confuse facing.

**Update:** Ignored junit4 rule test replaced with documentation.

## Checklist


BorisOsipov:
why do you choice `AssertJ`? what was wrong with hamcrest?

vinogradoff:
Actually hamcrest was bundled with Junit4, that was the reason for me to
prefer it over AssertJ. Junit 5 bundles none of them, so now the choice is
tougher, because in any case you add extra dependency. I also like AssertJ
a bit more :)

Viele Grüße

Alexei Vinogradov

src/main/java/com/codeborne/selenide/junit5/BrowserStrategyExtension.java
```diff
@@ -0,0 +1,33 @@
+/*
+ * Copyright 2018 Aliaksandr Rasolka
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
```
BorisOsipov: Selenide is licensed by MIT. Why we have here Apache 2?

rosolko: Oops. Sorry about that. Copied from another project.

rosolko:
@BorisOsipov 
For first `hamcrest` for now development really slow.
For second `hamcrest` going from junit4 and in assert `expected` going before `actual` so it's really confusing.
Also `assertj` provide more fluent api and contains a lot of smart assertions for example - exceptions.
As it's dsl - IDE autocomplete matchers and possible asserts.
Less imports.

src/test/java/com/codeborne/selenide/UnitTest.java
```diff
@@ -0,0 +1,10 @@
+package com.codeborne.selenide;
+
+import org.assertj.core.api.WithAssertions;
+
+/**
+ * @author Aliaksandr Rasolka
+ * @since 4.12.2
+ */
+public class UnitTest implements WithAssertions {
```
vinogradoff: I don't understand what the metaphor should stand behind this class.
In any case, it should not appear in the main package of selenide, because as far as I know some of Selenide users - there will be many of them trying to use this class in every crazy and terrible way they can find :)

build.gradle
```diff
@@ -7,6 +7,7 @@ plugins {
   id 'checkstyle'
   id 'com.github.spotbugs' version '1.6.2'
   id 'org.sonarqube' version '2.6.2'
+  id 'com.github.ben-manes.versions' version '0.20.0'
```
asolntsev: What is this for?

src/main/java/com/codeborne/selenide/junit5/BrowserStrategyExtension.java
```diff
@@ -0,0 +1,17 @@
...
+import org.junit.jupiter.api.extension.AfterAllCallback;
+import org.junit.jupiter.api.extension.ExtensionContext;
+
+import static com.codeborne.selenide.WebDriverRunner.closeWebDriver;
+
+/**
+ * @author Aliaksandr Rasolka
+ * @since 4.12.2
+ */
+public class BrowserStrategyExtension implements AfterAllCallback {
```
asolntsev: 1. Add javadoc with usage sample. 
2. It's not clear from javadoc, is this "browser per class" or "browser per method"?

src/main/java/com/codeborne/selenide/junit5/ScreenShooterExtension.java
```diff
@@ -0,0 +1,83 @@
...
+ * @since 4.12.2
+ */
+public class ScreenShooterExtension implements BeforeAllCallback, AfterEachCallback, AfterAllCallback {
+  private static final Logger log = Logger.getLogger(ScreenShooterExtension.class.getName());
+
+  private boolean captureSuccessfulTests;
+
+  /**
+   * Initialize default screen shooter extension with disabled successful tests capture.
+   */
```
asolntsev: 1. remove useless javadoc
2. call `this(false)`

src/main/java/com/codeborne/selenide/junit5/ScreenShooterExtension.java
```diff
@@ -0,0 +1,83 @@
...
+import static com.codeborne.selenide.ex.ErrorMessages.screenshot;
+
+/**
+ * @author Aliaksandr Rasolka
+ * @since 4.12.2
+ */
+public class ScreenShooterExtension implements BeforeAllCallback, AfterEachCallback, AfterAllCallback {
+  private static final Logger log = Logger.getLogger(ScreenShooterExtension.class.getName());
+
+  private boolean captureSuccessfulTests;
```
asolntsev: make the field `final`

src/main/java/com/codeborne/selenide/junit5/SoftAssertsExtension.java
```diff
@@ -0,0 +1,51 @@
...
+    errorsCollector = new ErrorsCollector();
+  }
+
+  /**
+   * Return error collector object.
+   *
+   * @return ErrorsCollector instance
+   *
+   * @see ErrorsCollector
+   */
```
asolntsev: remove useless javadoc

src/main/java/com/codeborne/selenide/junit5/SoftAssertsExtension.java
```diff
@@ -0,0 +1,51 @@
...
+public class SoftAssertsExtension implements BeforeEachCallback, AfterEachCallback {
+  private final ErrorsCollector errorsCollector;
+
+  /**
+   * Create SoftAssertsExtension instance with empty ErrorsCollector object.
+   * <br/>
+   * Used during extension registration.
+   *
+   * @see ErrorsCollector
+   */
```
asolntsev: instead of useless javadoc, please add a code sample: how to use this class?

Issue #672, opened 2018-01-16
Lazy field initialisation via collections is broken


pavelpp:
## The problem
FIY, this  [PR](https://github.com/codeborne/selenide/issues/641) breaks lazy initialisation for fields in form `SelenideElement el = $$(css).get(n);`. Now it will actually try to initialise the field and try to look for the element whenever class default constructor is called.

## Tell us about your environment
* **Selenide Version:** 4.10
* **Chrome\Firefox\IE Version:**  N/A
* **Browser Driver Version:** N/A
* **Selenium Version:** N/A
* **OS Version:** N/A

BorisOsipov:
Looks like it was broken in https://github.com/codeborne/selenide/pull/653/

In this place
https://github.com/codeborne/selenide/blob/master/src/main/java/com/codeborne/selenide/ElementsCollection.java#L243-L245

BorisOsipov:
@CaBocuk can you have look?

CaBocuk:
Right - that's because of my fix. 
I'll try to investigate how to fix it without returning to re-evaluating of actual elements on every iteration.
Also it's a point for a unit test :)

vinogradoff:
Looks very critical to me, we should probably revert the pr, if not fixed
very soon.

Viele Grüße

Alexei Vinogradov

CaBocuk:
I think I have some hypotheses about a compromise for this case. Will check it tomorrow morning.

BorisOsipov:
I guess guy from this thread http://automated-testing.info/t/selenide-4-10-perestal-zapuskat-chrome-novye-izmeneniya-v-versii/18420/3 ran into the same issue.
It very similar to @pavelpp problem behaviour.

>we should probably revert the pr, if not fixed
very soon.
+1

asolntsev:
@BorisOsipov @CaBocuk @vinogradoff @pavelpp I have a fix. I am preparing a pull request.

asolntsev:
@BorisOsipov @CaBocuk @vinogradoff Please review the pull request: https://github.com/codeborne/selenide/pull/674

asolntsev:
@pavelpp I personally find that declaring page object field using `$$(css).get(n)` is quite strange. I don't like it. As a reader of this code, it's not obvious that "get" doesn't actually **get**. Lazy evaluation in Selenide was not intended for such things :)

pavelpp:
@asolntsev me neither, so I refactored this finding a better way. ;)

[...]

Issue #641, opened 2017-12-05
ElementsCollection performance using iterators


CaBocuk:
## The problem
Once I had to iterate through an Elements collection of 100+ elements I faced that it takes too much time (~10 seconds).

## Details
So I investigated the issue and found that when we do the iteration - Selenide uses **SelenideElementIterator** to iterate **BySelectorCollection**. And it calls **BySelectorCollection's** method `getActualElements()` on EVERY `hasNext()` and `next()` call!

So i tried an experiment: I've changed the code of BySelectorCollection::getActualElements to:
```
  public List<WebElement> getActualElements() {
    SearchContext searchContext = parent == null ? getWebDriver() : parent;
    List<WebElement> elements = WebElementSelector.instance.findElements(searchContext, selector);
    System.out.println("Running "+ ++counter + " iteration of getActualElements(). Size = " +elements.size());
    return elements;
  }
```
and tested it on:
```
public static void main(String...args){
      Configuration.browser = "chrome";
      open("https://yandex.by/video/");
      $$("span").forEach(item -> item.is(visible));
  }
```

As an output I got: `Running 365 iteration of getActualElements(). Size = 182`
So for iterating through collection of 182 elements it finds 365 * 182 = **66430 elements**

Holding experiments with elements amount I figured out the the amount found elements is **(2*N^2 + N)**, where N - amount of elements in collection.
So for iterating through collection of 500 elements Selenide needs to find 500,500 elements :)

Should it really behave like that?

## Tell us about your environment
* **Selenide Version: 4.8**
* **Chrome: 62.0.3202.94**
* **Browser Driver Version: automatically downloaded**

asolntsev:
@CaBocuk Thank you for reporting the issue. We definitely need to fix it.

CaBocuk:
Fixed it in #653 
If the changes are ok, we can close it I think 😃

Issue #277, opened 2016-01-30
Case to reproducing StaleElementReferenceException


admizh:
This test is demonstrate 100% reproduce throwing StaleElementReferenceException. How do we solve it?
Browser chrome: v 48.0.2564.97 m
Chrome driver 2.20

```java
@Test()
    public void refException() {
        open("http://ya.ru");
        List<SelenideElement> elements = new ArrayList<>();
        for (SelenideElement selenideElement : $$("#text")) {
            elements.add(selenideElement);
        }
        executeJavaScript("window.location.reload();");
        elements.get(0).sendKeys("Val");
    }
```

```exeption
java.lang.reflect.UndeclaredThrowableException
	at com.sun.proxy.$Proxy15.sendKeys(Unknown Source)
	at onboarding.demo.Demo4WorkWithCollection.refException(Demo4WorkWithCollection.java:55)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.testng.internal.MethodInvocationHelper.invokeMethod(MethodInvocationHelper.java:86)
	at org.testng.internal.Invoker.invokeMethod(Invoker.java:643)
	at org.testng.internal.Invoker.invokeTestMethod(Invoker.java:820)
	at org.testng.internal.Invoker.invokeTestMethods(Invoker.java:1128)
	at org.testng.internal.TestMethodWorker.invokeTestMethods(TestMethodWorker.java:129)
	at org.testng.internal.TestMethodWorker.run(TestMethodWorker.java:112)
	at org.testng.TestRunner.privateRun(TestRunner.java:782)
	at org.testng.TestRunner.run(TestRunner.java:632)
	at org.testng.SuiteRunner.runTest(SuiteRunner.java:366)
	at org.testng.SuiteRunner.runSequentially(SuiteRunner.java:361)
	at org.testng.SuiteRunner.privateRun(SuiteRunner.java:319)
	at org.testng.SuiteRunner.run(SuiteRunner.java:268)
	at org.testng.SuiteRunnerWorker.runSuite(SuiteRunnerWorker.java:52)
[...]

PR #762, opened 2018-07-04
#696 reload collection elements on every call

### LINKED ISSUE #696: Proxied elements in ElementsCollection are not re-searched
After discussion on Gitter "selenide-ru" channel, it was decided to create an issue.
## The problem
When the elements in `ElementsCollection collection = $$(locator);` are already been searched, when doing a search one more time after page content changed - the elements in the `collection` remain unchanged, seems like proxied elements are not being re-searched if they already been searched once
## Details
Example steps:
The actual test could be found below, but the generic steps are the following:
1) Open website (imdb in my case) on the page with ratings list
2) Find the ratings list with `ElementsCollection movies = $$(locator)`
3) Call any method on `movies` that invokes element search
4) On the UI - change sort type so that the elements in the collection are changed
5) Again, call any method on the `movies` collection, so that the elements search is invoked again

**Actual result:** list `movies` has old elements not updated accordingly to UI change
**Expected result:** the `movies` list elements are updated and contain new items from the UI

I could reproduce it on the imdb website, but I can't say this is due to their page design/implementation since the test below passes on older versions of Selenide - **tried 4.9 and 4.7** and only fails on **4.10.01**.
I assume this is somehow related to the bug [672](https://github.com/codeborne/selenide/issues/672) and a pull request [674](https://github.com/codeborne/selenide/pull/674) which was supposed to fix it.

## The environment:
* **Selenide Version:*4.10.01*
* **Chrome\Firefox\IE Version:*64.0.3282.119 (Official Build) (64-bit)*
* **Browser Driver Version:*2.33*
* **Selenium Version:*3.8.1*
* **OS Version:*Mac OS Sierra - 10.11.3*

## The test
```
@Test
public void testReproxy() {
    open("http://www.imdb.com/chart/moviemeter?ref_=nv_mv_mpm_8");

    ElementsCollection movieTitles = $$x(".//td[@class='titleColumn']/a");

    // call any method, that would invoke element search
    assertEquals("Black Panther", movieTitles.get(0).getText());

    // sort movies by Release date.
    $(By.name("sort")).selectOption("Release Date");

    // now the the first movie should be "The Irishman"
    assertEquals("The Irishman", movieTitles.get(0).getText());
    // but we fail here, because the first element in movieTitles list is still Black Panther
}
```

asolntsev:
partially revert changes done in #641, #672 and #277

## Proposed changes
This change fixes #696, but breaks the old issue #277

## Checklist

### DIFF
--- a/src/main/java/com/codeborne/selenide/ElementsCollection.java
+++ b/src/main/java/com/codeborne/selenide/ElementsCollection.java
@@ -11,6 +11,7 @@ import com.codeborne.selenide.impl.SelenideElementIterator;
 import com.codeborne.selenide.impl.SelenideElementListIterator;
 import com.codeborne.selenide.impl.TailOfCollection;
 import com.codeborne.selenide.impl.WebElementsCollection;
+import com.codeborne.selenide.impl.WebElementsCollectionWrapper;
 import com.codeborne.selenide.logevents.SelenideLog;
 import com.codeborne.selenide.logevents.SelenideLogger;
 import org.openqa.selenium.WebDriverException;
@@ -99,7 +100,7 @@ public class ElementsCollection extends AbstractList<SelenideElement> {
     final long startTime = System.currentTimeMillis();
     do {
       try {
-        actualElements = collection.getActualElements();
+        actualElements = collection.getElements();
         if (condition.apply(actualElements)) {
           return;
         }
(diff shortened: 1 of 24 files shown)


rosolko:
@asolntsev 
Please take a looks at this tests:
-  shouldNotReloadCollectionWhenIterating
- reloadsCollectionOnEveryCall
- reproduceStaleElementException_priorToSelenide33

PR #770, opened 2018-07-28
Implementing a way of retrieving the maxlength value

symonk:
## Proposed changes
Adding a simple means of retrieving the maxlength attribute value of a given webelement

## Checklist

### DIFF
--- a/src/main/java/com/codeborne/selenide/SelenideElement.java
+++ b/src/main/java/com/codeborne/selenide/SelenideElement.java
@@ -170,6 +170,13 @@ public interface SelenideElement extends WebElement, FindsByLinkText, FindsById,
   String name();
 
   /**
+   * Get the 'maxlength' attribute of the element
+   * @return attribute 'maxlength' value or null if attribute is missing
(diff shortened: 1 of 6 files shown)


BorisOsipov:
I am not sure that we should add it. 

1) There is common method `attr(String attributeName);` and it is enough from my point of view.
@symonk why you can't use $().attr("maxlength")?

2) Max length for inputs can be stored in any dom elements attribut e.g. for Angular it is `ngMaxlength`. So users will be confused when they call $().maxLength() and it won't work correctly.

rosolko:
@symonk Agree with @BorisOsipov about using named attribute value instead of separate method.
In my opinion it is functionality duplication.

symonk:
@BorisOsipov @rosolko after more thought, yes I agree with you both, especially when modern frameworks such as angular are using other attribute names, could cause end user confusion and unnecessary issue(s).  Thanks.

rosolko:
SO. Let's close out this PR for now.

PR #785, opened 2018-08-21
Fix #784 Enable BasicAuth through Selenide proxy server

### LINKED ISSUE #784: Enable BasicAuth through Selenide proxy server
## The problem
Currently it's only possible to bypass BasicAuth by adding username+password in URL. And this is deprecated now (at least in Chrome). We could add BasicAuth header using Selenide's built-in proxy server.

rosolko:
## Proposed changes
Fix #784 issue. Enable BasicAuth through Selenide proxy server

## Checklist

### DIFF
--- a/src/main/java/com/codeborne/selenide/AuthenticationType.java
+++ b/src/main/java/com/codeborne/selenide/AuthenticationType.java
@@ -1,0 +1,25 @@
+package com.codeborne.selenide;
+
+/**
+ * Authentication schemes.
+ *
+ * @see <a href="https://developer.mozilla.org/en-US/docs/Web/HTTP/Authentication#Authentication_schemes">Web HTTP reference</a>
+ */
+public enum AuthenticationType {
+  BASIC("Basic"),
+  BEARER("Bearer"),
+  DIGEST("Digest"),
+  HOBA("HOBA"),
+  MUTUAL("Mutual"),
+  AWS4_HMAC_SHA256("AWS4-HMAC-SHA256");
+
+  private String value;
+
+  AuthenticationType(final String value) {
+    this.value = value;
+  }
+
+  public String getValue() {
+    return value;
+  }
+}
--- a/src/main/java/com/codeborne/selenide/AuthenticationType.java
+++ b/src/main/java/com/codeborne/selenide/AuthenticationType.java
@@ -13,9 +13,9 @@ public enum AuthenticationType {
   MUTUAL("Mutual"),
   AWS4_HMAC_SHA256("AWS4-HMAC-SHA256");
 
-  private String value;
+  private final String value;
 
-  AuthenticationType(final String value) {
+  AuthenticationType(String value) {
     this.value = value;
   }
 
--- a/src/main/java/com/codeborne/selenide/Credentials.java
+++ b/src/main/java/com/codeborne/selenide/Credentials.java
@@ -4,11 +4,11 @@ import java.util.Base64;
 
 import static java.nio.charset.StandardCharsets.UTF_8;
 
-final class Credentials {
+public final class Credentials {
   private String login;
   private String password;
 
-  Credentials(final String login, final String password) {
+  public Credentials(final String login, final String password) {
     this.login = login;
     this.password = password;
   }
@@ -18,7 +18,7 @@ final class Credentials {
    *
    * @return encoded string
    */
-  String encode() {
+  public String encode() {
     final byte[] credentialsBytes = combine().getBytes(UTF_8);
     return Base64.getEncoder().encodeToString(credentialsBytes);
   }
--- a/src/main/java/com/codeborne/selenide/Credentials.java
+++ b/src/main/java/com/codeborne/selenide/Credentials.java
@@ -5,10 +5,10 @@ import java.util.Base64;
 import static java.nio.charset.StandardCharsets.UTF_8;
 
 public final class Credentials {
-  private String login;
-  private String password;
+  private final String login;
+  private final String password;
 
-  public Credentials(final String login, final String password) {
+  public Credentials(String login, String password) {
     this.login = login;
     this.password = password;
   }
@@ -19,15 +19,10 @@ public final class Credentials {
    * @return encoded string
    */
   public String encode() {
-    final byte[] credentialsBytes = combine().getBytes(UTF_8);
+    byte[] credentialsBytes = combine().getBytes(UTF_8);
     return Base64.getEncoder().encodeToString(credentialsBytes);
   }
 
-  /**
-   * Combine credentials with a colon (login:password).
-   *
-   * @return combined credentials
-   */
   private String combine() {
     return String.format("%s:%s", login, password);
   }
--- a/src/main/java/com/codeborne/selenide/Credentials.java
+++ b/src/main/java/com/codeborne/selenide/Credentials.java
@@ -1,0 +1,34 @@
+package com.codeborne.selenide;
+
+import java.util.Base64;
+
+import static java.nio.charset.StandardCharsets.UTF_8;
+
+final class Credentials {
+  private String login;
+  private String password;
+
+  Credentials(final String login, final String password) {
+    this.login = login;
+    this.password = password;
+  }
+
+  /**
+   * The resulting string is base64 encoded (YWxhZGRpbjpvcGVuc2VzYW1l).
+   *
+   * @return encoded string
+   */
+  String encode() {
+    final byte[] credentialsBytes = combine().getBytes(UTF_8);
+    return Base64.getEncoder().encodeToString(credentialsBytes);
+  }
+
+  /**
+   * Combine credentials with a colon (login:password).
+   *
+   * @return combined credentials
+   */
+  private String combine() {
+    return String.format("%s:%s", login, password);
+  }
+}
--- a/src/main/java/com/codeborne/selenide/Selenide.java
+++ b/src/main/java/com/codeborne/selenide/Selenide.java
@@ -9,7 +9,6 @@ import com.codeborne.selenide.impl.Navigator;
 import com.codeborne.selenide.impl.SelenideFieldDecorator;
 import com.codeborne.selenide.impl.WebElementsCollectionWrapper;
 
-import net.lightbody.bmp.BrowserMobProxy;
 import org.openqa.selenium.Alert;
 import org.openqa.selenium.By;
 import org.openqa.selenium.JavascriptExecutor;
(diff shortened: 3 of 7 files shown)


src/main/java/com/codeborne/selenide/Selenide.java
```diff
@@ -101,6 +104,35 @@ public static void open(String relativeOrAbsoluteUrl, String domain, String logi
...
+   * @param authenticationType
+   * @param login
+   * @param password
+   *
+   * @see <a href="https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Proxy-Authorization">Web HTTP reference</a>
+   * @see AuthenticationType
+   */
+  public static void open(final String relativeOrAbsoluteUrl, final AuthenticationType authenticationType,
+                          final String login, final String password) {
+    getAndCheckWebDriver();
```
asolntsev: Here is too much logic for `Selenide` class. All this stuff should be in `SelenideProxyServer` and/or `navigator.open`. In this case we would not need to increase `ClassFanOutComplexity` in checkstyle.xml

src/main/java/com/codeborne/selenide/Selenide.java
```diff
@@ -101,6 +104,35 @@ public static void open(String relativeOrAbsoluteUrl, String domain, String logi
...
+    final BrowserMobProxy proxy = getSelenideProxy().getProxy();
+    final Credentials credentials = new Credentials(login, password);
+    final String authorization = String.format("%s %s", authenticationType.getValue(), credentials.encode());
```
asolntsev: I personally do not like `final` near every local variable and method parameter. It's a noise.

src/main/java/com/codeborne/selenide/Selenide.java
```diff
@@ -101,6 +104,35 @@ public static void open(String relativeOrAbsoluteUrl, String domain, String logi
...
+   * @param login
+   * @param password
+   *
+   * @see <a href="https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Proxy-Authorization">Web HTTP reference</a>
+   * @see AuthenticationType
+   */
+  public static void open(final String relativeOrAbsoluteUrl, final AuthenticationType authenticationType,
+                          final String login, final String password) {
+    getAndCheckWebDriver();
+    final BrowserMobProxy proxy = getSelenideProxy().getProxy();
```
asolntsev: It's not a good idea to use BrowserMobProxy directly. Much better would be to add `RequestFilter` using `getSelenideProxy().addRequestFilter()`. This filter could add required headers.

rosolko:
@asolntsev Done.

src/main/java/com/codeborne/selenide/impl/Navigator.java
```diff
@@ -39,6 +43,19 @@ public void open(URL url, String domain, String login, String password) {
...
   }
 
+  public void open(String relativeOrAbsoluteUrl, AuthenticationType authenticationType, String login, String password) {
+    getAndCheckWebDriver();
+    Credentials credentials = new Credentials(login, password);
+    String authorization = String.format("%s %s", authenticationType.getValue(), credentials.encode());
+    getSelenideProxy().addRequestFilter("headers.request", (request, contents, messageInfo) -> {
+      final HttpHeaders headers = request.headers();
+      headers.add("Authorization", authorization);
+      headers.add("Proxy-Authorization", authorization);
```
asolntsev: @rosolko Isn't it a problem that we add this filter, but never remove?
It means that it will add `Authorization` header to ALL subsequent requests. Is it ok? Or we should remove this filter?

src/main/java/com/codeborne/selenide/Credentials.java
```diff
@@ -0,0 +1,34 @@
...
+  public String encode() {
+    final byte[] credentialsBytes = combine().getBytes(UTF_8);
+    return Base64.getEncoder().encodeToString(credentialsBytes);
+  }
+
+  /**
+   * Combine credentials with a colon (login:password).
+   *
+   * @return combined credentials
+   */
```
asolntsev: Do we really need javadoc for private methods?

src/main/java/com/codeborne/selenide/Credentials.java
```diff
@@ -0,0 +1,34 @@
+package com.codeborne.selenide;
+
+import java.util.Base64;
+
+import static java.nio.charset.StandardCharsets.UTF_8;
+
+public final class Credentials {
```
asolntsev: I would move this class to package `com.codeborne.selenide.impl` and make package-private. It's only used in `Navigator`.

src/main/java/com/codeborne/selenide/AuthenticationType.java
```diff
@@ -0,0 +1,25 @@
...
+ */
+public enum AuthenticationType {
+  BASIC("Basic"),
+  BEARER("Bearer"),
+  DIGEST("Digest"),
+  HOBA("HOBA"),
+  MUTUAL("Mutual"),
+  AWS4_HMAC_SHA256("AWS4-HMAC-SHA256");
+
+  private String value;
```
asolntsev: I would add `private final` here.

src/main/java/com/codeborne/selenide/AuthenticationType.java
```diff
@@ -0,0 +1,25 @@
...
+
+  AuthenticationType(final String value) {
```
asolntsev: I would remove `final` because it does't add any value.

src/main/java/com/codeborne/selenide/Selenide.java
```diff
@@ -101,6 +101,28 @@ public static void open(String relativeOrAbsoluteUrl, String domain, String logi
...
+   * Open browser and pass authentication using build-in proxy.
+   * <p>
+   * A common authenticationType is "Basic". See Web HTTP reference for other types.
+   * <p>
+   * {@code Configuration.fileDownload == Configuration.FileDownloadMode.PROXY;} - is required configuration.
+   *
+   * @param relativeOrAbsoluteUrl
+   * @param authenticationType
+   * @param login
+   * @param password
```
asolntsev: I would remove all useless lines from javadoc, like `@param login` without description.

src/main/java/com/codeborne/selenide/Selenide.java
```diff
@@ -101,6 +101,28 @@ public static void open(String relativeOrAbsoluteUrl, String domain, String logi
...
 
+  /**
+   *
+   * The main starting point in your tests.
+   * <p>
+   * Open browser and pass authentication using build-in proxy.
+   * <p>
+   * A common authenticationType is "Basic". See Web HTTP reference for other types.
+   * <p>
+   * {@code Configuration.fileDownload == Configuration.FileDownloadMode.PROXY;} - is required configuration.
```
asolntsev: I would rephrase it to something like "this method can only work if `{@code Configuration.fileDownload == Configuration.FileDownloadMode.PROXY;}`"

src/main/java/com/codeborne/selenide/impl/Navigator.java
rosolko: @asolntsev 
1. Previously you sad that we shouldn't call internal `getProxy()` and call only top level selenide proxy `getSelenideProxy()`
2. `autoAuthorization` method supports only `BASIC` auth type
3. If we remove filter isn't application logout user after another action in application?
4. If remove - how can we remove filters from selenide proxy? I see only add functionality.

src/main/java/com/codeborne/selenide/Credentials.java
rosolko: @asolntsev 
Yes we do.
The main point thats this object is public and method like this (I mean do some combining logic with special format) is necessary to de described in any way.

src/main/java/com/codeborne/selenide/Credentials.java
rosolko: @asolntsev 
I consider to move this class to top level because I think about extra `open` method with signatures containing `credentials` instead of `login` and `password` pair.
What do you thinks?

src/main/java/com/codeborne/selenide/AuthenticationType.java
rosolko: @asolntsev Good catch!

src/main/java/com/codeborne/selenide/Selenide.java
rosolko: @asolntsev 
We already had such empty javadocs lines for old `open(relativeOrAbsoluteUrl, domain, login, password)` method. 
I do it in one style during new implementation.
In this case we had to reformat all methods related to this point.

src/main/java/com/codeborne/selenide/Selenide.java
rosolko: @asolntsev Understood. Thanks!

src/main/java/com/codeborne/selenide/impl/Navigator.java
asolntsev: 1. Agree
2. Agree
3. I don't know exactly if logout happens. Probably. But anyway, we need to be able to somehow disable filter (when user calls `open` without basicauth) or reconfigure it (when user calls `open` with another username/password).
4. Right, it seems that we cannot remove filter from BMP. So, we can add filter only once, and enable/disable or configure existing filter.

src/main/java/com/codeborne/selenide/Credentials.java
asolntsev: Let's make it public when it's really needed. Currently it's not used, so there is no reason to make it public.

src/main/java/com/codeborne/selenide/Credentials.java
asolntsev: Yes, one parameter `Credentials` is better than two Strings "login"/"password".

src/main/java/com/codeborne/selenide/Selenide.java
asolntsev: I agree, we can cleanup javadoc later.

src/main/java/com/codeborne/selenide/Credentials.java
rosolko: Agree. Just remove javadoc since it's specially private method.

src/main/java/com/codeborne/selenide/Credentials.java
rosolko: Or we can just overload method and give ability to do open with `Credentials` and `login/password` pair?

PR #812, opened 2018-09-25
Configuration cleanup and refactoring.

### LINKED ISSUE #810: Selenide should NOT maximize browser by default
## The problem
Currently Selenide starts browser "maximized" (makes browser window as big as screen size). 
It has two problems:
1. It takes time
2. it can cause flaky tests because screen size can vary depending on environment, XVFB settings, headless mode etc. 

## Solution
I suggest to turn `Configuration.startMaximized` setting to `false` by default. 
Instead, I recommend to set browser window size explicitly by using `Configuration.browserSize=1024x768` (a minimal browser size that your application should support).

### LINKED ISSUE #811: Make Chrome the default browser in Selenide 5.0.0
Since a major 5.0.0 release is coming, I think it's a good opportunity to finally make Chrome the default browser instead of Firefox. IMHO that's what most people are using nowadays as a default browser anyway.

rosolko:
Set chrome as default browser and close #811.
Do not maximize browser by default and close #810.
Set default browser size to 1024x768.
Set default browser position to 30x30.
Style all configuration with camelCase format.
Remove deprecate from internet explorer IE alias.

## Proposed changes

## Checklist

### DIFF
--- a/src/main/java/com/codeborne/selenide/SelenideConfig.java
+++ b/src/main/java/com/codeborne/selenide/SelenideConfig.java
@@ -4,25 +4,22 @@ import com.codeborne.selenide.impl.JenkinsReportUrl;
 import org.openqa.selenium.remote.DesiredCapabilities;
 
 import static com.codeborne.selenide.AssertionMode.STRICT;
-import static com.codeborne.selenide.Browsers.FIREFOX;
+import static com.codeborne.selenide.Browsers.CHROME;
 import static com.codeborne.selenide.FileDownloadMode.HTTPGET;
 import static com.codeborne.selenide.SelectorMode.CSS;
 
 public class SelenideConfig implements Config {
-  private String browser = System.getProperty("selenide.browser", System.getProperty("browser", FIREFOX));
+  private String browser = System.getProperty("selenide.browser", System.getProperty("browser", CHROME));
   private boolean headless = Boolean.parseBoolean(System.getProperty("selenide.headless", "false"));
   private String remote = System.getProperty("selenide.remote", System.getProperty("remote"));
-  private String browserSize = System.getProperty("selenide.browserSize", System.getProperty("selenide.browser-size"));
-  private String browserVersion = System.getProperty("selenide.browserVersion",
-    System.getProperty("selenide.browser.version", System.getProperty("browser.version")));
-  private String browserPosition = System.getProperty("selenide.browserPosition");
-  private boolean startMaximized = Boolean.parseBoolean(System.getProperty("selenide.startMaximized",
-    System.getProperty("selenide.start-maximized", "true")));
+  private String browserSize = System.getProperty("selenide.browserSize", "1024x768");
+  private String browserVersion = System.getProperty("selenide.browserVersion");
+  private String browserPosition = System.getProperty("selenide.browserPosition", "30x30");
+  private boolean startMaximized = Boolean.parseBoolean(System.getProperty("selenide.startMaximized", "false"));
   private boolean driverManagerEnabled = Boolean.parseBoolean(System.getProperty("selenide.driverManagerEnabled", "true"));
   private String browserBinary = System.getProperty("selenide.browserBinary", "");
-  private String chromeSwitches = System.getProperty("selenide.chrome.switches", System.getProperty("chrome.switches"));
-  private String pageLoadStrategy = System.getProperty("selenide.pageLoadStrategy",
-    System.getProperty("selenide.page-load-strategy", "normal"));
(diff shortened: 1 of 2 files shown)


src/main/java/com/codeborne/selenide/SelenideConfig.java
```diff
@@ -4,25 +4,22 @@
...
+  private String browser = System.getProperty("selenide.browser", System.getProperty("browser", CHROME));
   private boolean headless = Boolean.parseBoolean(System.getProperty("selenide.headless", "false"));
   private String remote = System.getProperty("selenide.remote", System.getProperty("remote"));
-  private String browserSize = System.getProperty("selenide.browserSize", System.getProperty("selenide.browser-size"));
-  private String browserVersion = System.getProperty("selenide.browserVersion",
-    System.getProperty("selenide.browser.version", System.getProperty("browser.version")));
-  private String browserPosition = System.getProperty("selenide.browserPosition");
-  private boolean startMaximized = Boolean.parseBoolean(System.getProperty("selenide.startMaximized",
-    System.getProperty("selenide.start-maximized", "true")));
+  private String browserSize = System.getProperty("selenide.browserSize", "1024x768");
```
BorisOsipov: maybe 1366x768 default will be better? 
See statistics https://www.w3schools.com/browsers/browsers_display.asp

rosolko: This value based on note from #810:
> I recommend to set browser window size explicitly by using Configuration.browserSize=1024x768 (a minimal browser size that your application should support).

BorisOsipov: yep I got it. But I think `1024x768` is legacy screen resolution

asolntsev: Yes, I put `1024x768` without any reason. Just some memories from my childhood. :)

src/main/java/com/codeborne/selenide/SelenideConfig.java
```diff
@@ -9,9 +9,9 @@
...
 
 public class SelenideConfig implements Config {
-  private String browser = System.getProperty("selenide.browser", System.getProperty("browser", CHROME));
+  private String browser = System.getProperty("selenide.browser", CHROME);
   private boolean headless = Boolean.parseBoolean(System.getProperty("selenide.headless", "false"));
-  private String remote = System.getProperty("selenide.remote", System.getProperty("remote"));
+  private String remote = System.getProperty("selenide.remote");
   private String browserSize = System.getProperty("selenide.browserSize", "1024x768");
   private String browserVersion = System.getProperty("selenide.browserVersion");
   private String browserPosition = System.getProperty("selenide.browserPosition", "30x30");
```
asolntsev: I am not sure setting `30x30` is a good idea. Because moving the browser takes time. It will make many tests slower. 
And it was never needed - at least in my projects.

src/main/java/com/codeborne/selenide/SelenideConfig.java
```diff
@@ -33,14 +30,12 @@
...
   private long closeBrowserTimeoutMs = Long.parseLong(System.getProperty("selenide.closeBrowserTimeout", "5000"));
-  private boolean clickViaJs = Boolean.parseBoolean(System.getProperty("selenide.clickViaJs",
-    System.getProperty("selenide.click-via-js", "false")));
+  private boolean clickViaJs = Boolean.parseBoolean(System.getProperty("selenide.clickViaJs", "false"));
   private boolean captureJavascriptErrors = Boolean.parseBoolean(System.getProperty("selenide.captureJavascriptErrors", "true"));
   private boolean screenshots = Boolean.parseBoolean(System.getProperty("selenide.screenshots", "true"));
 
   private boolean savePageSource = Boolean.parseBoolean(System.getProperty("selenide.savePageSource", "true"));
-  private String reportsFolder = System.getProperty("selenide.reportsFolder",
-    System.getProperty("selenide.reports", "build/reports/tests"));
```
darklynx: Since you have dropped the support for `selenide.reports` system variable in favor of `selenide.reportsFolder`, you should probably update the official documentation: https://selenide.org/documentation/screenshots.html

asolntsev: It seems that it's already fixed? I see the following text now:

> Yes. You can use property `-Dselenide.reportsFolder=test-result/reports` to set any directory to store screenshots to. For v4.x use `-Dselenide.reports=test-result/reports`

darklynx: yes, it was fixed in scope of issue #894 by @BorisOsipov

PR #816, opened 2018-09-27
Remove deprecates. Close #806 issue. 

### LINKED ISSUE #806: Remove deprecated APIs
In upcoming major release Selenide 5.0.0, we want to delete deprecated and unused functionality from Selenide. We need to know your opinion. Please say in comments if you use some of settings/functions below. 

I suggest to delete the following settings:
- [x] `Configuraton.collectionsTimeout`  (let it be same as `Configuraton.timeout`).
- [x] `Configuration.collectionsPollingInterval` (let it be same as `Configuration.pollingInterval`).
- [x] `Configiration.closeBrowserTimeoutMs` (browser is closed anyway in a separate daemon thread).
- [x] `Configuration.chromeSwitches` (it has been deprecated for a while)
- [x] `Configuration.captureJavascriptErrors`  (it doesn't work well anyway) - remove `JavascriptErrorsCollector` as well
- [x] `Configuration.setValueChangeEvent` (we introduced this setting just in case to avoid breaking changes. Now it's clear for me that it's not needed).

I also suggest to delete the following deprecated `Condition`s:
- [x] `present`
- [x] `hasAttribute`
- [x] `hasValue`
- [x] `hasText`
- [x] `hasClass`

**UPD**.  @eichisanden @Vyachelav @SergeyPirogov 
The community voted that these two features should **not be removed** because people do actually use them:
1. `Configuration.assertionMode` (it seems to me that soft assertion is not really useful in Selenide, and nobody actually uses it).
2. `Configuration.holdBrowserOpen` (it's only needed during local debug - but you can debug without a setting anyway)

rosolko:
## Proposed changes

Close #806 issue.

Note:
1. Remove also `JavascriptErrorsCollector` class with `Configuration.captureJavascriptErrors`
2. Move value directly to `CloseDriverCommand` ctor from `Configuration.closeBrowserTimeoutMsconfiguration`
3. Change default `Configuration.pollingInterval` from `100` to `200` related to removed `Configuration.collectionPollingInterval`

## Checklist


src/main/java/com/codeborne/selenide/SelenideElement.java
```diff
@@ -6,7 +6,7 @@
 import org.openqa.selenium.WebElement;
 import org.openqa.selenium.WrapsDriver;
 import org.openqa.selenium.internal.WrapsElement;
-import org.openqa.selenium.interactions.Locatable;
+import org.openqa.selenium.interactions.internal.Locatable;
```
asolntsev: why? it's deprecated.

src/main/java/com/codeborne/selenide/commands/SetValue.java
```diff
@@ -50,19 +50,13 @@ private void setValueForTextInput(Driver driver, WebElement element, String text
     else if (driver.config().fastSetValue()) {
       String error = setValueByJs(driver, element, text);
       if (error != null) throw new InvalidStateException(driver, error);
```
asolntsev: I guess we cannot delete this part. When we set value by JavaScript, we still need to trigger `change` and other events. It can only be skipped when we set value by `sendKeys()` (because browser does it automatically anyway).

src/main/java/com/codeborne/selenide/drivercommands/CloseDriverCommand.java
```diff
@@ -14,10 +14,10 @@
   private final SelenideProxyServer selenideProxyServer;
   private final long closeBrowserTimeoutMs;
 
-  CloseDriverCommand(WebDriver webDriver, SelenideProxyServer selenideProxyServer, long closeBrowserTimeoutMs) {
+  CloseDriverCommand(WebDriver webDriver, SelenideProxyServer selenideProxyServer) {
     this.webDriver = webDriver;
     this.selenideProxyServer = selenideProxyServer;
-    this.closeBrowserTimeoutMs = closeBrowserTimeoutMs;
+    this.closeBrowserTimeoutMs = 5000;
```
asolntsev: I suggest to remove this field at all.
There is no need to limit lifetime of `CloseDriverCommand`. Let it work as long as it takes.

src/main/java/com/codeborne/selenide/SelenideConfig.java
```diff
@@ -23,22 +23,17 @@
 
   private String baseUrl = System.getProperty("selenide.baseUrl", "http://localhost:8080");
   private long timeout = Long.parseLong(System.getProperty("selenide.timeout", "4000"));
-  private long collectionsTimeout = Long.parseLong(System.getProperty("selenide.collectionsTimeout", "6000"));
-  private long pollingInterval = Long.parseLong(System.getProperty("selenide.pollingInterval", "100"));
-  private long collectionsPollingInterval = Long.parseLong(System.getProperty("selenide.collectionsPollingInterval", "200"));
+  private long pollingInterval = Long.parseLong(System.getProperty("selenide.pollingInterval", "200"));
```
asolntsev: Are you really sure we want to increase default `pollingInterval` from 100ms to 200ms? Why?
As to me, 100 was a reasonable default. Probably we could make it even fewer.

rosolko: This value going from collections polling interval.
With `100` value related test going to be unstable.
There are two options:
1. Set to `100` and revise tests
2. Leave `200`

src/main/java/com/codeborne/selenide/drivercommands/CloseDriverCommand.java
rosolko: Ok

src/main/java/com/codeborne/selenide/commands/SetValue.java
rosolko: But in this case based on setValueChangeEvent config we just add an extra "change" event.

In other case we trigger all other "keydown", "keypress", "input", "keyup", and now we also trigger them.

src/main/java/com/codeborne/selenide/SelenideConfig.java
asolntsev: @rosolko ok, we can leave 200, and later investigate why some tests fail with value 100.

PR #818, opened 2018-09-29
Fix FirefoxDriverFactory overwrites Firefox profile provided by Configuration issue

### LINKED ISSUE #817: FirefoxDriverFactory overwrites Firefox profile provided by Configuration
## The problem
FirefoxDriverFactory overwrites Firefox profile provided by Configuration.browserCapabilities 

## Details
I was trying to start Firefox with a custom profile with some certificates so the easiest approach was to set         `Configuration.browserCapabilities = capabilitiesWithFirefoxProfile;` . However, when using default Firefox driver provider, profile capability is overwritten in `com.codeborne.selenide.webdriver.FirefoxDriverFactory#transferFirefoxProfileFromSystemProperties`.
It can be worked around by implementing provider on your own but it's something I'd rather avoid if that is the only reason to implement it :)

* **Selenide Version: 4.14.2**
* **Firefox version: 62.0**

## Code To Reproduce Issue
```
FirefoxProfile profile = new FirefoxProfile();
profile.setPreference("security.default_personal_cert", "Select Automatically");
FirefoxOptions firefoxOptions = new FirefoxOptions().setProfile(profile);

Configuration.browser = "firefox";
Configuration.browserCapabilities = new DesiredCapabilities(firefoxOptions);

Selenide.open("https://github.com/codeborne/selenide");
```
Then navigate to `about:config` and check `security.default_personal_cert` property, it should be set to "Select Automatically"

## Possible fix
My suggestion would be to load FirefoxProfile from passed `FirefoxOptions` in `com.codeborne.selenide.webdriver.FirefoxDriverFactory#transferFirefoxProfileFromSystemProperties`, this would allow to set the profile in code and add another profile preferences from system properties.
The method would look like this:
```
private FirefoxOptions transferFirefoxProfileFromSystemProperties(FirefoxOptions currentFirefoxOptions) {
        String prefix = "firefoxprofile.";
        FirefoxProfile profile = Optional.ofNullable(currentFirefoxOptions.getProfile())
                                         .orElseGet(FirefoxProfile::new);
        for (String key : System.getProperties().stringPropertyNames()) {
            if (key.startsWith(prefix)) {
                String capability = key.substring(prefix.length());
                String value = System.getProperties().getProperty(key);
                log.config("Use " + key + "=" + value);
                if (value.equals("true") || value.equals("false")) {
                    profile.setPreference(capability, Boolean.valueOf(value));
                }
                else if (value.matches("^-?\\d+$")) { //if integer
                    profile.setPreference(capability, Integer.parseInt(value));
                }
                else {
                    profile.setPreference(capability, value);
                }
            }
        }
        return currentFirefoxOptions.setProfile(profile);
    }
```

BorisOsipov:
Change default value for Configuration.browserCapabilities to not null
Fix FirefoxDriverFactory overwrites Firefox profile provided by Configuration issue

## Checklist

### DIFF
--- a/src/main/java/com/codeborne/selenide/Configuration.java
+++ b/src/main/java/com/codeborne/selenide/Configuration.java
@@ -149,9 +149,9 @@ public class Configuration {
    * Browser capabilities.
    * Warning: this capabilities will override capabilities were set by system properties.
    * <br>
-   * Default value: null
+   * Default value: DesiredCapabilities::new
    */
-  public static DesiredCapabilities browserCapabilities;
+  public static DesiredCapabilities browserCapabilities = new DesiredCapabilities();
   /**
    * Should webdriver wait until page is completely loaded.
    * Possible values: "none", "normal" and "eager".
--- a/src/main/java/com/codeborne/selenide/webdriver/FirefoxDriverFactory.java
+++ b/src/main/java/com/codeborne/selenide/webdriver/FirefoxDriverFactory.java
@@ -1,6 +1,7 @@
 package com.codeborne.selenide.webdriver;
 
 import com.codeborne.selenide.WebDriverRunner;
+import java.util.Optional;
 import org.openqa.selenium.Proxy;
 import org.openqa.selenium.WebDriver;
 import org.openqa.selenium.firefox.FirefoxDriver;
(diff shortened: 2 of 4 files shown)


BorisOsipov:
Fixed https://github.com/codeborne/selenide/issues/819 too

BorisOsipov:
I make PR to v4 branch. It was made from latest v4 version tag. I will port it to master after this PR will be approved.

asolntsev:
@BorisOsipov Approved!

PR #820, opened 2018-09-30
Port https://github.com/codeborne/selenide/pull/818 to master

BorisOsipov:
## Proposed changes
Port https://github.com/codeborne/selenide/pull/818 to master


BorisOsipov:
Mistake.

PR #825, opened 2018-10-01
Workaround for webdrivermanager

asolntsev:
## Proposed changes
Let's store in a file `~/.m2/repository/webdriver/webdriver.gecko.driver.mac64.timestamp` the timestamp when WDM checked for updates last time.

It will allows us to avoid WDM calling github too often thus getting "403 error" too often.

## Checklist

### DIFF
--- a/src/main/java/com/codeborne/selenide/webdriver/WebDriverBinaryManager.java
+++ b/src/main/java/com/codeborne/selenide/webdriver/WebDriverBinaryManager.java
@@ -5,12 +5,16 @@ import io.github.bonigarcia.wdm.WebDriverManager;
 
 import java.io.File;
 import java.io.IOException;
+import java.util.Date;
 import java.util.concurrent.TimeUnit;
+import java.util.logging.Logger;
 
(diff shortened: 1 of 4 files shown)


src/main/java/com/codeborne/selenide/webdriver/WebDriverBinaryManager.java
```diff
@@ -58,20 +62,35 @@ private void setupFirefox() {
 
   private void cacheMeIfYouCan(String systemPropertyName, Runnable webdriverSetup) {
     if (webdriverIsAlreadyInitialized(systemPropertyName)) {
+      log.info("Skip: webdriver is already initialized: " + System.getProperty(systemPropertyName));
```
rosolko: Do we really need it on `info` level?
I think debug will be enough

asolntsev: I planned to delete this commit in future. It's temporary needed to debug the problem.

PR #829, opened 2018-10-03
Run task in parallel using build stages

rosolko:
## Proposed changes
See travis-ci build for details.
Link: https://travis-ci.org/codeborne/selenide/builds/436621455

Probably we need to test build without storing cache in future.
Because it cache each build stage. 5 vs 1 cache operation.

As result we have:
```
Ran for 9 min 29 sec
Total time 31 min 3 sec 
```
Versus on master:
```
Ran for 23 min 49 sec 
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,5 @@
+conditions: v1
+
 git:
   depth: 99999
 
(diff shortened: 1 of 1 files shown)


rosolko:
```

PR #830, opened 2018-10-03
Fix minor code smell issues reported by Sonar

plagov:
* remove catch for `NoSuchFrameException` because it is a subclass of
`WebDriverException`
* make `isFirefox62Bug()` method static since it's a private method


plagov:
Sonar also reports three code smells with major status to replace with an if/else structure. So was it done on purpose and these warnings from Sonar can be ignored? Or it would be better to replace these ternary operators with if/else?

BorisOsipov:
>Probably we should reconfigure Sonar to skip some checks that don't seem relevant?

Sure we can.

rosolko:
Close for now. First we need to describe and promote final code style.

PR #832, opened 2018-10-05
Screenshots outside of "user.dir" in CI server

admizh:
## Proposed changes
1. When you use reportsFolder a located outside of  "user.dir" in CI server, selenide generate wrong path in final error message. 
My changes provide additional ability to catch this, now: 
```
config.reportsFolder("C://artifacts-storage/"); //directory, that not in 'user.dir'
config.reportsUrl=http://ci.ru/
```
then selenide will do simple concatenation [config.reportsUrl + screanshot/pagesource file name]

2. Added Teamcity detection

## Checklist
- [X] Checkstyle and unit tests pass locally with my changes by running `gradle check chrome htmlunit` command


rosolko:
Please, run all verifications and test locally first.
`./gradlew check htmlunit chrome_headless firefox_headless`

Test still failing:
```
com.codeborne.selenide.ex.ErrorMessagesTest > convertsReportUrlForOutsideSavedScreenshot() FAILED
    java.lang.AssertionError at ErrorMessagesTest.java:76
```

admizh:
> Please, run all verification and tests locally first.
> `./gradlew check htmlunit chrome_headless firefox_headless`

yes, it executed. But some issues here (in Windows). I see checkstyle failures for fresh fetched repository, without any changes:
```
[ant:checkstyle] [ERROR] C:\Develop\projects\selenide\selenide-fork\src\main\java\com\codeborne\selenide\webdriver\SafariDriverFactory.java:0: File does not end with a newline. [NewlineAtEndOfFile]
...
...
[ant:checkstyle] [ERROR] C:\Develop\projects\selenide\selenide-fork\src\main\java\com\codeborne\selenide\webdriver\WebDriverBinaryManager.java:0: File does not end with a newline. [NewlineAtEndOfFile]
[ant:checkstyle] [ERROR] C:\PDevelop\projects\selenide\selenide-fork\src\main\java\com\codeborne\selenide\webdriver\WebDriverFactory.java:0: File does not end with a newline. [NewlineAtEndOfFile]

> Task :checkstyleMain FAILED
:checkstyleMain (Thread[Task worker for ':',5,main]) completed. Took 3.648 secs.

```
@rosolko i think check-style tests depends of operation system, and some other jvm settings. 
 


> Test still failing:
> 
> ```
> com.codeborne.selenide.ex.ErrorMessagesTest > convertsReportUrlForOutsideSavedScreenshot() FAILED
>     java.lang.AssertionError at ErrorMessagesTest.java:76
> ```
fixed in:  https://github.com/codeborne/selenide/pull/832/commits/6315e783e4653957d79b7bf69e0198bbdf1d2206

admizh:
@rosolko please describe your points ?

rosolko:
@admizh CI type identifier not generic enough. It's ok for now, but we need to refactor and improve this method in future.

src/test/java/com/codeborne/selenide/SelenideConfigTest.java
```diff
@@ -18,6 +18,14 @@ void canConstructReportsUrlFromJenkinsProperty() {
     assertThat(new SelenideConfig().reportsUrl()).isEqualTo("http://ci.org/job/123/artifact/");
   }
 
+  @Test
+  void canConstructReportsUrlFromTeamcityProperty() {
+    System.setProperty("teamcity.serverUrl", "http://ci.org/");
+    System.setProperty("teamcity.buildType.id", "my-build");
+    System.setProperty("build.number", "1");
+    assertThat(new SelenideConfig().reportsUrl()).isEqualTo("http://ci.org//repository/download/my-build/1:id/");
```
rosolko: Double slash after `.org` part, is it correct?

src/main/java/com/codeborne/selenide/impl/CiReportUrl.java
```diff
@@ -0,0 +1,54 @@
...
+    }
+    LOG.config("Variable selenide.reportsUrl not found");
+    return reportsUrl;
+  }
+
+  private String getTeamCityUrl() {
+    String url = System.getProperty("teamcity.serverUrl");
+    String build_type = System.getProperty("teamcity.buildType.id");
+    String build_number = System.getProperty("build.number");
+    String result = url + "/repository/download/" + build_type + "/" + build_number + ":id/";
```
rosolko: Let's use `Paths.get(url, "repository", "download", build_type, ...).toString()` instead of summ operation.
In this case we don't case about platform's folders/url dividers.
And avoid `/` symbol duplication.

src/main/java/com/codeborne/selenide/impl/CiReportUrl.java
```diff
@@ -0,0 +1,54 @@
...
+    } else {
+      LOG.config("Using teamcity artifacts: " + result);
+      return result;
+    }
+  }
+
+  private String getJenkinsReportsUrl() {
+    String build_url = System.getProperty("BUILD_URL");
+    if (!isEmpty(build_url)) {
+      return build_url + "artifact/";
```
rosolko: Here the same.

src/main/java/com/codeborne/selenide/impl/CiReportUrl.java
admizh: We can get illegal characters from the reportUrl like ':id', Paths solution not fit for here. We should use url builders and etc. but it is overhead, hence added url format checking and normalization (for remove double slashes)

asolntsev:
@admizh @rosolko Thank you! Squashed and merged. Releasing Selenide 5.0.0-rc.3

PR #836, opened 2018-10-11
Ignore escaping commas

FrameBassman:
## Proposed changes
Fix for https://github.com/codeborne/selenide/issues/697 

## Checklist


FrameBassman:
I think that current suggested solution in issue is not effective due to complex string processing. 
I suggest to escape all important commas in arguments via backslash () character.

BorisOsipov:
@vinogradoff can you look into this and give our feedback?

vinogradoff:
Hm, I think \\, is not intuitive, and only can be used if documented well.

Is Quotes processing really that complex?

vinogradoff:
Propose to close - solved in #883

asolntsev:
@vinogradoff Can you resolve conflicts?

vinogradoff:
Solved with #883

PR #841, opened 2018-10-20
deprecated exist and replace it by existInDOM

vinogradoff:
Exist is ambiguous and is often confused with Visible, causing flaky tests by design.

Typical example:

 ```
 alertBox.should(exist); // instead of alertBox.shouldBe(visible);
  // some actions which are okay, if alertBox is visible
```

Now, if alertBox is just hidden, but exist in DOM (very often the case), it will make test flaky.

The goal of this PR is to make very clear, that this condition is not the same as visible, and thus - reduce the use of it to only cases where it is really necessary (like checking some hidden fields to be present).

alertBox.should(exist) is an example of wrong usage. 
Typical usage for exist is for example

`  $("input[type=hidden][name=special_discount_field]").should(exist);
`

### DIFF
--- a/src/main/java/com/codeborne/selenide/Condition.java
+++ b/src/main/java/com/codeborne/selenide/Condition.java
@@ -49,7 +49,7 @@ public abstract class Condition {
    *
    * <p>Sample: {@code $("input[type=hidden]").should(existInDOM);}</p>
    */
-  public static final Condition existInDOM = new Condition("existInDOM") {
+  public static final Condition existInDOM = new Condition("exist in DOM") {
     @Override
     public boolean apply(Driver driver, WebElement element) {
       try {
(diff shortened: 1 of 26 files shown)


vinogradoff:
> I'm also down-vote this changes. `exist` is quiet obvious.

Sorry, but exist is anything but obvious, I myself got trapped when I was using Selenide the first months. 

For now the tests are failing which is bad, I will of course look to fix it.

rosolko:
Still not convincing enough. I'm not sure about this changes.

Denysss:
> Exist is ambiguous and is often confused with Visible ...
Maybe, there is a need to provide more descriptive information for the Selenide Conditions (exist, visible etc.) for avoiding it. For instance, create a comparison table of the conditions.

asolntsev:
I am not sure that using "DOM" term is a good idea. "DOM" is only applicable for web applications, but Selenide can also test mobile and desktop applications. In Selenide, `$(locator).should(exist)` means that elements should be findable by given locator. I cannot suggest better name for it than "exists".

rosolko:
@asolntsev @BorisOsipov propose to close this PR, give preference to `exist` condition and leave it without changes

Issue #201, opened 2015-07-26
displayed:false on visible elements


zolrath:
Hi! While testing the Selenide API I found a case using the default Firefox Driver (using Firefox 39.0 on OS X Yosemite) where a plainly visible element on the page fails to be interacted with as Selenide believes it isn't displayed.

```
! com.codeborne.selenide.ex.ElementShould: Element should be visible {By.name: session[username_or_email]}
! Element: '<input autocomplete="on" class="js-username-field email-input js-initial-focus" name="session[username_or_email]" type="text" displayed:false></input>'
```

gets thrown when the following code is run:

```
open("https://twitter.com/login/");
$(By.name("session[username_or_email]")).val("name");
$(By.name("session[password]")).setValue("password");
$(".submit").click();
```

asolntsev:
@zolrath The problem is that there is 3 such fields on the page.

Try to add the following line to your test:
```java
$$(By.name("session[username_or_email]")).shouldHave(size(1));
```

Then you will get the folloing error:

```java
ListSizeMismatch : expected: 1, actual: 3, collection: By.name: session[username_or_email]
Elements: [
	<input autocomplete="on" class="js-username-field email-input js-initial-focus" name="session[username_or_email]" type="text" displayed:false></input>,
	<input autocomplete="on" class="js-username-field email-input js-initial-focus" name="session[username_or_email]" placeholder="Phone, email or username" type="text"></input>,
	<input autocomplete="on" class="text-input email-input" id="signin-email" name="session[username_or_email]" placeholder="Phone, email or username" type="text" displayed:false></input>
]
```

There is another feature request that addresses the same problem: https://github.com/codeborne/selenide/issues/129

Thank you for reporting a very easy example (Twitter). I will use it for implementing issue https://github.com/codeborne/selenide/issues/129.

zolrath:
Thanks for the quick response! 
`$$(By.name("session[username_or_email]")).filter(visible).get(0).setValue("name");` was my workaround in this case.

cunhanohaole:
I am facing the same error and my page definetly does not have more than one field with the same id.

Any idea what else could be causing this?

I am trying to select a check box by id and it says it is not visible...

asolntsev:
Hi!
Can you code sample and html?

stupchiy:
@asolntsev I have the same issue currently and it is also, the same as from last commented, not connected with more than one element with the same id.

Initially I had PageObject with defined element using @FindBy:
```
[...]

PR #851, opened 2018-11-02
#201 can click elements with zero opacity

asolntsev:
This is a possible solution for https://github.com/selenide/selenide/issues/201

## Proposed changes
Change method `$.click()` so that it doesn't require element to be visible.

## Checklist

### DIFF
--- a/src/main/java/com/codeborne/selenide/SelenideElement.java
+++ b/src/main/java/com/codeborne/selenide/SelenideElement.java
@@ -739,6 +739,17 @@ public interface SelenideElement extends WebElement, WrapsDriver, WrapsElement, 
   @Override void click();
 
   /**
+   * Click the element ignore it's visibility. Check only element existence
+   * <p>
+   * You can use it for element that have opacity equals to zero.
+   * <p>
+   * Example: {@code <button style="opacity: 0;">Zero Opacity</button>}
+   *
+   * @see com.codeborne.selenide.commands.ClickIgnoreVisibility
+   */
+  void clickIgnoreVisibility();
+
+  /**
    * Click the element with a relative offset from the upper left corner of the element
    */
   void click(int offsetX, int offsetY);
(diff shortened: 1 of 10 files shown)


asolntsev:
@rosolko yes, that's why I am in doubt. Test definitely should not click hidden elements. But in this case the element is visible (even with `opacity: 0`). This is some css magic.

BorisOsipov:
I let myself introduce more ugly, but more safe solution: add a new click method that will work with invisible elements. E.g. `$().superclick`
Pros:
- more safe solution - users that don't run into issue with css magic will not be able to shout in his legs when they will click to invisible element by mistake in locator(for me it is most common case when visibilty check safes our nerve cells).
- it is clear in the code that is edge case

Cons:
- new method and additional effort to support
- ugly method name

rosolko:
@BorisOsipov @asolntsev 
If add some new method I think it would be better to give more obvious name, something like `clickIgnoreVisibility` for example.

BorisOsipov:
>If add some new method I think it would be better to give more obvious name,

@rosolko Sure. I've called it `superclick` just for example

asolntsev:
@rosolko @BorisOsipov @gedankennebel @zolrath I like it, but name seems to be invalid. There should not be two verbs in a method name. Probably
* `clickIgnoringVisibility()`
* `click(false)`
* `forcedClick()`
* `mayTheForceBeWithClick()`
* `clickThrough()`

more ideas?

BorisOsipov:
@asolntsev does the new method name fit you? It will be nice to merge this.

symonk:
LGTM

vinogradoff:
I am against adding a method for an action a user cannot do himself in browser

jkromski:
I created test for no-size element click in pr below please take a look and merge it with click-hidden-elements if it ok with you
https://github.com/selenide/selenide/pull/879

rosolko:
implemented in #874

PR #859, opened 2018-11-10
Added attribute contains condition

ostap-oleksyn:
## Proposed changes
Added condition to check for attribute to contain a certain value. 

## Checklist

### DIFF
--- a/src/main/java/com/codeborne/selenide/Condition.java
+++ b/src/main/java/com/codeborne/selenide/Condition.java
@@ -140,6 +140,34 @@ public abstract class Condition {
     };
   }
 
+  /**
+   * Check if element attribute contains given value
+   * <p>Case sensitive</p>
+   *
+   * <p>Sample: <code>$("#task").shouldHave(valueInAttribute("class", "done"));</code></p>
+   *
+   * @param attributeName name of attribute, not null
+   * @param value attribute value, not null
+   * @return true if attribute contains given value
+   */
+  public static Condition valueInAttribute(String attributeName, String value) {
+    return new Condition("attribute") {
+      @Override
+      public boolean apply(Driver driver, WebElement element) {
+        return getAttributeValue(element, attributeName).contains(value);
+      }
+    };
+  }
+
+  /**
+   * Synonym for {@link #valueInAttribute(String, String)} - may be used for better readability:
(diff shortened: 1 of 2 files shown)


BorisOsipov:
LGTM. In fact I don't understand why you need to check what particular attribute contains specific substring.(exclude class\value attrs, we already have corresponding method for them and it makes sense).

BorisOsipov:
>Check /button.png" ...>

I rework your example /button.png" ...>
And your test is green but nothing shows in app.

I am not sure it is really need to add to the code base. I prefer to just add this to https://github.com/selenide/selenide/wiki/Community-conditions

rosolko:
@BorisOsipov Good point. Let's move it to wiki?

ostap-oleksyn:
@BorisOsipov I had a case where there were a group of input fields that had different class names, but when you input text in that field, each input class would receive an additional class value 'done'. In my case I needed to fill all those fields in a loop and wait for each of those inputs to contain a 'done' substring in class attribute. 
Additionally regular selenium expected conditions have attribute contains condition, not sure why we can't add one.

vinogradoff:
Yes,  but I can probably build some nasty false negatives for every test
case :) idea is that test should not be dependent on what temporary folder
button.png is located, but just that button.png is displayed

Viele Grüße

Alexei Vinogradov

BorisOsipov:
@ostap-oleksyn 

>In my case I needed to fill all those fields in a loop and wait for each of those inputs to contain a 'done' substring in class attribute.

I guess [Condition.cssClass](https://github.com/selenide/selenide/blob/c9d39e99309d575d4a91c00d00cc2638ad12bdad/src/main/java/com/codeborne/selenide/Condition.java#L334) does exactly what you need. isn't it?

BorisOsipov:
@ostap-oleksyn @rosolko @vinogradoff guys, finally, will we merge this or save as community condition in our wiki?

rosolko:
@BorisOsipov One point for wiki

asolntsev:
@BorisOsipov @rosolko At a first glance, I would probably vote for wiki because it's a very rarely used case. For example, I never needed such a check in my projects. 

@ostap-oleksyn Thank you for the contribution!

PR #869, opened 2018-11-29
store exception details in fields in ElementShould and ElementShouldNot

agdula:
It would be useful to have access to the information that is passed to
the exception constructor. Currently there is no way to access it as raw
data.

In my case I would use that data to create a custom description of what
happened as the default message does not fit my use case.

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@
 [![Maven Central](https://maven-badges.herokuapp.com/maven-central/com.codeborne/selenide/badge.svg)](https://maven-badges.herokuapp.com/maven-central/com.codeborne/selenide)
 [![Sonar Quality Gate](https://sonarcloud.io/api/project_badges/measure?project=com.codeborne%3Aselenide&metric=alert_status)](https://sonarcloud.io/dashboard?id=com.codeborne%3Aselenide)
 [![Sonar Coverage](https://sonarcloud.io/api/project_badges/measure?project=com.codeborne%3Aselenide&metric=coverage)](https://sonarcloud.io/dashboard?id=com.codeborne%3Aselenide)
-[![MIT License](http://img.shields.io/badge/license-MIT-green.svg)](https://github.com/codeborne/selenide/blob/master/LICENSE)
+[![MIT License](http://img.shields.io/badge/license-MIT-green.svg)](https://github.com/selenide/selenide/blob/master/LICENSE)
 ![Free](https://img.shields.io/badge/free-open--source-green.svg)
 
 [![Join the chat at https://gitter.im/codeborne/selenide](https://img.shields.io/badge/welcome%20to-chat-green.svg)](https://gitter.im/codeborne/selenide?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge&utm_content=badge)
@@ -28,17 +28,17 @@ Selenide is based on and is compatible to Selenium WebDriver 2.0 and 3.0
       $("#username").shouldHave(text("Hello, Johny!"));
     }
 
-Look for [detailed comparison of Selenide and Selenium WebDriver API](https://github.com/codeborne/selenide/wiki/Selenide-vs-Selenium).
+Look for [detailed comparison of Selenide and Selenium WebDriver API](https://github.com/selenide/selenide/wiki/Selenide-vs-Selenium).
 
 ## Changelog
 
-Here is [CHANGELOG](https://github.com/codeborne/selenide/blob/master/CHANGELOG.md)
+Here is [CHANGELOG](https://github.com/selenide/selenide/blob/master/CHANGELOG.md)
 
 ## How to start?
 
 Just put selenide.jar to your project and import the following methods: `import static com.codeborne.selenide.Selenide.*;`
 
-Look for [Quick Start](https://github.com/codeborne/selenide/wiki/Quick-Start) for details.
+Look for [Quick Start](https://github.com/selenide/selenide/wiki/Quick-Start) for details.
 
 ## Resources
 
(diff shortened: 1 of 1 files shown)


agdula:
The failure looks like a flaky test. "Failed to submit a listener notification task. Event loop shut down..."
To me it's a pass :)

asolntsev:
@agdula Thank you for the contribution!
1. Can you give an example of your error message? Probably it's a good idea to use the new format in Selenide too?
2. I personally don't like useless boilerplate code like setters/getters. Instead of adding million of private fields and public getters, I suggest adding `protected final` fields. Thus, only subclasses can read them, and nobody can modify.

agdula:
1. I'm PERFECTLY fine with what Selenide is showing as an exception explanation BUT There are people in my team that find it hard to read in comparison with what classic assertion would show. For example if you're making condition on a text of element you get
`
checkApplicantsHeader(com.***.smoke.SideNavTest): run 1 failed: 'Element should have textCaseSensitive '1 Applicants (3)' {By.xpath: //app-applicants/h2}
Element: '<h2>Applicants (3)</h2>''
com.codeborne.selenide.ex.ElementShould: Element should have textCaseSensitive '1 Applicants (3)' {By.xpath: //app-applicants/h2}
Element: '<h2>Applicants (3)</h2>'
`
You get expected text and the whole node which in their opinion clouds the image. 
It is even worse for them in case you're checking for a tag value. This prints the whole tag and they complain they have to look for the actual tag an then to search for the value.

What I was going to do is to add a log entry before the original one so that would extract the expected value from the data in exception so that they would not complain that they have to search through all of the tags. 

IF you think it makes sense to change how Selenide is showing values I'm happy to make a different PR.

2. Agreed. I this change makes any sense to you I'll comply.

PS.

They also asked me to add terminal colors to the errors in TextReport output. I managed to do this without modifying Selenide. If you wish to see and maybe use it in Selenide let me know. I can create PR for this too.

asolntsev:
@agdula Could you bring an example of "readable" log? 
It's probably a totally valid idea to update the built-in reporting style in Selenide. 

P.S. It's interesting idea about colors. Could you share your code snippet somehow? I have no idea how to do such things...

agdula:
@asolntsev Regarding Colors. Here is [gist ](https://gist.github.com/dainkaplan/4651352) I've found that you can try out `Ansi.colorize`

If it comes to the "readable" log. I'm working on it.

agdula:
Here is the readable log Ultimate(-ish) solution :) https://github.com/agdula/selenide/commit/3635ee8158dd2668de2c69f777b090eb21e0376c
```java
COMPONENT_CONTAINER.shouldHave(attribute("name","test-name"))
```
in the exception description you can see what is what is wrong without putting too much effort because it's always in the same place
```log
2018-12-08 14:50:04 [main] WARN  com.*** - check***(com.***Test): run 3 failed: '
Should Have name:
	Expected: 'test-name'
	  Actual: ''

Element should have attribute name=test-name {By.xpath: //mat-tab-body//app-summary-page}
Element: '<app-summary-page id="someid" aria-hidden="true" autocomplete="off" name="q" tabindex="-1" type="url" jsaction="mousedown:ntp.fkbxclk" style="opacity: 0;" class="ng-star-inserted">Lorem ipsum dolor sit amet, consectetur adip
iscing elit, sed do eiusmod tempor
 incididunt ut labore et dolore
  magna aliqua. Ut enim ad minim veniam, quis 
 nostrud exercitation ullamco laboris n
 isi ut aliquip ex ea commodo consequat. Duis
  aute irure dolor in reprehenderit in volup
  tate velit esse cillum dolore eu fugiat 
  nulla pariatur. Excepteur 
  sint occaecat cupidatat
   non proident, sunt in culpa qui
   officia deserunt mollit 
   anim id est laborum.
</app-summary-page>''
```

I would argue that the `Should Have name` could be more like `Should Have attribute 'name'` It's a matter of taste.

Same in case of this
```java
TOTAL_ADJUSTED_SUM.shouldHave(exactText("TOTAL  1 ADJUSTED SUM"));
```


```log
2018-12-08 14:50:42 [main] ERROR c.c.t.c.recovery.LoggingRunListener - FAIL: checkTotalAdjIncomeTextLoad(com.***.Test): 
Should Match Text (Case Insensitive, Spaces Normalized):
	Expected: 'total 1 adjusted sum'
	  Actual: 'total adjusted sum'

Element should have exact text 'TOTAL  1 ADJUSTED SUM' {By.xpath: .//h4}
Element: '<div  id="someid" aria-hidden="true" autocomplete="off" name="q" tabindex="-1" type="url" jsaction="mousedown:ntp.fkbxclk" style="opacity: 0;" class="ng-star-inserted">TOTAL ADJUSTED SUM</div>'
```

If you think this makes sense I can convert the draft idea into a real Pull Request

agdula:
@asolntsev is the proposed solution sensible ?

asolntsev:
@agdula Sorry for the delay, I will get back soon.

asolntsev:
@agdula Sorry for the long-long delay.
Now I got your idea. Yes, your error message looks good. 

So, the idea is ok, but I have one more question. Why you expose fields only in `ElementShould` and `ElementShouldNot`? What about other errors like `ElementNotFound`, `TextsMismatch` etc.? Probably we should create all those fields in their common superclass `UIAssertionError`?

agdula:
@asolntsev It's been a while :)

asolntsev:
@agdula Wow. Sometimes it's better to just wait a little bit. :)

So, are you going to send the new change with another pull request?

agdula:
the code base has unfortunately changed to the point that I'd have to re-do
all the changes :/. I'll try to find some time to make it work.

niedz., 14 lip 2019 o 23:20 Andrei Solntsev <notifications@github.com>
napisał(a):

Issue #865, opened 2018-11-19
Add aliases for $ and $$ for Kotlin


nbdnnm:
Hello,

Kotlin doesn't support $ and $$ without \` \`, so how about to add aliases, like S and SS or something else?

d-astanin:
For your kotlin project you can use code like that
```
fun s(locator: String): SelenideElement {
    return Selenide.`$`(locator)
}

fun ss(locator: String): ElementsCollection {
    return Selenide.`$$`(locator)
}
```

vinogradoff:
…which looks ugly


I would also support aliases, e.g. S/SS or s/ss or something else.

Viele Grüße

Alexei Vinogradov

On 19. November 2018 at 13:15:38, nbdnnm (notifications@github.com) wrote:

The problem

Briefly describe the issue you are experiencing (or the feature you want to
see added to Selenide).
Tell us what you were trying to do and what happened instead.
Details

If necessary, describe the problem you have been experiencing in more
detail.
Tell us about your environment

   - *Selenide Version:*
   - *Chrome\Firefox\IE Version:*
   - *Browser Driver Version:*
   - *Selenium Version:*
   - *OS Version:*

Code To Reproduce Issue [ Good To Have ]

Please remember that with sample code it's easier to reproduce the bug.

—
You are receiving this because you are subscribed to this thread.
Reply to this email directly, view it on GitHub
<https://github.com/selenide/selenide/issues/865>, or mute the thread

evgeny-tumashchick:
Hi folks, actually I use Selenide in my Kotlin project.
I call mentioned aliases like this:
`$`(fieldLocator)   / `$x`(fieldLocator)
`$$`(fieldLocator)   / `$$x`(fieldLocator)

jkromski:
wouldn't be better to follow naming from SelenideElement which is `find` for $?

vinogradoff:
Not quite. $() doesn't actually do any search yet. Find is misleading then.

Viele Grüße

Alexei Vinogradov

jkromski:
Interface says it is the same
```java
/**
   * ATTENTION! This method doesn't start any search yet!
   * Same as {@link #find(String)}
   * @see com.codeborne.selenide.commands.Find
   */
  SelenideElement $(By selector);
```
```java
/**
   * ATTENTION! This method doesn't start any search yet!
   * Same as {@link #find(String)}
   * @see com.codeborne.selenide.commands.Find
   */
  SelenideElement find(By selector);
```

[...]

PR #870, opened 2018-11-30
865 aliases for kotlin

jkromski:
## Proposed changes
- add find, findX, findAll, findAllX methods in Selenide class
- deprecate Selenide.getElement* methods as they are the same as find* and they don't follow naming convention from SelenideDriver or SelenideElement

https://github.com/selenide/selenide/issues/865

## Checklist
- [ ] Checkstyle and unit tests pass locally with my changes by running `gradle check chrome htmlunit` command - **there are error not related with this change**


asolntsev:
@jkromski @rosolko @BorisOsipov 
1. I don't like `findX` and `findAllX`, but
2. I like `find` and `findAll`

I think we can add methods `find` and `findAll` - they can be used in Kotlin, why not. It's better than `s`, `S` and other meaningless suggestions. Why not?

vinogradoff:
I don’t like find() because SelenideElement x=find(“…); doesn’t search and
doesn’t find and it is misleading.

Viele Grüße

Alexei Vinogradov

On 7. December 2018 at 23:42:08, Andrei Solntsev (notifications@github.com)
wrote:

@jkromski <https://github.com/jkromski> @rosolko
<https://github.com/rosolko> @BorisOsipov <https://github.com/BorisOsipov>

   1. I don't like findX and findAllX, but
   2. I like find and findAll

I think we can add methods find and findAll - they can be used in Kotlin,
why not. It's better than s, S and other meaningless suggestions. Why not?

—
You are receiving this because you are subscribed to this thread.
Reply to this email directly, view it on GitHub
<https://github.com/selenide/selenide/pull/870#issuecomment-445388233>, or mute
the thread

vinogradoff:
It actually defines how it should be searched/located/found.


Lets try another names?

Viele Grüße

Alexei Vinogradov

On 7. December 2018 at 23:47:06, Alexei Vinogradov (alexei@vinogradoff.de)
wrote:

I don’t like find() because SelenideElement x=find(“…); doesn’t search and
doesn’t find and it is misleading.

Viele Grüße

Alexei Vinogradov

On 7. December 2018 at 23:42:08, Andrei Solntsev (notifications@github.com)
wrote:

@jkromski <https://github.com/jkromski> @rosolko
<https://github.com/rosolko> @BorisOsipov <https://github.com/BorisOsipov>

   1. I don't like findX and findAllX, but
   2. I like find and findAll

I think we can add methods find and findAll - they can be used in Kotlin,
why not. It's better than s, S and other meaningless suggestions. Why not?

—
You are receiving this because you are subscribed to this thread.
Reply to this email directly, view it on GitHub
<https://github.com/selenide/selenide/pull/870#issuecomment-445388233>, or mute
the thread

jkromski:
if you don't like `find`, what other names you got? Propose something please. In my opinion it should be verb in simple present form

@asolntsev I removed findX and findAllX

BTW. @vinogradoff Could you please don't add previous messages in yours, it creates noise. It is harder to read through what you exactly mean.

vinogradoff:
I don't like verbs here. This method defines how element will be searched. But it doesn't search. 

We have element() or el() as other candidates...

jkromski:
looks good to me. It seems that it doesn't have to be a verb then.
so methods would look like:
`element("a.link").shouldBe(visible);`
and 
`elements("a.link").shouldBe(visible);`
@asolntsev and @BorisOsipov what do you think?

asolntsev:
@jkromski I still don't see any problems with `find` and `findAll`. 
`element`/`elements` is also a good option (except that it's too long). 

Actually Selenide already has very similar methods `getElement`/`getElements`. 
Does changing `getElement` to `element` really matter? I am not sure...
So, by the moment, I suggest to make a pause until we find some shorter name or agree on `find`.

vinogradoff:
lol, actually never knew about getElement (est. 2013) which is alias to $.

But element looks preciser. Still I would deprecate getElement() for both
element() and el() as a shorter version of “element”


$() is by far most used function in Selenide- it makes sense to have a
short version .



Viele Grüße

Alexei Vinogradov

On 12. December 2018 at 22:56:35, Andrei Solntsev (notifications@github.com)
wrote:

@jkromski <https://github.com/jkromski> I still don't see any problems with
find and findAll.
element/elements is also a good option (except that it's too long).

Actually Selenide already has very similar methods getElement/getElements.
Does changing getElement to element really matter? I am not sure...
So, by the moment, I suggest to make a pause until we find some shorter
name or agree on find.

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/selenide/selenide/pull/870#issuecomment-446759245>, or mute
the thread

vinogradoff:
Jacek, actually we don't have a use case for "search" an element, you need element either to interact with it or to assert some of his properties, but not just to "search".

so `$(element).shouldBe(visible).$(subelement)` -> is not complete.
---
Analogy. Let's say you have WYSIWYG palette in Delphi/VB/Visual C++ whatever IDE. 
Then you drop widgets like Buttons, input fields, labels and so on your canvas, connecting them with some operations. Our browser view is analogy for canvas with widgets.

Then you code something. You probably need to interact with some widgets on the canvas (e.g. `button1.Click()`) or read their properties (e.g. `inputField1.getText()`). 
There is though, no use case - like `search` button1 or inputField1 on the canvas. It is not necessary to search button1 and inputField1, they are declared and can be referenced when you want to use them.

I want to write tests the same way. Element declared with their locating methods/expressions, and then used for interacting or assertions. Luckily it is also the way Selenide works now.

That's why I don't want to have declaration to look like search operation.

vinogradoff:
Then I take your examples, but with arguments right above. I just use `element` but i could be anything we agree on.

To declare an element use:
`SelenideElement container=element("div.container");`

To declare an element inside another element use:
`SelenideElement link=element("div.container").element("a.link");` or
`SelenideElement link=container.element("a.link");`

To interact with previously defined element use:
`link.click();`

To assert properties of previously defined element use:
`container.shouldBe(visible);`

Does it make sense?

jkromski:
Can we agree on what to do over here?
I would do **a** option now. Option **b** we can move to other ticket and if we agree, do it for v5.2 or v6.0.

But I can do what you decide/agree

asolntsev:
@jkromski it seems that we don't have an "ideal" solution that would be better than current state. 
So, it's probably good idea just to do nothing and stay where we are (until we get a new idea). 

P.S. I found some symbols that are allowed in Kotlin. Probably we could use on of them instead of `$`?

```kotlin
        Œ(".active")
        Š(".active")
        ƒ(".active")
        œ(".active")
        Ÿ(".active")
        ª(".active")
        µ(".active")
        À(".active")
```

jkromski:
the symbols you gave I cannot find on the keyboard and I guess most of the people won't so it will be pain in the ... to use it in any language. 
From what I'm guessing Selenide uses $ because it became some kind of a standard which came from JS frameworks, see https://api.jquery.com/jQuery and these frameworks have workaround for $ which is an alias in jQuery it is `jQuery()` and `find()`

asolntsev:
Since we haven't a common opinion now, I suggest to avoid doing anything
(until we find a good idea that we all like).

jkromski:
To be honest, right now after doing some more tests, I think @vinogradoff proposition make a lot of sense as it fits scenarios which I mentioned before and if you would like to assure that element does not exists then you got:
`element("a.link").shouldNot(exist)`
which is very intuitive.

`find("a.link").shouldNot(exist)` does not make much sense here

asolntsev:
Wow.
This is really a great argument. I absolutely agree that both
`find("a.link").shouldNot(exist)` and
`getElement("a.link").shouldNot(exist)` do not sound good, while
`element("a.link").shouldNot(exist)` is perfect.

So, let's create method `element` and deprecate method `getElement`.

And what about collections? Does `elements("a.link").shouldBe(empty)` sound
good?  And `elements("a.link").shouldHave(size(5))`?

jkromski:
yup, collections looks good too. Shall I mark find* methods deprecated too or we leave it for later?

rosolko:
@jkromski Just introduce `element/elements` and deprecate `getElement/getElements`

vinogradoff:
I have another idea for collections. 

Actually 
elements("a.link").shouldHaveSize(3)
 or 
all("a.link").shouldHaveSize(3)
both not perfect 

This looks roght:
collection("a.link").shouldHaveSize(3)
or even
collectionOf("a.link").shouldHaveSize(3)

Downside: these are pretty long names. 
Perhaps that helps to find more better ideas?

jkromski:
@vinogradoff personally I prefer element/elements. It was long discussion and I think at the end your proposition was the best. Pull request is ready to be re-reviewed

vinogradoff:
I am okay with elements

vinogradoff:
let's also deprecate find?

PR #872, opened 2018-11-30
resolve issue with executing tests on statics

jkromski-fh:
## Proposed changes
on Intellij I get test folder marked as source root, not test source root. This results with a lot of issues showing in Editor: cannot resolve method assert*

## Checklist
- [ ] Checkstyle and unit tests pass locally with my changes by running `gradle check chrome htmlunit` command **(other errors)**
- [ ] I have added tests that prove my fix is effective or that my feature works **not related**
- [ ] I have added necessary documentation (if appropriate) **not related**

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -9,6 +9,8 @@ plugins {
   id 'org.sonarqube' version '2.6.2'
 }
 
+apply plugin: 'idea'
+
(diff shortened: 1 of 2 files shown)


jkromski:
thanks for the correction

PR #874, opened 2018-12-05
Fixing inconsistency with transparent elements

vinogradoff:
fix for #201. Now Selenide can interact with element that have opacity:0 the same way as browser users can.

## Checklist


BorisOsipov:
I have to mention one disadvantage for this approach - it requires one more request to browser on each interaction. So we will decrease perfomance for end users.

vinogradoff:
Very good point. Actually it should only do the second call if visible is
false - which is okay. If Condition.or doesn't evaluate further conditions
after any conditions evaluates true. Could you please check or method?
Viele Grüße

Alexei Vinogradov

BorisOsipov:
>If Condition.or doesn't evaluate further conditions
after any conditions evaluates true. Could you please check or method?
Yep It works as expected.

vinogradoff:
Thanks! Then the fix makes not  slower as it is now, because now it makes 2
checks only when original version would fail the test.

The next step would be try to drop the both checks and just do click
relying on underlying Selenium behavior (though catching and wrapping
exceptions for readable diagnostics).

But I see it as separate piece of work. Do you agree?


Viele Grüße

Alexei Vinogradov

vinogradoff:
Lets put discussion to new issue?
I'll create it.

Is the fix okay for now in your opinion?)

Viele Grüße

Alexei Vinogradov

jkromski:
can you this test too please
https://github.com/selenide/selenide/pull/881

Issue #766, opened 2018-07-20
Add new feature getSelectionRange / hasSelectionRange


mseele:
## The problem
There is no method to check the selection range of text for an input text field.

## The Details
Whould be cool to integrade a method `SelenideElement.getSelectionRange` and/or `Condition.selectionRange(String selection)` to selenide.

## Tell us about your environment
* **Selenide Version:** 4.12.2
* **Selenium Version:** 3.13.0
* **OS Version:** Windows 10

## Code To Reproduce Issue [ Good To Have ]

Right now we check it like this:

```
private String getSelectionRange(SelenideElement element) {
	return executeJavaScript("return arguments[0].value.substring(arguments[0].selectionStart, arguments[0].selectionEnd);", element); //$NON-NLS-1$
}
```

symonk:
Hi @mseele can you provide a little more info? is this for the length of entered text ('value') or the maximum length value on a given field?  I will create a PR when you clarify, thanks.

vinogradoff:
What is selectionRange exactly? Characters selected with a mouse (to copy to clipboard?) inside of any element?

mseele:
It's the start and end positions of the current text selection in an <input> element.

Please see https://developer.mozilla.org/en-US/docs/Web/API/HTMLInputElement/setSelectionRange for more details.

It is, for example, usefull if you have an input field that selects the whole input on focus (by code) and you want to test this.

asolntsev:
@mseele 
1. If I understand correctly, your method `getSelectionRange` actually returns selected text (not start/end positions). Thus it should be named `getSelectedText()`. Am I right?
2. In Selenide style, we should add condition instead of just getter. Like this:

```java
$("textarea#bio").shouldHave(selectedText("London"));
$("textarea#bio").shouldHave(text("birth place: London"));
```

symonk:
given the following image, it would be returning the value for 'selected' (basically what has been selected / highlighted?

![image](https://user-images.githubusercontent.com/17887843/43411396-f786a258-9421-11e8-9528-ad6d732027dd.png)

if so, do we think this is a worthwhile feature to implement?

mseele:
@symonk you example is exactly what i meant.

asolntsev:
@symonk yes, why not. But I remind that I still prefer adding condition instead of just getter:

```java
$("textarea#bio").shouldHave(selectedText("London"));
```

asolntsev:
@symonk Hi! How is your PR going?

[...]

PR #876, opened 2018-12-05
#766 selected text condition @WIP

symonk:
## Proposed changes
Provides capabilities for selenide to read and assert against the 'selected' or 'highlighted' text of a given input field as requested in #766.

By 'highlighted' or 'selected' we mean => The text on the element that if a user was to issue a copy command, would be copied to the clipboard.

Would like some feedback on this as an approach to this feature, this PR is a work in progress.  Example output where the selected text is not matching:

```
Element should have selectedText 't2his ' {By.id: selected}
Element: '<input id="selected" type="text" value="this is a long piece of selectable text"></input>'
Actual value: 'this '
```

Thoughts? thanks for your time.

## Checklist
- [X] Checkstyle and unit tests pass locally with my changes by running `gradle check chrome htmlunit` command

### DIFF
--- a/src/main/java/com/codeborne/selenide/Condition.java
+++ b/src/main/java/com/codeborne/selenide/Condition.java
@@ -256,6 +256,31 @@ public abstract class Condition {
   }
 
   /**
+   * Checks for selected (highlighted) text on a given web element
+   *
+   * <p>Sample: {@code $("input").shouldHave(selectedText("Text"))}</p>
+   *
+   * <p>NB! Case sensitive</p>
+   *
+   * @param expectedText expected highlighted text of the element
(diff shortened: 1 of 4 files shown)


symonk:
I think WDM decided to give up on travis :(

rosolko:
@asolntsev @BorisOsipov In my opinion it's also should be move to community conditions tab.

rosolko:
@symonk create block on community condition because it's not common specific case

rosolko:
@symonk any progress?

asolntsev:
@symonk Thank you for the pull request. Merged!

Sorry for the delay.

PR #879, opened 2018-12-07
add test for click on no-size link

jkromski:


rosolko:
closed in favor of #881

Issue #867, opened 2018-11-26
Selenide 5.0.0. WebDriverRunner.setWebDriver(new ChromeDriver()) incorrect behaviour


alexmuzyka:
Code:
...
WebDriver instance1 = WebDriverRunner.getWebDriver();
WebDriverRunner.setWebDriver(new ChromeDriver());
WebDriver instance2 = WebDriverRunner.getWebDriver();
...

On Selenide 4... it worked as we expected: we have two instances, and can easely switching between them.

On Selenide 5.0.0 after WebDriverRunner.setWebDriver(new ChromeDriver()) - WebDriverRunner is closing opened browser (instance1), and create a new one. 

Can you return back this functionality?

Thanks

PS: original comment (request) from Andrei Solntsev: https://selenide.org/2018/10/10/selenide-5.0.0/

rosolko:
No, we can't.
Use a new approach for two instances:
```
SelenideDriver browser1 = new SelenideDriver();
SelenideDriver browser2 = new SelenideDriver();
```

BorisOsipov:
@alexmuzyka We suggest to use new approach as described @rosolko 

But it seems to need ahve look at this issue deeper.

asolntsev:
See PR https://github.com/selenide/selenide/pull/890

PR #890, opened 2018-12-12
#867 user can switch between custom webdrivers many times

asolntsev:
... and Selenide should not close them

Fix for issue #867

## Checklist

### DIFF
--- a/src/main/java/com/codeborne/selenide/drivercommands/WebDriverWrapper.java
+++ b/src/main/java/com/codeborne/selenide/drivercommands/WebDriverWrapper.java
@@ -40,10 +40,11 @@ public class WebDriverWrapper implements Driver {
     return webDriver;
   }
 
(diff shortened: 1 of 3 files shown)


src/main/java/com/codeborne/selenide/drivercommands/WebDriverWrapper.java
```diff
@@ -40,10 +40,11 @@ public WebDriver getAndCheckWebDriver() {
     return webDriver;
   }
 
+  /**
+   * Does not close webdriver.
+   * This class holds a webdriver created by user - in this case user is responsible for closing webdriver by himself.
+   */
   @Override
   public void close() {
```
rosolko: I think in this case we should at least log a message that driver wasn't closed automatically and user need additional code to properly close a session

PR #893, opened 2018-12-13
892 generate unique location for every downloaded file

asolntsev:
## Checklist

### DIFF
--- a/src/main/java/com/codeborne/selenide/impl/Randomizer.java
+++ b/src/main/java/com/codeborne/selenide/impl/Randomizer.java
@@ -1,0 +1,9 @@
+package com.codeborne.selenide.impl;
+
+import java.util.UUID;
+
+public class Randomizer {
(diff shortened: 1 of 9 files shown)


src/main/java/com/codeborne/selenide/impl/Randomizer.java
```diff
@@ -0,0 +1,9 @@
+package com.codeborne.selenide.impl;
+
+import java.util.UUID;
+
+public class Randomizer {
+  public String text() {
```
rosolko: Not a text but an uuid. Method signature indicates the return type is a String and in general it's a text

asolntsev: Sorry, I don't understand. The intention of this method is to generate random text. The fact that it uses UUID is just an implementation detail.

PR #896, opened 2018-12-20
Fix to close the browser with new SelenideDriver().close()

vinogradoff:
the webdriver (and so the browser) wasn't actually closed.

## Checklist

### DIFF
--- a/src/main/java/com/codeborne/selenide/SelenideDriver.java
+++ b/src/main/java/com/codeborne/selenide/SelenideDriver.java
@@ -136,11 +136,11 @@ public class SelenideDriver {
   }
 
   public WebDriver getWebDriver() {
-    return driver.getWebDriver();
+    return driver().getWebDriver();
   }
 
   public WebDriver getAndCheckWebDriver() {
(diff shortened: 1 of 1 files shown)


vinogradoff:
Yes, it looks like a deeper look is necessary, some tests now failing with IllegalStateException 

It didn't close the browser created with customer code
```
browser=new SelenideDriver(....new FirefoxDriver(options)....);
browser.close();
```

For example in FirefoxWithProfileTest

vinogradoff:
Okay, now I understand what is going on:

WebDriverWrapper is used for custom drivers, having this code:
```
 /**
   * Does not close webdriver.
   * This class holds a webdriver created by user - in this case user is responsible for closing webdriver by himself.
   */
  @Override
  public void close() {
  }
```

So at the moment it "works as intended", but perhaps let's find the way to help users who creates the driver themselves to easier understand how to work with it.

Let's chat?

asolntsev:
@vinogradoff superseded by PR https://github.com/selenide/selenide/pull/989. Thank you for the effort.

PR #904, opened 2019-01-30
Including condition explanation into CollectionCondition toString()

rkliuha:
## Proposed changes
These changes add condition explanation into CollectionCondition toString() as it is in single Condition.
It's needed for better events logging because LogEvent.toString() for collections verification does not include explanation message (because)

## Checklist

### DIFF
--- a/src/main/java/com/codeborne/selenide/CollectionCondition.java
+++ b/src/main/java/com/codeborne/selenide/CollectionCondition.java
@@ -1,6 +1,14 @@
 package com.codeborne.selenide;
 
-import com.codeborne.selenide.collections.*;
+import com.codeborne.selenide.collections.ExactTexts;
+import com.codeborne.selenide.collections.ListSize;
+import com.codeborne.selenide.collections.SizeGreaterThan;
+import com.codeborne.selenide.collections.SizeGreaterThanOrEqual;
+import com.codeborne.selenide.collections.SizeLessThan;
+import com.codeborne.selenide.collections.SizeLessThanOrEqual;
+import com.codeborne.selenide.collections.SizeNotEqual;
+import com.codeborne.selenide.collections.Texts;
+import com.codeborne.selenide.collections.TextsInAnyOrder;
 import com.codeborne.selenide.impl.WebElementsCollection;
 import com.google.common.base.Predicate;
 import org.checkerframework.checker.nullness.compatqual.NullableDecl;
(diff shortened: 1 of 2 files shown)


src/main/java/com/codeborne/selenide/CollectionCondition.java
```diff
@@ -1,16 +1,9 @@
...
-import com.codeborne.selenide.collections.ExactTexts;
-import com.codeborne.selenide.collections.ListSize;
-import com.codeborne.selenide.collections.SizeGreaterThan;
-import com.codeborne.selenide.collections.SizeGreaterThanOrEqual;
-import com.codeborne.selenide.collections.SizeLessThan;
-import com.codeborne.selenide.collections.SizeLessThanOrEqual;
-import com.codeborne.selenide.collections.SizeNotEqual;
-import com.codeborne.selenide.collections.Texts;
-import com.codeborne.selenide.collections.TextsInAnyOrder;
+import com.codeborne.selenide.collections.*;
```
rosolko: Why don't you run gradle `check` task locally?
This import changes spawn an error:
```
[ant:checkstyle] [ERROR] /home/travis/build/selenide/selenide/src/main/java/com/codeborne/selenide/CollectionCondition.java:3: Using the '.*' form of import should be avoided - com.codeborne.selenide.collections.*. [AvoidStarImport]
```

src/main/java/com/codeborne/selenide/CollectionCondition.java
```diff
@@ -103,12 +96,42 @@ public static CollectionCondition exactTexts(List<String> expectedTexts) {
     return new ExactTexts(expectedTexts);
   }
 
+  private static class ExplainedCollectionCondition extends CollectionCondition {
```
rosolko: Also new class and method lack of documentation.

rosolko:
@rkliuha Please take a look at the pull request check list

src/main/java/com/codeborne/selenide/CollectionCondition.java
rkliuha: I've encountered with some checkstyle issue while running it locally. Now It's fixed. Thank you

src/main/java/com/codeborne/selenide/CollectionCondition.java
rkliuha: Added overall explanation

rkliuha:
@rosolko Thank you for review! Tests on CI are still failing but I'm not sure that it's my changes

symonk:
I don't think your changes are breaking tests.  Firefox on travis has been problematic for quite a while, we should fix it :) chrome is surprisingly failing on proxy/download/auth related issues

PR #912, opened 2019-02-19
extract Conditions to separate classes

asolntsev:
## Proposed changes
split big `Condition` class to many small classes.

## Checklist

### DIFF
--- a/src/main/java/com/codeborne/selenide/conditions/AttributeWithValue.java
+++ b/src/main/java/com/codeborne/selenide/conditions/AttributeWithValue.java
@@ -1,0 +1,31 @@
+package com.codeborne.selenide.conditions;
+
+import com.codeborne.selenide.Condition;
+import com.codeborne.selenide.Driver;
+import org.openqa.selenium.WebElement;
+
+public class AttributeWithValue extends Condition {
+  private final String attributeName;
+  private final String expectedAttributeValue;
+
+  public AttributeWithValue(String attributeName, String expectedAttributeValue) {
+    super("attribute");
+    this.attributeName = attributeName;
+    this.expectedAttributeValue = expectedAttributeValue;
+  }
+
+  @Override
+  public boolean apply(Driver driver, WebElement element) {
+    return expectedAttributeValue.equals(getAttributeValue(element, attributeName));
+  }
+
+  @Override
+  public String toString() {
+    return getName() + " " + attributeName + '=' + expectedAttributeValue;
+  }
+
+  private String getAttributeValue(WebElement element, String attributeName) {
+    String attr = element.getAttribute(attributeName);
+    return attr == null ? "" : attr;
+  }
+}
--- a/src/main/java/com/codeborne/selenide/conditions/CssClass.java
+++ b/src/main/java/com/codeborne/selenide/conditions/CssClass.java
@@ -1,0 +1,34 @@
+package com.codeborne.selenide.conditions;
+
+import com.codeborne.selenide.Condition;
+import com.codeborne.selenide.Driver;
+import org.openqa.selenium.WebElement;
+
+public class CssClass extends Condition {
+  private final String expectedCssClass;
+
+  public CssClass(String expectedCssClass) {
+    super("css class");
+    this.expectedCssClass = expectedCssClass;
+  }
+
+  @Override
+  public boolean apply(Driver driver, WebElement element) {
+    String actualCssClasses = element.getAttribute("class");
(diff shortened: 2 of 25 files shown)


src/main/java/com/codeborne/selenide/conditions/CssClass.java
```diff
@@ -0,0 +1,34 @@
...
+    String actualCssClasses = element.getAttribute("class");
+    return actualCssClasses != null && contains(actualCssClasses.split(" "), expectedCssClass);
+  }
+
+  @Override
+  public String toString() {
+    return getName() + " '" + expectedCssClass + '\'';
+  }
+
+  private <T> boolean contains(T[] objects, T object) {
```
muescha: this generic should better go into a `util` or `helper` class

src/main/java/com/codeborne/selenide/conditions/AttributeWithValue.java
```diff
@@ -0,0 +1,31 @@
...
+  }
+
+  @Override
+  public String toString() {
+    return getName() + " " + attributeName + '=' + expectedAttributeValue;
+  }
+
+  private String getAttributeValue(WebElement element, String attributeName) {
+    String attr = element.getAttribute(attributeName);
+    return attr == null ? "" : attr;
```
muescha: this can also be into a `util` or `helper` class 
or just use `org.apache.commons.lang3.StringUtils.defaultString`

src/main/java/com/codeborne/selenide/conditions/CssClass.java
muescha: or use 

[`ArrayUtils.contains(Object[] array, Object objectToFind)`](https://commons.apache.org/proper/commons-lang/apidocs/org/apache/commons/lang3/ArrayUtils.html#contains-java.lang.Object:A-java.lang.Object-)

src/main/java/com/codeborne/selenide/conditions/Value.java
```diff
@@ -0,0 +1,30 @@
...
+  }
+
+  @Override
+  public String toString() {
+    return getName() + " '" + expectedValue + "'";
+  }
+
+  private String getValueAttribute(WebElement element) {
+    String attr = element.getAttribute("value");
+    return attr == null ? "" : attr;
```
muescha: `StringUtils.defaultString`

src/main/java/com/codeborne/selenide/conditions/SelectedText.java
```diff
@@ -0,0 +1,33 @@
...
+  }
+
+  @Override
+  public String actualValue(Driver driver, WebElement element) {
+    return "'" + actualResult + "'";
+  }
+
+  @Override
+  public String toString() {
+    return getName() + " '" + expectedText + '\'';
```
muescha: why sometimes `"'"` and sometimes `'\''`?

src/main/java/com/codeborne/selenide/conditions/CssClass.java
asolntsev: @muescha Thank you for your comments! I will try to make these improvements in a next pull request.

asolntsev: @muescha Finally, I have reviewed your comments. 
Initially Selenide didn't depend on commons-lang.jar. I didn't want to introduce a new dependency just because of 2 tiny methods. That's why I wrote those 2 methods by myself (`ArrayUtils.contains` and  `StringUtils.defaultString`).

Now Selenide depends on commons-lang.jar, but not directly (through BMP and HtmlUnit). Most of Selenide users automatically get commons-lang.jar, but there can be some users who exclude BMP and HtmlUnit, and they would also exclude commons-lang.jar.

I am not sure if this is a strong argument to avoid using commons-lang.jar directly in Selenide code?...

src/main/java/com/codeborne/selenide/conditions/SelectedText.java
asolntsev: @muescha  It's because IDEA sometimes suggest to replace one-characters String (like `"x"`) to a character (like `'x'`). I guess it's some kind of memory optimization. 

In this case, optimization is not needed, and `"'"` is more readable than `'\''`. So I replaces `'\''` everywhere in the code (see https://github.com/selenide/selenide/commit/1dbb92915844d6190e1d7f56e4e61602292852dc).

PR #927, opened 2019-04-04
Log event listener improvements

pavelpp:
## Proposed changes
Currently SelenideLogger only logs stuff after each event happened, so in order to apply some logic before some action happens we have to resort to implementing custom WebDriverEventListener, which does not support all the methods that we might be interested in. 

Example: I want to log click() methods before click happened. Currently I can not do it, as onEvent logs after event happened.

Added beforeEvent() method to LogEventListener and renamed onEvent to afterEvent to make it clear and consistent. beforeEvent() will be called in SelenideLogger.beginStep(...) method, which is used in SelenideElementProxy.invoke(...) method before dispatchAndRetry(), which will ensure the event is logged before action.

## Checklist

### DIFF
--- a/src/test/java/integration/LogEventListenerTest.java
+++ b/src/test/java/integration/LogEventListenerTest.java
@@ -1,0 +1,70 @@
+package integration;
+
+import com.codeborne.selenide.SelenideConfig;
+import com.codeborne.selenide.SelenideDriver;
+import com.codeborne.selenide.SelenideElement;
+import com.codeborne.selenide.logevents.LogEvent;
+import com.codeborne.selenide.logevents.LogEventListener;
+import com.codeborne.selenide.logevents.SelenideLogger;
+
+import org.assertj.core.api.SoftAssertions;
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.Test;
+
+import java.util.ArrayList;
+import java.util.List;
+
+import static com.codeborne.selenide.Condition.visible;
+import static java.lang.String.format;
+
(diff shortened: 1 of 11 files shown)


src/test/java/integration/LogEventListenerTest.java
```diff
@@ -0,0 +1,70 @@
...
+  @AfterEach
+  void tearDown() {
+    driver.close();
+  }
+
+  @Test
+  void canCatchBeforeAndAfterEvents() {
+    SelenideLogger.addListener("log events collector", new SelenideListener());
+
+    beforeEvents = new ArrayList<>();
```
rosolko: ditto

src/test/java/integration/LogEventListenerTest.java
```diff
@@ -0,0 +1,70 @@
...
+    afterEvents = new ArrayList<>();
```
rosolko: ditto

src/test/java/integration/LogEventListenerTest.java
```diff
@@ -0,0 +1,70 @@
...
+import org.junit.jupiter.api.Test;
+
+import java.util.ArrayList;
+import java.util.List;
+
+import static com.codeborne.selenide.Condition.visible;
+import static java.lang.String.format;
+
+public class LogEventListenerTest extends BaseIntegrationTest {
+  SelenideDriver driver = new SelenideDriver(new SelenideConfig().baseUrl(getBaseUrl()));
```
rosolko: Move it to before block and made private

PR #931, opened 2019-04-24
upgrade to selenium-java 4.0.0-alpha-2

asolntsev:
## Proposed changes

## Checklist

### DIFF
--- a/src/main/java/com/codeborne/selenide/CollectionCondition.java
+++ b/src/main/java/com/codeborne/selenide/CollectionCondition.java
@@ -17,7 +17,6 @@ import com.google.common.base.Predicate;
 import org.checkerframework.checker.nullness.qual.Nullable;
 import org.openqa.selenium.WebElement;
 
-import javax.annotation.Nullable;
 import java.util.List;
 
 public abstract class CollectionCondition implements Predicate<List<WebElement>> {
--- a/src/main/java/com/codeborne/selenide/CollectionCondition.java
+++ b/src/main/java/com/codeborne/selenide/CollectionCondition.java
@@ -17,6 +17,7 @@ import com.google.common.base.Predicate;
 import org.checkerframework.checker.nullness.qual.Nullable;
 import org.openqa.selenium.WebElement;
 
+import javax.annotation.Nullable;
 import java.util.List;
 
 public abstract class CollectionCondition implements Predicate<List<WebElement>> {
--- a/gradle/dependencies.gradle
+++ b/gradle/dependencies.gradle
@@ -4,7 +4,7 @@ repositories {
 
 ext {
   testngVersion = '7.1.0'
-  seleniumVersion = '3.141.59'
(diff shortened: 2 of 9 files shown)


src/main/java/com/codeborne/selenide/CollectionCondition.java
```diff
@@ -11,9 +11,9 @@
 import com.codeborne.selenide.collections.TextsInAnyOrder;
 import com.codeborne.selenide.impl.WebElementsCollection;
 import com.google.common.base.Predicate;
-import org.checkerframework.checker.nullness.compatqual.NullableDecl;
 import org.openqa.selenium.WebElement;
 
+import javax.annotation.Nullable;
```
rosolko: @asolntsev Maybe it's will be better to import `org.checkerframework.checker.nullness.qual.Nullable;` to match implemented method signature from guava library?

src/main/java/com/codeborne/selenide/impl/FilteringCollection.java
```diff
@@ -3,9 +3,9 @@
 import com.codeborne.selenide.Condition;
 import com.codeborne.selenide.Driver;
 import com.google.common.base.Predicate;
-import org.checkerframework.checker.nullness.compatqual.NullableDecl;
 import org.openqa.selenium.WebElement;
 
+import javax.annotation.Nullable;
```
rosolko: Same here

martinschneider:
What's the status of this? Any help required? I would like to use Selenide with Selenium 4.0.0-alpha3. Let me know if there's a way to contribute to the upgrade efforts.

asolntsev:
@martinschneider We are not working on this pull request. Since Selenium 4.0 is not released yet, I don't see any reason why we should upgrade to it. If I understood correctly, it will not be released soon. 

By the way, why do you want to use Selenium 4.0.0-alpha3?

martinschneider:
@asolntsev Fair enough if the plan is to wait for an official release. My point is, that once Selenium 4 is out, people will start upgrading and expect other libraries (like Selenide) to "just work".

While I agree that it doesn't make sense, to _release_ a version of Selenide which uses an alpha version of Selenium, I think it would make a lot of sense to keep this PR active and track the implications and amount of work supporting Selenium 4 will bring.

After all, that's what alpha releases are there for, to give tool developers like us a heads-up of what's coming.

asolntsev:
@martinschneider Thank you for the remainder!
I am absolutely agree, alpha releases are there for us to try them as early as possible.

I have merged master to this PR, fixed failing tests. Will keep it up-to-date until Selenium 4 is released.

rosolko:
@martinschneider Hey, but the way, if you want to use version from this branch you can easily build it and get for using via jitpack:
https://jitpack.io/#selenide/selenide/upgrade-to-selenium4-SNAPSHOT

asolntsev:
Overwritten by https://github.com/selenide/selenide/pull/1099

PR #941, opened 2019-05-24
Add support for Chrome option "useAutomationExtension"

harshtuna:
https://bugs.chromium.org/p/chromedriver/issues/detail?id=1749#c5

## Proposed changes
Configure useAutomationExtension using system property - same as args and prefs

## Checklist
> Caused by: com.puppycrawl.tools.checkstyle.api.CheckstyleException: Unable to find: config/checkstyle/suppressions.xml

### DIFF
--- a/src/main/java/com/codeborne/selenide/webdriver/ChromeDriverFactory.java
+++ b/src/main/java/com/codeborne/selenide/webdriver/ChromeDriverFactory.java
@@ -18,6 +18,7 @@ import java.util.stream.Stream;
 
 class ChromeDriverFactory extends AbstractDriverFactory {
   private static final Logger log = Logger.getLogger(ChromeDriverFactory.class.getName());
+  static final String USE_AUTOMATION_EXTENSION = "chromeoptions.useautomationextension";
 
   @Override
   WebDriver create(Config config, Proxy proxy) {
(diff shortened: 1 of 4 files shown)


BorisOsipov:
Hi @harshtuna 

If we follow such approach(add sys env for each experimentalOption name) at the end we will have thousands of them - it is bad.

harshtuna:
> Hi @harshtuna
> 
> If we follow such approach(add sys env for each experimentalOption name) at the end we will have thousands of them - it is bad.

What are the options then? Maybe allow injecting partly configured webdriver instance and merge selenide configuration on top? Not using selenide factory for complex setup and use selenide api directly is another choice.

BorisOsipov:
>Maybe allow injecting partly configured webdriver instance and merge selenide configuration on top?

It isn't possible. Once driver was created there is no ability to change e.g. capabilities.

Did you consider using `WebDriverProvider` in your project? See https://github.com/selenide/selenide/wiki/How-Selenide-creates-WebDriver

harshtuna:
> > Maybe allow injecting partly configured webdriver instance and merge selenide configuration on top?
> 
> It isn't possible. Once driver was created there is no ability to change e.g. capabilities.
> 
> Did you consider using `WebDriverProvider` in your project? See https://github.com/selenide/selenide/wiki/How-Selenide-creates-WebDriver

Yes, that was our plan B. Rejecting the PR then.

Issue #454, opened 2017-01-15
$$.shouldHave(texts()) should check size of collection first


asolntsev:
If table has 3 columns, then the following check could fails with "size does not match" error:

```java
$$("//table/thead/tr/th").shouldHave(texts("First Name, Last name, Username"));

// it's a bug in test. The correct code should be:
// $$("//table/thead/tr/th").shouldHave(texts("First Name", "Last name", "Username"));
```

Now it report an error that is quite misleading:

```
TextsMismatch
Actual: [First Name, Last name, Username]
Expected: [First Name, Last name, Username]
```

asolntsev:
Will be release in Selenide 5.2.4.

PR #944, opened 2019-05-31
First check size of collection on `shouldHave(texts())` assert 

xaknem:
## Proposed changes
Closes https://github.com/selenide/selenide/issues/454
I added a new exception TextsSizeMismatch to divide meaning with TextsMismatch.
New if statement to fail method with check for collections size. 
I has idea with changing returning value from boolean to Boolean(reference type to return null), but current realization is the easier and needs fewer changes.
And I write and rewrite current tests for these exceptions, because now we can check size before checking texts.  

It my first contribution and I would hear and correct all your comments with pleasure.

## Checklist
*except proxy and download tests, because they dont work on MacOS at latest version of master branch* and I don't affect this functions.

### DIFF
--- a/src/main/java/com/codeborne/selenide/collections/ExactTexts.java
+++ b/src/main/java/com/codeborne/selenide/collections/ExactTexts.java
@@ -4,6 +4,7 @@ import com.codeborne.selenide.CollectionCondition;
 import com.codeborne.selenide.ElementsCollection;
 import com.codeborne.selenide.ex.ElementNotFound;
 import com.codeborne.selenide.ex.TextsMismatch;
+import com.codeborne.selenide.ex.TextsSizeMismatch;
 import com.codeborne.selenide.impl.Html;
 import com.codeborne.selenide.impl.WebElementsCollection;
 import org.openqa.selenium.WebElement;
@@ -49,6 +50,9 @@ public class ExactTexts extends CollectionCondition {
       ElementNotFound elementNotFound = new ElementNotFound(collection, expectedTexts, lastError);
       elementNotFound.timeoutMs = timeoutMs;
       throw elementNotFound;
+    }
+    if (elements.size() != collection.getElements().size()) {
+      throw new TextsSizeMismatch(collection, ElementsCollection.texts(elements), expectedTexts, explanation, timeoutMs);
     } else {
       throw new TextsMismatch(collection, ElementsCollection.texts(elements), expectedTexts, explanation, timeoutMs);
     }
--- a/src/main/java/com/codeborne/selenide/collections/ExactTexts.java
+++ b/src/main/java/com/codeborne/selenide/collections/ExactTexts.java
@@ -51,7 +51,7 @@ public class ExactTexts extends CollectionCondition {
       elementNotFound.timeoutMs = timeoutMs;
       throw elementNotFound;
     }
-    if (elements.size() != expectedTexts.size()) {
+    else if (elements.size() != expectedTexts.size()) {
       throw new TextsSizeMismatch(collection, ElementsCollection.texts(elements), expectedTexts, explanation, timeoutMs);
     }
     else {
--- a/src/main/java/com/codeborne/selenide/collections/ExactTexts.java
+++ b/src/main/java/com/codeborne/selenide/collections/ExactTexts.java
@@ -51,9 +51,10 @@ public class ExactTexts extends CollectionCondition {
       elementNotFound.timeoutMs = timeoutMs;
       throw elementNotFound;
     }
-    if (elements.size() != collection.getElements().size()) {
+    if (elements.size() != expectedTexts.size()) {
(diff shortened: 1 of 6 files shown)


src/main/java/com/codeborne/selenide/ex/TextsSizeMismatch.java
```diff
@@ -0,0 +1,22 @@
...
+
+import com.codeborne.selenide.impl.WebElementsCollection;
+
+import java.util.List;
+
+public class TextsSizeMismatch extends UIAssertionError {
+  public TextsSizeMismatch(WebElementsCollection collection, List<String> actualTexts,
+                       List<String> expectedTexts, String explanation, long timeoutMs) {
+    super(collection.driver(),
+        "\nActual: " + actualTexts + ", List size: " + actualTexts.size() +
```
rosolko: Can you please use platform specific [System.lineSeparator()](https://docs.oracle.com/javase/8/docs/api/java/lang/System.html#lineSeparator--) insead of `\n` in exception class.

src/main/java/com/codeborne/selenide/collections/ExactTexts.java
```diff
@@ -49,7 +50,11 @@ public void fail(WebElementsCollection collection, List<WebElement> elements, Ex
       ElementNotFound elementNotFound = new ElementNotFound(collection, expectedTexts, lastError);
       elementNotFound.timeoutMs = timeoutMs;
       throw elementNotFound;
-    } else {
+    }
+    if (elements.size() != expectedTexts.size()) {
```
rosolko: Better to use `if { } else if { } else { }` construction in this case. Also please leave condition verification in one line as it was before

src/test/java/com/codeborne/selenide/collections/ExactTextsTest.java
```diff
@@ -98,6 +99,28 @@ void failOnTextMismatch() {
     }
   }
 
+  @Test
```
rosolko: Also since we are using AssertJ you can use [assertThatThrownBy()](https://joel-costigliola.github.io/assertj/assertj-core-features-highlight.html#exception-assertion) assertions for exceptions. You can find example in our existing tests.

src/main/java/com/codeborne/selenide/ex/TextsSizeMismatch.java
xaknem: fixed
``` super(collection.driver(),
        lineSeparator() + "Actual: " + actualTexts + ", List size: " + actualTexts.size() +
        lineSeparator() + "Expected: " + expectedTexts + ", List size: " + expectedTexts.size() +
        (explanation == null ? "" : lineSeparator() + "Because: " + explanation) +
        lineSeparator() + "Collection: " + collection.description());
```

src/main/java/com/codeborne/selenide/collections/ExactTexts.java
xaknem: fixed

src/test/java/com/codeborne/selenide/collections/ExactTextsTest.java
xaknem: fixed 
```assertThatThrownBy(() -> exactTexts.fail(mockedElementsCollection,
      singletonList(mockedWebElement), exception, 10000))
      .isInstanceOf(TextsSizeMismatch.class)
      .hasMessageContaining("Actual: [One, Two], List size: 1")
      .hasMessageContaining("Expected: [One, Two], List size: 2")
      .hasMessageEndingWith("Collection: Collection description");
```

src/test/java/com/codeborne/selenide/collections/ExactTextsTest.java
```diff
@@ -110,15 +110,12 @@ void failOnTextSizeMismatch() {
...
-        exception,
-        10000);
-    } catch (TextsSizeMismatch ex) {
-      assertThat(ex)
-        .hasMessage("\nActual: [One, Two], List size: 1\nExpected: [One, Two], List size: 2\nCollection: Collection description");
-    }
+    assertThatThrownBy(() -> exactTexts.fail(mockedElementsCollection,
+      singletonList(mockedWebElement), exception, 10000))
+      .isInstanceOf(TextsSizeMismatch.class)
+      .hasMessageContaining("Actual: [One, Two], List size: 1")
```
rosolko: Maybe better to use `hasMessageStartingWith` in this case?

xaknem: There is System.getLineSeparator() in the beginning. 
Do you think that really meaningfull here? It can worsen readability.

rosolko: oh. correct! let's go with current assertion. thanks

PR #948, opened 2019-06-10
Add support for hidden strings - passwords, sensitive data

firyx:
## Proposed changes
Sometimes you do not want to log/report passwords/sensitive data to logs/reports when using SelenideElement setValue.

I have created a simple HiddenString class to store plainTextValue (e.g. password) and user defined toStringValue. User defined toStringValue can be set e.g. to:

- "********"
- "Description of password for user A"
- "HashiCorp Vault path: 'path-for-password-for-user-a', Vault crypted value: 'sg363hey343h366g'"

The HashiCorp Vault version allows:

- to have all passwords be stored in HashiCorp Vault
- all test logs and reports be public without exposing or leaking passwords or sensitive data
- privileged users be able to decrypt values from logs and reports from HashiCorp Vault

You can use setValue("some string text") as before (backward compatibility)
or:

```
HiddenString hiddenPassword = new HiddenString(PasswordManager.getPasswordForUserA(), "This is description of my password visible in logs and reports");

$("#password").setValue(hiddenPassword);
```

Or with Vault in a way (simplified):

```
String vaultPath = "vault-store/prod-users/team-a/user-123";
String plaintextPassword = VaultUtils.getPasswordForUser(vaultPath);
String cryptedPassword = VaultUtils.encryptPassword(vaultPath, plaintextPassword);
String toStringValue = "Path: '" + vaultPath + "'";
toStringValue +=  "; CryptedPassword: ''" + cryptedPassword + "";

HiddenString hiddenPassword = new HiddenString(plaintextPassword, toStringValue);

$("#password").setValue(hiddenPassword);
```

### DIFF
--- a/src/main/java/com/codeborne/selenide/HiddenString.java
+++ b/src/main/java/com/codeborne/selenide/HiddenString.java
@@ -1,7 +1,6 @@
 package com.codeborne.selenide;
 
 public class HiddenString {
-
   private final String plainTextValue;
   private final String toStringValue;
 
@@ -18,5 +17,4 @@ public class HiddenString {
   public String toString() {
     return toStringValue;
   }
-
 }
--- a/src/main/java/com/codeborne/selenide/HiddenString.java
+++ b/src/main/java/com/codeborne/selenide/HiddenString.java
@@ -1,0 +1,22 @@
+package com.codeborne.selenide;
+
+public class HiddenString {
+
+  private final String plainTextValue;
+  private final String toStringValue;
+
+  public HiddenString(String plainTextValue, String toStringValue) {
+    this.plainTextValue = plainTextValue;
+    this.toStringValue = toStringValue;
+  }
+
+  public String getPlainTextValue() {
+    return plainTextValue;
+  }
+
+  @Override
+  public String toString() {
+    return toStringValue;
+  }
+
+}
--- a/src/main/java/com/codeborne/selenide/SelenideElement.java
+++ b/src/main/java/com/codeborne/selenide/SelenideElement.java
@@ -40,6 +40,29 @@ public interface SelenideElement extends WebElement, WrapsDriver, WrapsElement, 
   SelenideElement setValue(String text);
 
   /**
+   *
+   * <b>Implementation details:</b>
+   *
+   * <p>If Configuration.versatileSetValue is true, can work as 'selectOptionByValue', 'selectRadio'</p>
+   *
+   * <p>If Configuration.fastSetValue is true, sets value by javascript instead of using Selenium built-in "sendKey" function
+   * and trigger "focus", "keydown", "keypress", "input", "keyup", "change" events.
+   *
+   * <p>In other case behavior will be:
+   * <pre>
+   * 1. WebElement.clear()
+   * 2. WebElement.sendKeys(text)
+   * 3. Trigger change event
+   * </pre>
+   *
+   * @param hiddenString Useful for entering passwords or sensitive data. HiddenString plainTextValue to enter into the text field or set by value for select/radio,
+   * HiddenString toStringValue to log and report.
+   *
+   * @see com.codeborne.selenide.commands.SetValue
+   */
+  SelenideElement setValue(HiddenString hiddenString);
+
+  /**
    * Same as #setValue(java.lang.String)
    * @see com.codeborne.selenide.commands.Val
    */
--- a/src/main/java/com/codeborne/selenide/SelenideElement.java
+++ b/src/main/java/com/codeborne/selenide/SelenideElement.java
@@ -40,23 +40,9 @@ public interface SelenideElement extends WebElement, WrapsDriver, WrapsElement, 
   SelenideElement setValue(String text);
 
   /**
-   *
-   * <b>Implementation details:</b>
-   *
-   * <p>If Configuration.versatileSetValue is true, can work as 'selectOptionByValue', 'selectRadio'</p>
-   *
-   * <p>If Configuration.fastSetValue is true, sets value by javascript instead of using Selenium built-in "sendKey" function
-   * and trigger "focus", "keydown", "keypress", "input", "keyup", "change" events.
-   *
-   * <p>In other case behavior will be:
-   * <pre>
-   * 1. WebElement.clear()
-   * 2. WebElement.sendKeys(text)
-   * 3. Trigger change event
-   * </pre>
-   *
-   * @param hiddenString Useful for entering passwords or sensitive data. HiddenString plainTextValue to enter into the text field or set by value for select/radio,
-   * HiddenString toStringValue to log and report.
+   * Useful for entering passwords or sensitive data.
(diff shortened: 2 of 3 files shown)


src/main/java/com/codeborne/selenide/HiddenString.java
```diff
@@ -0,0 +1,22 @@
+package com.codeborne.selenide;
+
+public class HiddenString {
+
```
rosolko: Remove blank line

src/main/java/com/codeborne/selenide/HiddenString.java
```diff
@@ -0,0 +1,22 @@
...
+  public String getPlainTextValue() {
+    return plainTextValue;
+  }
+
+  @Override
+  public String toString() {
+    return toStringValue;
+  }
+
```
rosolko: ditto

src/main/java/com/codeborne/selenide/SelenideElement.java
```diff
@@ -39,6 +39,29 @@
    */
   SelenideElement setValue(String text);
 
+  /**
```
rosolko: This doc is not assignable to this method.
Use inherit doc or even better just leave block with `@see com.codeborne.selenide.commands.SetValue`
And leave only related method documentstion

src/main/java/com/codeborne/selenide/SelenideElement.java
```diff
@@ -39,6 +39,29 @@
...
+   * and trigger "focus", "keydown", "keypress", "input", "keyup", "change" events.
+   *
+   * <p>In other case behavior will be:
+   * <pre>
+   * 1. WebElement.clear()
+   * 2. WebElement.sendKeys(text)
+   * 3. Trigger change event
+   * </pre>
+   *
+   * @param hiddenString Useful for entering passwords or sensitive data. HiddenString plainTextValue to enter into the text field or set by value for select/radio,
```
rosolko: No use cases in documentation for this feature

src/main/java/com/codeborne/selenide/commands/SetValue.java
```diff
@@ -26,6 +27,12 @@ public SetValue(SelectOptionByValue selectOptionByValue, SelectRadio selectRadio
   @Override
   public WebElement execute(SelenideElement proxy, WebElementSource locator, Object[] args) {
     String text = (String) args[0];
+
+    if (args[0] instanceof HiddenString) {
```
rosolko: Modifying default setValue method you will push user to pass HiddenString instead just string

src/main/java/com/codeborne/selenide/SelenideElement.java
```diff
@@ -39,6 +39,29 @@
...
+   * HiddenString toStringValue to log and report.
+   *
+   * @see com.codeborne.selenide.commands.SetValue
+   */
+  SelenideElement setValue(HiddenString hiddenString);
```
rosolko: By the way in test you will pass string in open format such as `new HiddenString("asdf", "***")` and looks like there are no reason to do it in such way.
Again - add tests and proper documentation

firyx: You can get passwords from password manager (e.g. HashiCorp Vault) and not store them in git.

This is our use case and that is why we want to remove plaintext passwords also from logs/reports,
so that anybody can see logs/reports (e.g. Allure report) without leaking sensitive information.

src/main/java/com/codeborne/selenide/commands/SetValue.java
firyx: SetValue.execute method accepts Object[] args.

Where the first argument can be String (backwards compatibility) or HiddenString (added feature).

In SelenideElement there are existing setValue and new setValue methods:
SelenideElement setValue(String text);
SelenideElement setValue(HiddenString hiddenString);
Both of these methods are supported by SetValue.execute(...).

src/main/java/com/codeborne/selenide/HiddenString.java
```diff
@@ -0,0 +1,22 @@
...
+    this.toStringValue = toStringValue;
+  }
+
+  public String getPlainTextValue() {
+    return plainTextValue;
+  }
+
+  @Override
+  public String toString() {
+    return toStringValue;
```
rosolko: Since you're promote to mask password with asterisks why do we need to pass plain and tostring value. Just pass orignal value and in toString method just replace chars with asterisks. 

Something like that:
```
public class SensitiveValue {
    private final String plainValue;

    SensitiveData(final String plainValue) {
        this.plainValue = plainValue;
    }

    @Override
    public String toString() {
        return StringUtils.repeat("*", plainValue.length());
    }
}
```

BorisOsipov: @rosolko make sense. But it will be nice to have an ability to change asterisks with custom placeholder

firyx: We have use case,
where we will store in toString value path in HashiCorp Vault and crypted value from HashiCorp Vault.

This way we can make reports public and users,
that need to debug, decrypt the hidden values from the report know the Vault path and crypted value from the logs or reports and they can decrypt the value if they have permissions to this path in Vault.

So we need to keep it up to the user, what toString value he wants to set.

src/main/java/com/codeborne/selenide/SelenideElement.java
firyx: I hope I did it correctly.

src/main/java/com/codeborne/selenide/HiddenString.java
firyx: I have updated initial PR comment > Proposed changes with use case and example use.

src/main/java/com/codeborne/selenide/SelenideElement.java
firyx: Please see Vault use case and example in Initial comment > Proposed changes.

firyx:
What can I do with failing Firefox checks?
https://travis-ci.org/selenide/selenide/builds/543708087

asolntsev:
@firyx Technically, your solution is good.

But I am in doubt because:
1. Why do you use real passwords in tests? I think it's not a good practice. You should use a dedicated environment with test data for running tests. Isn't it a world-wide best practice?
2. It doesn't solve the problem anyway because passwords are still visible e.g. in chromedriver.log
For example, command `$(By.name("password")).val("sherlyn");` produces the following lines in "chromedriver.log":

```
[1563916584.390][INFO]: [5e5bb4972119c35fe8247bafaa10eb24] RESPONSE ClearElement
[1563916584.396][INFO]: [5e5bb4972119c35fe8247bafaa10eb24] COMMAND TypeElement {
   "id": "e39c1d05-f4f5-493a-ba16-1a34053da12a",
   "text": "sherlyn",
   "value": [ "s", "h", "e", "r", "l", "y", "n" ]
}
```

asolntsev:
Closed due to inactivity.

PR #962, opened 2019-08-05
Issue 469 improve "or" error message

asolntsev:
## Proposed changes

## Checklist

### DIFF
--- a/src/main/java/com/codeborne/selenide/conditions/Or.java
+++ b/src/main/java/com/codeborne/selenide/conditions/Or.java
@@ -6,10 +6,11 @@ import org.openqa.selenium.WebElement;
 
 import java.util.List;
 
+import static java.util.stream.Collectors.joining;
+
 public class Or extends Condition {
 
   private final List<Condition> conditions;
-  private Condition firstFailedCondition;
 
   public Or(String name, List<Condition> conditions) {
     super(name);
@@ -18,14 +19,9 @@ public class Or extends Condition {
 
   @Override
   public boolean apply(Driver driver, WebElement element) {
-    firstFailedCondition = null;
-
     for (Condition c : conditions) {
       if (c.apply(driver, element)) {
         return true;
-      }
-      else if (firstFailedCondition == null) {
-        firstFailedCondition = c;
       }
     }
     return false;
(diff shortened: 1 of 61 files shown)


src/main/java/com/codeborne/selenide/conditions/Or.java
```diff
@@ -18,26 +19,21 @@ public Or(String name, List<Condition> conditions) {
...
   @Override
   public String actualValue(Driver driver, WebElement element) {
-    return firstFailedCondition == null ? null : firstFailedCondition.actualValue(driver, element);
+    return conditions.stream().map(condition -> condition.actualValue(driver, element)).collect(joining(", "));
   }
 
   @Override
   public String toString() {
-    return firstFailedCondition == null ? super.toString() : firstFailedCondition.toString();
+    return getName() + ": " + conditions.stream().map(Condition::toString).collect(joining(" or "));
```
rosolko: It's better to introduce a new variable to complex stream operation and use string format for a final message

src/main/resources/versions.properties
```diff
@@ -1,14 +1,17 @@
+# This file is copied from https://raw.githubusercontent.com/bonigarcia/webdrivermanager/master/src/main/resources/versions.properties
```
rosolko: Maybe it's better to fetch this file from github url to use latest updates from source repository? And if it isn't available use local file instead?

src/test/java/com/codeborne/selenide/conditions/AttributeWithValueTest.java
```diff
@@ -0,0 +1,36 @@
...
+  }
+
+  @Test
+  void actualValue() {
+    assertThat(new AttributeWithValue("data-id", "expected").actualValue(driver, element)).isEqualTo("data-id=\"actual\"");
+  }
+
+  @Test
+  void tostring() {
+    assertThat(new AttributeWithValue("data-id", "expected").toString()).isEqualTo("attribute data-id=\"expected\"");
```
rosolko: Use `hasToString` assert

src/main/java/com/codeborne/selenide/conditions/Or.java
asolntsev: done

Issue #963, opened 2019-08-05
Add check on Xpath's start to SelenideElement find methods 


rusakovStas:
## The problem
When we use method find (ByXpath) or $/$$ on SelenideElement we expects that this method would be search element from this SelenideElement, but commonly it's not  true.  
For example: 

We have DOM: 
```
<A>
 <B>
   <C>
       <button>H!</button>
       <D> 
             <button>H!</button>
       </D>
    </C>
 </B>
 <B1> 
    <C>
             <button>H!</button>
    </C>
 </B1>
</A>

```
`$x("//B").findAll(By.xpath("//button"))  //should find "//B//button", but found just "//button"  `

To get "//B//button" we should specify Xpath like that 
`$x("//B").findAll(By.xpath(".//button")) `

To avoid that nonintuitive behavior i suggest to add check for Xpath's start to all find methods which working with Xpath to SelenideElement. That check will be throw exception with message like "You must start Xpath with '.' (dot symbol) when you want to find element from another." in case when search method on SelenideElement called with Xpath starts on no '.' (dot) symbols. 
Maybe, it make sense to add Configuration flag to turn off that behavior for backward compatibility and _people who wants strange_

rusakovStas:
I think the example is not very successful... 
Should i specify all examples (on all methods) or it better to do in real tests?

AlexKvasko:
You  should start your xpath from dot (.) if you want to find element in element
In Your example you should write $x("//B").findAll(By.xpath(".//button"))

rusakovStas:
Agree. Thanks)

asolntsev:
@rusakovStas I agree, we should add such a check. I don't think we need a `Configuration` flag because this behaviour is just never needed.

SeleniumTestAB:
I suggest different approach, maybe the '.' should be injected at start of string if someone uses xpath? From what i see it needs to be added in find(), findAll() and maybe $ methods.

asolntsev:
@SeleniumTestAB Probably it's a good idea. Hard to choose one: if Selenide should throw an exception or just add "." automatically. 

Probably exception is better because it forces people to learn XPath?

AlexKvasko:
> Probably exception is better because it would force people to learn XPath?
Please do not do anything with this functionality. This is a part ofXPath functionality not Selenium or Selenide.

rusakovStas:
@AlexKvasko 
Could you write an example of usage that functionality?

AlexKvasko:
[...]

PR #975, opened 2019-09-21
#963 fail fast if xpath for searching inside a web element starts with /

asolntsev:
## Proposed changes

## Checklist


rosolko:
@asolntsev we need to specify somethere how to chain multiple xpath search queries one by one.
I mean `$x("//a").$x("div")` example

asolntsev:
@rosolko What do you mean? 
In this case, it should be `$x("//a").$x("./div")`? Or even better: `$x("//a/div")`.

rosolko:
@asolntsev I saw few examples where parent element was described in separate field an reused for others :|
```
parent = $("//a")
child1 = parent.$("./div")
child2 = parent.$("./span")
```
I think we need to notify some how users about rules.
I predict that later we receive some bugs or questions about this changes :)
Maybe it's better to extend error message with hint with possible `./` solution?

Issue #1000, opened 2019-10-21
consider adding SelenideElement execute(Command<Void> command);


yashaka:
In 5.3.0 we added custom commands execution. 
the `execute` method accept only `Command<SelenideElement>`...
Yet, even native Selenide Click command is of type: `Command<Void>`... Hence, it would be natural and consistent to allow users to define their "void" commands and pass to the `execute` method...

asolntsev:
I was also thinking about it.
But it seemed to be useless, because `$.execute()` is mostly useful when chained with other commands. If you have just one `void` method, you can just call that method directly.

yashaka:
We also have native commands in Selenide that are of `Command<WebElement>` type...
So probably... we would need to be able to pass such commands to the execute method...

Maybe even it would be better originally allowing any Command<T> to be accepted by the execute command... And making it return "this" selenideelement in any case...

yashaka:
@asolntsev 
you can accept any Command<T> type, but return "this" selenideelement always... yet keeping it "chainable"

yashaka:
By the way... probably execute is uesfull not just for chaining...
this ability in Selenide is useful because you have an ability to define your custom command and not just call it separately but call it under the hood of same "waiting" logic that is available in native SelenideElement implementation

yashaka:
> BTW, probably it would be a good idea to convert method $.click() to non-void.

often this will be a workaround... 
For example... my custom command is supposed to return same "this" SelenideElement...
But I can't implement such command (correct me If I am wrong...), because inside Command<SelenideElement> implementation I have no access to "this" SelenideElement... 
So I will have to return "null" just to match the signature... 
I can also create new SelenideElement based on the webelement that I have... But then this will be a cached version of SelenideElement, that I also will not want in all scenarios...

asolntsev:
Implemented in https://github.com/selenide/selenide/pull/1001

PR #1001, opened 2019-10-21
#1000 make `$.execute(command)` generic: it now can return any value,…

asolntsev:
… or even be Void.

## Proposed changes

## Checklist


rosolko:
@asolntsev Make sense. Let's merge it!

sgulyaev:
So now execute will sometimes break chaincalls. Is it a bug or feature?

asolntsev:
@sgulyaev good point. Yes, now it can break, and this is a feature. Chaining is not always needed. You can still use only "chainable" commands if you wish.

PR #1004, opened 2019-10-21
Small refactor to ScreenshotLab and possible fix for junit5 screenshoter issue

SeleniumTestAB:
## Proposed changes
Small refactoring for ScreenshotLab class for readability and possible fix for Junit5 issue.

## Checklist


SeleniumTestAB:
The aforementioned fix is actually a bad one, i would suggest injecting dependency of Selenide from Allure and check if Listeners pool contains object of instance of class AllureSelenide.

PR #1006, opened 2019-10-23
Moving save page source mechanics to PageSourceExtractor class

SeleniumTestAB:
## Proposed changes
Moving some of the logic of saving page source to different class. This will align with good practice of making Single Responsibility for ScreenshotLab.

## Checklist


asolntsev:
@SeleniumTestAB It seems that this PR has common changes with https://github.com/selenide/selenide/pull/1004. Which one we choose?

PR #1009, opened 2019-10-27
#1006 refactor PageSourceExtractor: have public `extract` method with…

asolntsev:
…out boolean parameter

## Proposed changes

## Checklist

### DIFF
--- a/src/main/java/com/codeborne/selenide/impl/IO.java
+++ b/src/main/java/com/codeborne/selenide/impl/IO.java
@@ -1,0 +1,62 @@
+package com.codeborne.selenide.impl;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import javax.imageio.ImageIO;
+import java.awt.image.BufferedImage;
+import java.io.ByteArrayInputStream;
+import java.io.File;
+import java.io.FileInputStream;
+import java.io.FileOutputStream;
+import java.io.IOException;
+import java.io.InputStream;
+
+import static java.nio.charset.StandardCharsets.UTF_8;
+
+public class IO {
+  private static final Logger log = LoggerFactory.getLogger(IO.class);
+
+  public void copyFile(File sourceFile, File targetFile) throws IOException {
+    try (FileInputStream in = new FileInputStream(sourceFile)) {
+      writeToFile(in, targetFile);
+    }
+  }
+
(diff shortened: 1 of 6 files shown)


src/main/java/com/codeborne/selenide/impl/IO.java
```diff
@@ -0,0 +1,62 @@
...
+import java.io.ByteArrayInputStream;
+import java.io.File;
+import java.io.FileInputStream;
+import java.io.FileOutputStream;
+import java.io.IOException;
+import java.io.InputStream;
+
+import static java.nio.charset.StandardCharsets.UTF_8;
+
+public class IO {
```
rosolko: Not sure about `IO` class name, maybe better to call it for example `FileUtililities` because now it looks like simple utility class.
Also `impl` folder required some interface implementation.

src/main/java/com/codeborne/selenide/impl/IO.java
```diff
@@ -0,0 +1,62 @@
...
+      }
+    }
+  }
+
+  public void writeToFile(String content, File targetFile) {
+    try (ByteArrayInputStream in = new ByteArrayInputStream(content.getBytes(UTF_8))) {
+      writeToFile(in, targetFile);
+    }
+    catch (IOException e) {
+      log.error("Failed to write file {}", targetFile.getAbsolutePath(), e);
```
rosolko: Also, can you please mind me why don't we throw an exception in such case?

SeleniumTestAB:
@asolntsev i have a question. Since ```extract()``` method is not a static one you still need and object of PageSourceExtractor. So why you pass so much args to extract instead of passing them to object constructor?

src/main/java/com/codeborne/selenide/impl/IO.java
asolntsev: @rosolko This code is typically called in case of test failure. We want to save screenshot and page source as an additional information for the error message.

If we throw an exception, then we will loose the original cause of test failure.

asolntsev:
@SeleniumTestAB It's not that my solution is better, it's just a different approach. 
What I tried to do is like "dependency injection". We inject _bean_ `PageSourceExtractor` (which could be Singleton) and use it as a service. 

(still not sure that we really gain some benefit from this change :) )

asolntsev:
still not sure that we really gain some benefit from this change.
Let's cancel it.

Issue #1030, opened 2019-12-29
[RTC] Implement executeAsyncScript


tyge68:
## The problem
As a test implementer , I would like to test an sync method call on the browser using the webdriver .executeAsyncScript(...) method 

## Details
I couldn't find the equivalent of the .executeAsyncScript method (1) on the Selenide class. Only the non async one is available.

## Tell us about your environment
Not relevant

## Code To Reproduce Issue [ Good To Have ]
Current workaround:
JavascriptExecutor js = (JavascriptExecutor) WebDriverRunner.getWebDriver();
js.executeAsyncScript("someJavaScriptAsync(arguments[0])");

Currently I would like to have following:
Selenide.executeAsyncScript("someJavaScriptAsync(arguments[0])")

(1) https://selenium.dev/selenium/docs/api/java/org/openqa/selenium/JavascriptExecutor.html#executeAsyncScript-java.lang.String-java.lang.Object...-

asolntsev:
Done in PR #1031

PR #1031, opened 2019-12-29
#1030 [RTC] Implement executeAsyncScript

tyge68:
## Proposed changes
Adding executeAsyncJavascript(...) to the static Selenide object.
Adding new integration tests in the SelenideMethodsTest
Adding implementation in the DriverStub for any other tests to eventually use it as well

## Checklist

### DIFF
--- a/statics/src/main/java/com/codeborne/selenide/Selenide.java
+++ b/statics/src/main/java/com/codeborne/selenide/Selenide.java
@@ -583,16 +583,14 @@ public class Selenide {
   }
 
   /**
-   * @see <a href="https://selenium.dev/selenium/docs/api/java/org/openqa/selenium/JavascriptExecutor.html
-#executeScript-java.lang.String-java.lang.Object...-">executeScript</a>
+   * @see <a href="https://selenium.dev/selenium/docs/api/java/org/openqa/selenium/JavascriptExecutor.html#executeScript-java.lang.String-java.lang.Object...-">org.openqa.selenium.JavascriptExecutor.executeScript(String jsCode, Object... arguments)</a>
    */
   public static <T> T executeJavaScript(String jsCode, Object... arguments) {
     return getSelenideDriver().executeJavaScript(jsCode, arguments);
   }
 
   /**
-   * @see <a href="https://selenium.dev/selenium/docs/api/java/org/openqa/selenium/JavascriptExecutor.html
-#executeAsyncScript-java.lang.String-java.lang.Object...-">executeAsyncScript</a>
+   * @see <a href="https://selenium.dev/selenium/docs/api/java/org/openqa/selenium/JavascriptExecutor.html#executeAsyncScript-java.lang.String-java.lang.Object...-">org.openqa.selenium.JavascriptExecutor.executeAsyncJavaScript(String jsCode, Object... arguments)</a>
    */
   public static <T> T executeAsyncJavaScript(String jsCode, Object... arguments) {
     return getSelenideDriver().executeAsyncJavaScript(jsCode, arguments);
--- a/statics/src/main/java/com/codeborne/selenide/Selenide.java
+++ b/statics/src/main/java/com/codeborne/selenide/Selenide.java
@@ -583,14 +583,16 @@ public class Selenide {
   }
 
   /**
-   * @see <a href="https://bit.ly/37mfHsy">executeScript</a>
+   * @see <a href="https://selenium.dev/selenium/docs/api/java/org/openqa/selenium/JavascriptExecutor.html
+#executeScript-java.lang.String-java.lang.Object...-">executeScript</a>
    */
   public static <T> T executeJavaScript(String jsCode, Object... arguments) {
     return getSelenideDriver().executeJavaScript(jsCode, arguments);
   }
 
   /**
-   * @see <a href="https://bit.ly/2tW7QTU">executeAsyncScript</a>
+   * @see <a href="https://selenium.dev/selenium/docs/api/java/org/openqa/selenium/JavascriptExecutor.html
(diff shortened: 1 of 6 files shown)


statics/src/main/java/com/codeborne/selenide/Selenide.java
```diff
@@ -582,10 +582,20 @@ public static ElementsCollection getElements(By criteria) {
     return getSelenideDriver().findAll(criteria);
   }
 
+  /**
+   * @see @see <a href="https://selenium.dev/selenium/docs/api/java/org/openqa/selenium/JavascriptExecutor.html#executeScript-java.lang.String-java.lang.Object...-">executeScript</a>
```
rosolko:
```suggestion
   * @see <a href="https://selenium.dev/selenium/docs/api/java/org/openqa/selenium/JavascriptExecutor.html#executeScript-java.lang.String-java.lang.Object...-">executeScript</a>
```

Looks like checkstyle isn't pass such long line. We have a limit equals to `136` characters.

statics/src/main/java/com/codeborne/selenide/Selenide.java
```diff
@@ -582,10 +582,20 @@ public static ElementsCollection getElements(By criteria) {
...
+   */
   public static <T> T executeJavaScript(String jsCode, Object... arguments) {
     return getSelenideDriver().executeJavaScript(jsCode, arguments);
   }
 
+  /**
+   * @see @see <a href="https://selenium.dev/selenium/docs/api/java/org/openqa/selenium/JavascriptExecutor.html#executeAsyncScript-java.lang.String-java.lang.Object...-">executeAsyncScript</a>
```
rosolko:
```suggestion
   * @see <a href="https://selenium.dev/selenium/docs/api/java/org/openqa/selenium/JavascriptExecutor.html#executeAsyncScript-java.lang.String-java.lang.Object...-">executeAsyncScript</a>
```

Same here.

rosolko:
@tyge68 Please execute locally gradle  `check` task

tyge68:
@rosolko thanks for the info, I changed to use a bit.ly short url as it is already used in this class for getWebDriverLogs(...) which avoid then to have line breaks. I also re-run "gradle check" and it passed.

Let me know if that works for you now.

rosolko:
@tyge68 bitly links isn't a good idea, we think about them previously. 
We have several options how to avoid them:
1. Replace link with text from doc
2. Use full link and ignore it on check
3. Try to split link to multiple row if possible.
4. Extend line length in a rules
@asolntsev ^

statics/src/main/java/com/codeborne/selenide/Selenide.java
```diff
@@ -582,10 +582,22 @@ public static ElementsCollection getElements(By criteria) {
     return getSelenideDriver().findAll(criteria);
   }
 
+  /**
+   * @see <a href="https://selenium.dev/selenium/docs/api/java/org/openqa/selenium/JavascriptExecutor.html
```
asolntsev: A proper way to refer other Java classes in javadoc is mentioning class name + method name with argument types:

```java
/**
 * @see org.openqa.selenium.By#className(java.lang.String)
 */
```

This way we don't need to bother about long or changing URLs.

statics/src/test/java/integration/SelenideMethodsTest.java
```diff
@@ -564,6 +565,20 @@ void canExecuteCustomCommand() {
...
+  @Test
+  void canExecuteJavascript() {
+    long value = (Long) Selenide.executeJavaScript("return 10;");
+    assertThat(value).isEqualTo(10);
+  }
+
+  @Test
+  void canExecuteAsyncJavascript() {
+    long value = (Long) Selenide.executeAsyncJavaScript(
+      "var callback = arguments[0]; setTimeout(function() { callback(10); }, 50);"
```
asolntsev: The proper way is `arguments[arguments.length - 1]` instead of `arguments[0]`.

tyge68: done in my next commit

statics/src/main/java/com/codeborne/selenide/Selenide.java
tyge68: true,  but then you would need to find out the page in google, the idea is to provide a link still to help the user.

I suggest to add an ignore pattern in the checkstyle related rule in my next commit, that allow longer line for " * @see " which is then more url friendly. That also permit to include a longer text inside the link itself.

tyge68:
> For example, we could introduce a "reserved word" like CALLBACK or RETURN or EXIT

I suppose you would like that such reserved keywords are replaced before doing the call to JavascriptExecutor method ?  That would be indeed a possibility. But for sure a bit outside the initial scope of this RTC. But if that is only this , why not adding this along.

statics/src/main/java/com/codeborne/selenide/Selenide.java
asolntsev: @tyge68 No, you don't need to google. IDEA can automatically download sources jar from maven central repo. You can easily browse sources/javadoc of any open-sources projects right in IDEA (not sure about Eclipse/NetBeans).

asolntsev:
@tyge68 
1. Yes, I meant that reserved keywords are replaced before doing the call to JavascriptExecutor.
2. You are right, my suggestion with CALLBACK is outside of the initial scope. We can discuss it later because it doesn't influence the API. 

Rather I was thinking about creating a better API **before** we release this feature. Thought, it seems that none of us has a good idea how to improve API. So we can release it as is.

statics/src/main/java/com/codeborne/selenide/Selenide.java
rosolko: +

Issue #1034, opened 2020-01-03
Selenide cannot download file if response doesn't contain "Content-Disposition" header


asolntsev:
## The problem
In "proxy" mode, Selenide can download file if the server response contains "Content-Disposition" header. 
But valid responses may not have this header. 
Selenide should be able to download such files too. 

## Code To Reproduce Issue 
```java
Configuration.fileDownload = PROXY;
open("https://selenide.org/quick-start.html");
File selenideJar = $(By.linkText("selenide.jar")).download();
```

Getting error:

```java
java.io.FileNotFoundException: Failed to download file {By.linkText: selenide.jar} in 4000 ms.Intercepted 3 responses:
  #1  https://search.maven.org/remotecontent?filepath=com/codeborne/selenide/5.6.0/selenide-5.6.0.jar -> 301 "Moved Permanently" {Server=awselb/2.0, Connection=keep-alive, Content-Length=150, Date=Fri, 03 Jan 2020 20:32:21 GMT, Content-Type=text/html, Location=https://repo1.maven.org:443/fromsearch?filepath=com/codeborne/selenide/5.6.0/selenide-5.6.0.jar} text/html  (150 bytes)
  #2  https://repo1.maven.org/fromsearch?filepath=com/codeborne/selenide/5.6.0/selenide-5.6.0.jar -> 301 "Moved Permanently" {Accept-Ranges=bytes, X-Cache=HIT, HIT, X-Served-By=cache-iad2151-IAD, cache-bma1628-BMA, Connection=keep-alive, Vary=Fastly-SSL,, Content-Length=0, X-Cache-Hits=1, 2, Date=Fri, 03 Jan 2020 20:32:21 GMT, Age=701038, Location=https://repo1.maven.org/maven2/com/codeborne/selenide/5.6.0/selenide-5.6.0.jar, Via=1.1 varnish, X-Timer=S1578083541.349571,VS0,VE0} application/octet-stream  (0 bytes)
[...]

PR #1035, opened 2020-01-05
Improve proxy download

asolntsev:
This PR fixes issue #1034


gradle/dependencies.gradle
```diff
@@ -16,9 +16,10 @@ dependencies {
   api('io.github.bonigarcia:webdrivermanager:3.7.1')
 
   implementation('com.browserup:browserup-proxy-core:2.0.1')
-  implementation('io.netty:netty-codec:4.1.42.Final')
+  implementation('io.netty:netty-all:4.1.44.Final')
```
rosolko: Do we really need `netty-all` dependency?
Is `netty-all` contains `netty-codec` dependency?

gradle/dependencies.gradle
```diff
@@ -16,9 +16,10 @@ dependencies {
...
+  implementation('io.netty:netty-codec:4.1.44.Final')
```
rosolko: Is `netty-all` contains `netty-codec` dependency?

gradle/dependencies.gradle
asolntsev: BrowserMobProxy always depended on `io.netty:netty-all` (but just older version). 
But yes, it seems that we don't need `io.netty:netty-codec` because `netty-all` already contains everything. I have removed it.

gradle/dependencies.gradle
asolntsev: yes. Removed `netty-codec`.

PR #1049, opened 2020-02-04
Fix the Jenkins integration issue

topikachu:
## Proposed changes
Fix the Jenkins integration issue
1. Read BUILD_URL from both system property and environment variable
2. Calculate the report relative path when resolving report URL.

## Checklist
It's about Jenkins integration. No test.
Fix an issue. No document.

### DIFF
--- a/src/main/java/com/codeborne/selenide/impl/CiReportUrl.java
+++ b/src/main/java/com/codeborne/selenide/impl/CiReportUrl.java
@@ -40,16 +40,15 @@ public class CiReportUrl {
   }
 
   private String getJenkinsReportsUrl() {
-    String build_url = System.getProperty("BUILD_URL", System.getenv("BUILD_URL"));
-    String workspace = System.getProperty("WORKSPACE", System.getenv("WORKSPACE"));
-
+    String build_url = System.getProperty("BUILD_URL");
     if (!isEmpty(build_url)) {
+      String workspace = System.getProperty("WORKSPACE", System.getenv("WORKSPACE"));
       String reportRelativePath = "";
       if (!isEmpty(workspace)) { // we have a workspace folder. Calculate the report relative path
         Path pathAbsoluteReportsFolder = Paths.get("").normalize().toAbsolutePath();
         Path pathAbsoluteWorkSpace = Paths.get(workspace).normalize().toAbsolutePath();
         Path pathRelative = pathAbsoluteWorkSpace.relativize(pathAbsoluteReportsFolder);
-        reportRelativePath = pathRelative.toString() + '/';
+        reportRelativePath = pathRelative.toString().replace('\\', '/') + '/';
       }
       return resolveUrlSource("%s/artifact/%s", build_url, reportRelativePath);
     } else {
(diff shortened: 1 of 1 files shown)


src/main/java/com/codeborne/selenide/impl/CiReportUrl.java
```diff
@@ -39,8 +41,23 @@ private String getTeamCityUrl() {
 
   private String getJenkinsReportsUrl() {
     String build_url = System.getProperty("BUILD_URL");
+    if (isEmpty(build_url)) {
+      build_url = System.getenv("BUILD_URL");
+    }
+    String workspace = System.getProperty("WORKSPACE");
+    if (isEmpty(workspace)) {
+      workspace = System.getenv("WORKSPACE");
+    }
```
asolntsev: It can be simplified: `String workspace = System.getProperty("WORKSPACE", System.getenv("WORKSPACE"));`

src/main/java/com/codeborne/selenide/impl/CiReportUrl.java
```diff
@@ -39,8 +41,23 @@ private String getTeamCityUrl() {
 
   private String getJenkinsReportsUrl() {
     String build_url = System.getProperty("BUILD_URL");
+    if (isEmpty(build_url)) {
+      build_url = System.getenv("BUILD_URL");
+    }
```
asolntsev: It can be simplified: `String buildUrl = System.getProperty("BUILD_URL", System.getenv("BUILD_URL"));`

src/main/java/com/codeborne/selenide/impl/CiReportUrl.java
```diff
@@ -39,8 +41,23 @@ private String getTeamCityUrl() {
...
     if (!isEmpty(build_url)) {
-      return resolveUrlSource("%s/artifact/", build_url);
+      String reportRelativePath = "";
+      if (!isEmpty(workspace)) { // we have a workspace folder. Calculate the report relative path
+        Path pathAbsoluteReportsFolder = Paths.get("").normalize().toAbsolutePath();
+        Path pathAbsoluteWorkSpace = Paths.get(workspace).normalize().toAbsolutePath();
+        Path pathRelative = pathAbsoluteWorkSpace.relativize(pathAbsoluteReportsFolder);
+        reportRelativePath = pathRelative.toString() + '/';
+      }
+      return resolveUrlSource("%s/artifact/%s", build_url, reportRelativePath);
```
asolntsev: This file needs tests, but I can do it later.

topikachu: I'd like to know how to test such integration.

rosolko: I think the way how to test is to set system properties with custom path for tests before tests, initialize driver with config and check config value with expected path.

topikachu:
@asolntsev  Don't merge this change. I found two issues.

1. Jenkins integration is automatically enabled. I think it's better to enable the CI report until BUILD_URL is set explicitly as old behavior.
2. The code doesn't work on Windows slaves.

I will refine the PR.

src/main/java/com/codeborne/selenide/impl/CiReportUrl.java
topikachu: > I think the way how to test is to set system properties with custom path for tests before tests, initialize driver with config and check config value with expected path.

Not so easy. I use  `Paths.get("")` to calculate the relative path. :(

Issue #814, opened 2018-09-26
Screenshots no longer available


afamee:
## The problem
I no longer get the screenshots upon failure ever since I upgraded from Selenide 2.17 to Selenide 4.12.
For the longest time, I had a very old Selenide code for reasons that are isolated to my setup. Finally I found a way to work around the issue hindering my upgrade. So I took a big leap and upgraded from Selenide 2.17 to 4.12. I was really excited considering I have now new api's at my disposal. I must say my expectation has not been met.

- Practically every failure has failed without a screenshot that I am used to which makes it difficult to see what was on the screen upon failure
- I get frequently now **NotFoundElement** Failures. Upon update i noticed quite some deprecated functions which I updated to most suitable apis. Now It seems my tests is a bit unstable which I am currently investigation to find out the root cause. I use a lot of should(Condition.appear, 2500).click() type calls around my code to ensure the elements have enough time to be rendered before the click.

## Tell us about your environment
**Selenide 4.12.2**
**Firefox 61.0.1**
**Selenium-firefox-driver 3.13.0** 
**Selenium 3.13.0**
**Windows 10 Pro**
**Fitnesse-20180127**

## Code To Reproduce Issue [ Good To Have ]
Please remember that with sample code it's easier to reproduce the bug.

**To reproduce:**
$Open any page of a given site
$Click on a element that is not there

It should throw a ElementNotFoundException and suppose to show a screenshot path to where the page is stored.

plagov:
This is a very basic functionality that works perfectly. And without your code example, it is impossible to see what is the problem you're experiencing.

plagov:
Stacktrace is impossible to read, it is uploaded as a screenshot that doesn't appear in full size. Place it as a plain text here.

Also, `Condition.appear` is deprecated. Better to use `Condition.visible`. And what is the `policy_num`? Do you define this variable?

I tried to reproduce your issue with the following test:

```java
  @Test
  public void scrapePolicyNumber(){
    open("https://www.google.com/");
    $("#PolicySummary_PolicyNumber span").shouldBe(appear).getText();
  }
```
As expected - it failed, I get a `NoSuchElementException` and screenshot was created. You can try the same test on your side and see that the result will be the same.

afamee:
[...]

PR #1052, opened 2020-02-04
Fix Screenshots 

petroOv-PDFfiller:
## Proposed changes
Fix concurrent problem when using Selenide + TestNg with ScreenShooter listener(related to [814](https://github.com/selenide/selenide/issues/814) ). To Screenshots field screenshots - was assigned new instance of ScreenShotLaboratory, that produce bug in ScreenShooter listener, finishContext() always returns empty List, because currentContextScreenshots - always was null when was started by method Screenshots. startContext(String className, String methodName).
Also if ScreenShotLaboratory planned to be used as Singleton, default Contstructor must be private. To avoid similar bugs in future

## Checklist

### DIFF
--- a/statics/src/main/java/com/codeborne/selenide/Screenshots.java
+++ b/statics/src/main/java/com/codeborne/selenide/Screenshots.java
@@ -10,7 +10,7 @@ import java.util.List;
 import static com.codeborne.selenide.WebDriverRunner.driver;
 
 public class Screenshots {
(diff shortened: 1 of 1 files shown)


petroOv-PDFfiller:
@asolntsev I agree with you, be great if selenide had a service-based screenshot feature with default implementation as ScreenShotLaboratory. Maybe you can add this to selenide roadmap and I, from my side, w'll try to help you with implementation

Issue #1057, opened 2020-02-05
Selenide could download files to `Configuration.downloadsFolder` instead of ~/Downloads


asolntsev:
## The problem
When using `Configuration.downloadMode=PROXY`, Selenide lets browser download file to a default system location (something like ~/Downloads), and then makes a copy of downloaded file into `Configuration.downloadsFolder`.

Thus we have duplicated files. And nobody clears them up from ~/Downloads.

## Suggestion
Since Selenide opens a browser itself (by default), it could configure the browser to download files immediately to `Configuration.downloadsFolder`. 

### Browsers support
I am not sure that all webdrivers allow such a configuration, but at least Chrome does:

```java
chromePreferences.put("download.default_directory", downloadsFolder.getAbsolutePath());
```

P.S. We could also disable a download dialog (at least for Chrome):
```java
chromePreferences.put("profile.default_content_settings.popups", 0);
```

asolntsev:
done in PR #1058

PR #1058, opened 2020-02-07
download files to `Configuration.downloadsFolder` instead of ~/Downloads

asolntsev:
implementation of https://github.com/selenide/selenide/issues/1057


src/main/java/com/codeborne/selenide/webdriver/FirefoxDriverFactory.java
```diff
@@ -49,31 +56,60 @@ FirefoxOptions createFirefoxOptions(Config config, Proxy proxy) {
...
         else {
           profile.setPreference(capability, value);
         }
       }
     }
-    return currentFirefoxOptions.setProfile(profile);
+  }
+
+  private boolean isBoolean(String value) {
+    return "true".equals(value) || "false".equals(value);
```
dkorobtsov: Suppose makes sense to convert value to lower case before comparing.

rosolko: Or even better - just use `String.equalsIgnoreCase` method

Issue #662, opened 2018-01-04
Add Condition with Predicate<WebElement> parameter as alternative to check element conditions


rumpfc:
Since Java 8 supports Lambda notations, there should be a possibility to check custom conditions with `Predicate<WebElement>` instead of writing a new Condition class which overrides `Condition.accept(WebElement).`

Example: Check if input element (id=myinputfield) doesn't contain any text (method name can be different)
```
Selenide.$("#myinputfield")
   .should(Condition.customCondition(w -> w.getAttribute("value").isEmpty()));
```

If giving each Condition a name is mandatory, maybe calling it like this:
```
Selenide.$("#myinputfield")
   .should(Condition.customCondition("Input element is empty", w -> w.getAttribute("value").isEmpty()));
```

- [ ] Add static function which accepts a Predicate<Webelement> (and String name) as parameter(s)

asolntsev:
Implemented in https://github.com/selenide/selenide/pull/1059

PR #1059, opened 2020-02-08
Custom condition

dstekanov:
## Proposed changes
Added possibility to make checks with a custom predicate.

Example:

`$("#element").should(match("tag=input", el -> el.getTagName().equals("input"))`

If resolves a feature request #662 

## Checklist

### DIFF
--- a/src/main/java/com/codeborne/selenide/CollectionCondition.java
+++ b/src/main/java/com/codeborne/selenide/CollectionCondition.java
@@ -1,5 +1,6 @@
 package com.codeborne.selenide;
 
+import com.codeborne.selenide.collections.AllMatch;
 import com.codeborne.selenide.collections.AnyMatch;
 import com.codeborne.selenide.collections.ExactTexts;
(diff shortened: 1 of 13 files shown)


dstekanov:
How do you think, is it a good idea to create something similar for $$ ?

For example: `$$("...").shouldBe(anyMatch(predicate));` , `$$("...").shouldBe(allMatch(predicate));`

rosolko:
> How do you think, is it a good idea to create something similar for $$ ?
> 
> For example: `$$("...").shouldBe(anyMatch(predicate));` , `$$("...").shouldBe(allMatch(predicate));`

Yes, definitely.
It will be great if such feature will be implemented in scope of this merge request since it's similare feature.

dstekanov:
Nice, will add it soon.

SeleniumTestAB:
@dstekanov I have a question. For what type of situation you need such feature?

dstekanov:
I've done it to resolve existing feature request #662 so it seems someone needed it...

PR #1062, opened 2020-02-12
Fixing #1060

### LINKED ISSUE #1060: Quotes in byAttribute selector are not handled properly.
## The problem
Quotes in byAttribute selector are not handled properly.

## Details
Selector like this _byAttribute("data-testid", "Checkbox Don't show again")_ will throw an error.
Same as _byAttribute("data-testid", "Checkbox Don\\'t show again")_.
Selector above is transformed into _[data-testid='Checkbox Don't show again']_ which is invalid.
Issue can be bypassed using css selector _"[data-testid=\"Checkbox Don't show again\"]"_.

[byAttribute Selector in code](https://github.com/selenide/selenide/blob/master/src/main/java/com/codeborne/selenide/Selectors.java#L65)

* **Selenide**: 5.6.1

## Code To Reproduce Issue [ Good To Have ]
$(byAttribute("data-testid", "Checkbox Don't show again"))

denysLystopadskyy:
[#1060](https://github.com/selenide/selenide/issues/1060)

### DIFF
--- a/src/main/java/com/codeborne/selenide/Selectors.java
+++ b/src/main/java/com/codeborne/selenide/Selectors.java
@@ -8,7 +8,7 @@ public class Selectors {
 
   /**
    * Find element CONTAINING given text (as a substring).
-   *
+   * <p>
    * This method ignores difference between space, \n, \r, \t and &nbsp;
    * This method ignores multiple spaces.
    *
@@ -21,7 +21,7 @@ public class Selectors {
 
   /**
    * Find element that has given text (the whole text, not a substring).
-   *
+   * <p>
    * This method ignores difference between space, \n, \r, \t and &nbsp;
    * This method ignores multiple spaces.
    *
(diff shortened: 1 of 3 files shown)


src/test/java/com/codeborne/selenide/SelectorsTest.java
```diff
@@ -57,35 +57,35 @@ void byTextEscapesQuotes() {
...
     assertThat(selector)
-      .hasToString("By.cssSelector: [value='катя']");
+      .hasToString("By.cssSelector: [value=\"катя\"]");
   }
 
   @Test
   void byAttributeEscapesQuotes() {
     By selector = Selectors.byAttribute("value", "Ludvig'van\"Beethoven");
     assertThat(selector)
-      .hasToString("By.cssSelector: [value='Ludvig'van\"Beethoven']");
```
vinogradoff: Just a fun fact - this test case has produced the wrong locator all the way along, but test was passed :)

src/test/java/com/codeborne/selenide/SelectorsTest.java
```diff
@@ -57,35 +57,35 @@ void byTextEscapesQuotes() {
...
+      .hasToString("By.cssSelector: [value=\"Ludvig'van\"Beethoven\"]");
```
vinogradoff: unfortunately this locator is also wrong.

One valid was to test it - run driver.findElement(By.cssSelector(locator)); - it should not throw any exceptions.

asolntsev:
@denysLystopadskyy Why have you closed this PR? It's a good change. 

P.S. I would suggest using method `Quotes.escape(attributeValue)`:

```java
  public static By byAttribute(String attributeName, String attributeValue) {
    return By.cssSelector(String.format("[%s=%s]", attributeName, Quotes.escape(attributeValue)));
  }
```

src/test/java/com/codeborne/selenide/SelectorsTest.java
BorisOsipov: it is still broken selector

src/main/java/com/codeborne/selenide/Selectors.java
```diff
@@ -62,7 +62,7 @@ public static By byText(String elementText) {
    * @return standard selenium By cssSelector criteria
    */
   public static By byAttribute(String attributeName, String attributeValue) {
-    return By.cssSelector(String.format("[%s='%s']", attributeName, attributeValue));
+    return By.cssSelector(String.format("[%s=\"%s\"]", attributeName, attributeValue));
```
vinogradoff: it doesn' solve the problem yet. You must also replace existing  "-s  and '-s  with \" and \'

src/test/java/com/codeborne/selenide/SelectorsTest.java
vinogradoff: Actually, I even don't know the working locator for this case. It is simple, when only ' or only " is there, but I don't know the solution if text is  `Ab'cd"de `

denysLystopadskyy:
@asolntsev I was not sure how to hide it before update.

denysLystopadskyy:
> @denysLystopadskyy Why have you closed this PR? It's a good change.
> 
> P.S. I would suggest using method `Quotes.escape(attributeValue)`:
> 
> ```java
>   public static By byAttribute(String attributeName, String attributeValue) {
>     return By.cssSelector(String.format("[%s=%s]", attributeName, Quotes.escape(attributeValue)));
>   }
> ```

@asolntsev In case of ***"Ludvig'van"Beethoven"*** this method returns ***concat("Ludvig'van", '"', "Beethoven")*** which does not look like a good css selector.

denysLystopadskyy:
@asolntsev Maybe we can use StringEscapeUtils?
For example escapeJava returns:
[value="Ludvig'van\"Beethoven"] for "Ludvig'van\"Beethoven"
[value="\u043A\u0430\u0442\u044F"] for катя

asolntsev:
1. This PR has failing unit-tests, but I will fix it by myself. 
2. As I mentioned in https://github.com/selenide/selenide/pull/1062#pullrequestreview-366358988, some cases are still not supported. But those are rare cases, and the current PR makes things better than before. So I am going to merge it.

Issue #845, opened 2018-10-29
Add the possibility to move to a certain sibling and preceding element, after `parent()` method


plagov:
## The problem
Add the possibility to move to a certain sibling and preceding element, after `parent()` method, relative to the one the `parent()` method was called.

## Details
Consider the following HTML:

```html
<tr>
  <td>
  <td>
</tr>
<tr>
  <td class="today">
  <td>
</tr>
<tr>
  <td>
  <td>
</tr>
```

So, if I would need to move to the `td` element in `tr` that is next to `tr td.today`, I would do: 
`$("td.today").parent().sibling(0).find("td", 1);`

Respectively, if moving to the `td` in preceding `tr` before `tr td.today`, then it would be:
`$("td.today").parent().preceding(0).find("td", 1);`

Argument of `sibling()` or `preceding()` method is the index of a sibling/preceding element relative to the current element where `parent()` method was called.

asolntsev:
Implemented in https://github.com/selenide/selenide/pull/1064

PR #1064, opened 2020-02-13
Added sibling() and preceding() element commands

dstekanov:
## Proposed changes
Added the ability to move to a sibling or preceding element of the current element.

The argument of sibling() or preceding() methods is the index of a sibling/preceding element relative to the current element, but there are two questions/concerns:

I am not sure about the index. 'preceding-sibling' and 'following-sibling' XPaths do not accept 0-based indexes. So, I'm adding +1 inside GetSibling/GetPreceding. But in the report, you will see (in case if the element was not found) your index+1 that is quite disappointing... 

The question is, if should I leave it as it is or to document SelenideElement.sibling() and preceding() that they should accept indexes starting from 1 ...

Feature request #845 

## Checklist

### DIFF
--- a/src/main/java/com/codeborne/selenide/SelenideElement.java
+++ b/src/main/java/com/codeborne/selenide/SelenideElement.java
@@ -384,6 +384,18 @@ public interface SelenideElement extends WebElement, WrapsDriver, WrapsElement, 
   SelenideElement parent();
 
   /**
+   * Get sibling element of this element
+   * ATTENTION! This method doesn't start any search yet!
+   * For example, $("td").sibling(0) will give first sibling of "td"
+   *
+   * @param index the index of sibling element
+   * @return Sibling element by index
+   *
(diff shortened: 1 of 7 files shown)


SeleniumTestAB:
As typical user i would be happy that there is documentation for this one. Its not so obvious.

dstekanov:
It would be nice to solve the question with indices first...

SeleniumTestAB:
@dstekanov i was refering to that question. You could document or add a link to Xpath documentation that the index can not be 0 for those. Another way would be to override the raport message and lower index by 1 after you added your 1. Wouldnt that be a good option?

dstekanov:
@SeleniumTestAB thanks! Where do you want me to leave the comment about indexes? Add it for SelenideElement.sibling()/preceding() methods?

Also, I am not sure that I understand about using string instead of indexes? I will appreciate if you will provide some examples...

PR #1066, opened 2020-02-17
Add swipe commands

shalak:
## Proposed changes
Adding native commands to swipe the center of element to the edge of the window

## Checklist


rosolko:
Since it's just using actions under hood - looks like it can be just a custom command not a part of regular code base.
You can see example [here](https://github.com/selenide/selenide/blob/master/statics/src/test/java/integration/SelenideMethodsTest.java#L559)

BorisOsipov:
agree with @rosolko. Also I believe that such swipe won't work in all cases - so it is project specific

Issue #1069, opened 2020-02-26
Provide IsImage Comand functionality as a Condition


gregorko:
I would like to use the IsImage check (mainly for checking if the image is loaded successfully) as a condition. 
``
$("img").should(isImage)
``

Currently this is not possible and you have to do:
``
assertThat($("img").isImage()).isTrue()
``

Which feels awkward when you are doing all your checks in the Selenide style. Also the screenshot functionality seems to not work when using assertThat checks.

rosolko:
You can easily implement and use custom condition for it
https://github.com/selenide/selenide/wiki/Custom-conditions

asolntsev:
Impemented in PR #1086

PR #1086, opened 2020-03-11
add isImage condition

dstekanov:
## Proposed changes
Added "isImage" condition to simplify checking if an image is loaded. Usage example: `$("img").should(isImage);` 

Iе resolves a feature request #1069 

## Checklist

### DIFF
--- a/src/main/java/com/codeborne/selenide/Condition.java
+++ b/src/main/java/com/codeborne/selenide/Condition.java
@@ -16,6 +16,7 @@ import com.codeborne.selenide.conditions.Exist;
 import com.codeborne.selenide.conditions.ExplainedCondition;
 import com.codeborne.selenide.conditions.Focused;
 import com.codeborne.selenide.conditions.Hidden;
+import com.codeborne.selenide.conditions.IsImage;
 import com.codeborne.selenide.conditions.MatchText;
 import com.codeborne.selenide.conditions.NamedCondition;
 import com.codeborne.selenide.conditions.Not;
@@ -327,6 +328,11 @@ public abstract class Condition {
   public static Condition match(String description, Predicate<WebElement> predicate) {
     return new CustomMatch(description, predicate);
   }
+
+  /**
+   * Check if image is loaded.
+   */
+  public static final Condition isImage = new IsImage();
 
   /**
    * Check if browser focus is currently in given element.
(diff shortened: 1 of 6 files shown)


src/main/java/com/codeborne/selenide/Condition.java
```diff
@@ -328,6 +329,11 @@ public static Condition match(String description, Predicate<WebElement> predicat
     return new CustomMatch(description, predicate);
   }
 
+  /**
+   * Check if image is loaded.
+   */
+  public static final Condition isImage = new IsImage();
```
asolntsev: let's rename it to `image`.

* This is not readable: `$.should(isImage);`
* This is readable: `$.shouldBe(image);`

src/main/java/com/codeborne/selenide/conditions/IsImage.java
```diff
@@ -0,0 +1,22 @@
...
+  @Override
+  public boolean apply(Driver driver, WebElement webElement) {
+    if (!"img".equalsIgnoreCase(webElement.getTagName())) {
+      return false;
+    }
+    return driver.executeJavaScript("return arguments[0].complete && " +
+      "typeof arguments[0].naturalWidth != 'undefined' && " +
+      "arguments[0].naturalWidth > 0", webElement);
+  }
+}
```
asolntsev: Now we have code duplication in `com.codeborne.selenide.commands.IsImage` and `com.codeborne.selenide. conditions.IsImage`. I think one of them should use the other.

dstekanov: Indeed! 
Will it be ok if I rename `conditions.IsImage` to `IsImageLoaded` and inside create public static method `isImage()` and use this method from commands.IsImage? 

Renaming will help to not write the full classpath (with package name)

So in commands.IsImage it will look like this: `return IsImageLoaded.isImage(locator.driver(), img);` ?

asolntsev: Yes, why not.

src/main/java/com/codeborne/selenide/Condition.java
dstekanov: done

src/main/java/com/codeborne/selenide/conditions/IsImage.java
dstekanov: done

dstekanov:
not sure why 'check' vailed...

BorisOsipov:
>not sure why 'check' failed...

Will resolved after https://github.com/selenide/selenide/pull/1088

BorisOsipov:
@dstekanov rebase onto master, please.

Issue #1014, opened 2019-11-11
Add support for Shadow DOM


asolntsev:
###  The problem

Shadow DOM is used by any framework that supports web components. For example, when testing a Vaadin app with Selenium it can become necessary to look inside the Shadow DOM of some of the Vaadin web components. If you go to the Vaadin homepage and then click "Login" at the top right corner to make the login dialog appear, then look for the element with the id "login-username". It is a Vaadin web component, which models an input field, but the real input element is inside its Shadow DOM. If you open the #shadow-root element in the Firefox or Chrome inspector, you'll find a HTML input element inside. And that element is where I would like to enter my text in a Selenide test.

>  I saw an example implementation for a selector "ByShadowCss" on this Stack Overflow thread (last answer on page):
> https://stackoverflow.com/questions/23920167/accessing-shadow-dom-tree-with-selenium

See this thread for more details: https://groups.google.com/forum/#!topic/selenide/sCbHqePe13M

dstekanov:
Hi @asolntsev! 
Is it a good idea to have such syntax `$(ByShadow.css("shadow-host >>> #search-button"));` ?

asolntsev:
@dstekanov Hm... I don't know. What is current alternative for this selector?

dstekanov:
I've experimented with it... and as I understood, in order to find elements inside the Shadow-DOM tree, first, we need to find the shadow host element to which the shadow dom is attached (so we got shadow-root element). 

I thought to make something like this: `$(Selectors.shadowCss("#shadow-host")).$("#element-inside-shadow-dom").shouldHave(text("some text"));`

First, to find shadow-host and return its shadow root. For example:
```
public WebElement findElement(SearchContext context) {
  WebElement result = ((FindsByCssSelector) context).findElementByCssSelector(cssSelector);
  return (WebElement) getJavascriptExecutor(context).executeScript("return arguments[0].shadowRoot", result);
}
```
This approach works in Chrome (I have a working copy for it)

But in Firefox `arguments[0].shadowRoot` returns the map with different attributes (not sure if I understand it completely), so the only way to start searching the element is to use  `arguments[0].shadowRoot.querySelector('target-element')` (it will work in both browsers)

As alternative, maybe something like this, because also we can have inner shadow-dom elements
`shadow$(target-element, shadow-host...).shouldHave(text("some text"));`

asolntsev:
Implemented in https://github.com/selenide/selenide/pull/1090 (Selenide 5.10.0)

PR #1090, opened 2020-03-14
Add shadow-dom support

dstekanov:
## Proposed changes
Add support for searching elements inside Shadow-DOM tree. Currently, it does not have native support by WebDriver so it is a good helper who faced with such elements.

In order to be able to findElement inside shadow-dom, you have to specify its shadow-host element (it is a root element that has attached shadow tree). There is a good explanation about it https://javascript.info/shadow-dom 

Also, one note is that if you have inner shadow roots, you must explicitly specify each its shadow host.

It resolves a feature request #1014 

## Checklist

### DIFF
--- a/statics/src/test/java/integration/ShadowDomInsideIFrameTest.java
+++ b/statics/src/test/java/integration/ShadowDomInsideIFrameTest.java
@@ -9,6 +9,8 @@ import static com.codeborne.selenide.Condition.value;
 import static com.codeborne.selenide.Selectors.shadowCss;
 import static com.codeborne.selenide.Selenide.$;
 import static com.codeborne.selenide.Selenide.switchTo;
+import static com.codeborne.selenide.WebDriverRunner.isFirefox;
+import static org.junit.jupiter.api.Assumptions.assumeFalse;
 
 class ShadowDomInsideIFrameTest extends IntegrationTest {
   @BeforeEach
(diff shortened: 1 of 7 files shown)


statics/src/test/java/integration/ShadowDomInsideIFrameTest.java
```diff
@@ -0,0 +1,39 @@
...
+
+  @Test
+  void getTargetElementViaShadowHostInsideIFrame() {
+    switchTo().frame("iframe_page");
+    $(shadowCss("p", "#shadow-host"))
+      .shouldHave(text("Inside Shadow-DOM"));
+  }
+
+  @Test
+  void setValueViaShadowDomInsideIFrame() {
```
BorisOsipov: add assume not FF check to avoid test fail on FF

dstekanov: thanks! done

Yohandah:
How to use this ?

How would you request `toast-message` in this example ?

I tried the following but it doesn't work like this

`$(By.tagName("ion-toast")).shadowRoot.$(By.className("toast-message")).getText();`


image

QuiWorks:
From Looking at the changes in this pull request (see SelectorsTest#byShadowCss) and the [java doc](https://selenide.org/javadoc/current//com/codeborne/selenide/selector/ByShadow.html) this seems like the intended use:

`SelenideElement element = $(ByShadow.cssSelector(targetCssSelector, hostCssSelector));`

DieterRogiest:
see example in the blog https://selenide.org/2020/03/18/selenide-5.10.0/ : 

`$(shadowCss("p", "#shadow-host", "#inner-shadow-host")).shouldHave(text("The Shadow-DOM inside another shadow tree"));`

PR #1091, opened 2020-03-14
Migrated Guava API to the equivalent Java API

wlsc:
## Proposed changes
~~Migrated Guava's Predicate to the equivalent Java's Predicate~~
Migrated Guava API to the equivalent Java API 

## Checklist
- [X] Executed same tests which are still green.

### DIFF
--- a/src/main/java/com/codeborne/selenide/CollectionCondition.java
+++ b/src/main/java/com/codeborne/selenide/CollectionCondition.java
@@ -13,7 +13,7 @@ import com.codeborne.selenide.collections.SizeNotEqual;
 import com.codeborne.selenide.collections.Texts;
 import com.codeborne.selenide.collections.TextsInAnyOrder;
 import com.codeborne.selenide.impl.WebElementsCollection;
-import com.google.common.base.Predicate;
+import java.util.function.Predicate;
 import org.checkerframework.checker.nullness.qual.Nullable;
 import org.openqa.selenium.WebElement;
 
@@ -166,8 +166,8 @@ public abstract class CollectionCondition implements Predicate<List<WebElement>>
     }
 
     @Override
-    public boolean apply(@Nullable List<WebElement> input) {
-      return delegate.apply(input);
+    public boolean test(@Nullable List<WebElement> input) {
+      return delegate.test(input);
     }
   }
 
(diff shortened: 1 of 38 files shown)


vinogradoff:
Can we then get rid if guava dependency or is
It used otherwise?

wlsc:
@vinogradoff it is used for other things, which could be probably also removed, but I haven't analyzed it yet.

vinogradoff:
My point is that getting rid of a dependency in favor of the same function of Java Standard Library has its merit. 

I am not sure,  if refactoring of only a part of dependencies brings benefits.

wlsc:
The point of this PR is to get rid of unnecessary Predicate since the codebase is using Java 1.8 as a base line version.
Moreover, every refactoring should be a step by step procedure to ensure it won't break anything in progress. This is such a step.
In additional, I can have Guava removed by just using Java 1.8 level API, but by doing this, I would need to exclude Guava from transitive dependencies, which is more in-depth intrusion into project's configuration.

wlsc:
ok, I've just pushed changes to remove complete usage of Guava framework. Is this something you searched for?

vinogradoff:
Looks okay for me.
@BorisOsipov @rosolko @asolntsev - second opinion?

BorisOsipov:
`CollectionCondition` is a public api. If we move from `guava Predicate` to `java.util` will we break all users custom collection conditions? Seems yes.

BorisOsipov:
But in general, I am ok with these changes

vinogradoff:
@BorisOsipov very good observation, but on the first sight, the `apply` in custom conditions is another one, from `Condition.java` which is not changed. I haven't tried it yet, but it is possible that the custom condition will still work. 

We need to check it.

vinogradoff:
@wlsc I haven't found any test on custom condition in selenide yet. If you had time to write on, it would prove the assumptions and help us to refactor easily in future :)

Here is some docs on it: https://github.com/selenide/selenide/wiki/Custom-conditions

asolntsev:
@vinogradoff @wlsc @BorisOsipov Yes, this PR will definitely break the backward compatibility. 
People who implemented their own `Condition`s will need to rename method `apply` to `test`. 

But it should be quite easy, and people usually don't have too many of custom conditions (I guess).
So probably it's ok to release such a change in Selenide 5.10.0

PR #1095, opened 2020-03-21
Adding a way to define additional options

SeleniumTestAB:
## Proposed changes

The user will be able to define his/her additional options without need to create additional class that implements WebDriverProvider interface. This could result in less complex code and more verbose way of defing settings by the user.

This feature supports Chrome, Firefox and Opera browser because rest of the browser still use Capabilities instead of options.

## Checklist


BorisOsipov:
There is already way how to pass any additional options to driver factories. It is `Configuration.browserCapabilities`. We dont need some new way to do it only for options.

SeleniumTestAB:
@BorisOsipov Capabilities itself are deprecated so its better to set those via options not to mention that only real reason to make WebDriverProvider is to set those options when you just simply could add them via Configuration property.

BorisOsipov:
>Capabilities itself are deprecated

Nope.
See docs - https://www.selenium.dev/selenium/docs/api/java/org/openqa/selenium/remote/DesiredCapabilities.html

> its better to set those via options not to mention that only real reason to make WebDriverProvider is to set those options when you just simply could add them via Configuration property.

Once again you can do the same by `Configuration.browserCapabilities` and it will work for all browsers not only for chrome\ff\opera.

We definitely will get rid of capabilities, when driver constructors will completely remove [this api](https://www.selenium.dev/selenium/docs/api/java/org/openqa/selenium/chrome/ChromeDriver.html#ChromeDriver-org.openqa.selenium.Capabilities-).

Now I don't understand why we have to create two ways to do the same and confuse people how combinations of this will work.

BorisOsipov:
@SeleniumTestAB I understand your conserns. Let's continue conversation in https://github.com/selenide/selenide/issues/1097

PR #1098, opened 2020-03-23
Normalize reports URL with spaces

### LINKED ISSUE #1072: Selenide generates invalid URL for screenshot if test name contains spaces
Assuming we have test name with spaces (for example, in Kotlin):

```java
class GoogleTest {
  @Test
  fun `can search any keyword`() {
    ...
  }
}
```

Selenide generates an error message with invalid URLs:

```java
Element not found {...}
Expected: ...
Screenshot: https://jenkins.some.com/job/blah/42/artifact/build/reports/tests/ui/GoogleTest/can search any keyword/1582843426252.1.png
Page source: https://jenkins.some.com/job/blah/42/artifact/build/reports/tests/ui/GoogleTest/can search any keyword/1582843426252.1.html
```

* **Selenide Version**:  5.7.0

dstekanov:
## Proposed changes

Fix #1072 

## Checklist
- [x] Checkstyle and unit tests pass locally with my changes by running `gradlew check chrome_headless firefox_headless` command

### DIFF
--- a/src/main/java/com/codeborne/selenide/impl/ScreenShotLaboratory.java
+++ b/src/main/java/com/codeborne/selenide/impl/ScreenShotLaboratory.java
@@ -348,19 +348,9 @@ public class ScreenShotLaboratory {
     }
 
     if (driver.config().reportsUrl() != null) {
-      Path current = Paths.get(System.getProperty("user.dir"));
-      Path target = Paths.get(screenshot).normalize();
-      String screenshotUrl;
-      if (isInsideFolder(current, target)) {
-        Path relativePath = current.relativize(target);
-        screenshotUrl = driver.config().reportsUrl() + relativePath.toString().replace('\\', '/');
-      } else {
-        String name = target.toFile().getName();
-        screenshotUrl = driver.config().reportsUrl() + name;
-      }
-      screenshotUrl = normalizeURL(screenshotUrl);
-      log.debug("Replaced screenshot file path '{}' by public CI URL '{}'", screenshot, screenshotUrl);
-      return screenshotUrl;
+      String screenShotURL = formatScreenShotURL(driver.config().reportsUrl(), screenshot);
+      log.info("Replaced screenshot file path '{}' by public CI URL '{}'", screenshot, screenShotURL);
+      return screenShotURL;
     }
 
     log.debug("reportsUrl is not configured. Returning screenshot file name '{}'", screenshot);
@@ -369,6 +359,18 @@ public class ScreenShotLaboratory {
     } catch (MalformedURLException e) {
       return "file://" + screenshot;
     }
+  }
+
+  private String formatScreenShotURL(String reportsURL, String screenshot) {
+    Path current = Paths.get(System.getProperty("user.dir"));
+    Path target = Paths.get(screenshot).normalize();
+    String screenShotPath;
+    if (isInsideFolder(current, target)) {
+      screenShotPath = current.relativize(target).toString().replace('\\', '/');
+    } else {
+      screenShotPath = target.toFile().getName();
+    }
+    return normalizeURL(reportsURL + screenShotPath);
   }
 
   private String normalizeURL(String url) {
--- a/src/main/java/com/codeborne/selenide/impl/ScreenShotLaboratory.java
+++ b/src/main/java/com/codeborne/selenide/impl/ScreenShotLaboratory.java
@@ -22,17 +22,19 @@ import java.io.FileInputStream;
 import java.io.FileOutputStream;
 import java.io.IOException;
 import java.io.InputStream;
+import java.io.UnsupportedEncodingException;
 import java.net.MalformedURLException;
-import java.net.URI;
-import java.net.URISyntaxException;
-import java.net.URL;
+import java.net.URLEncoder;
+import java.nio.charset.StandardCharsets;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
 import java.util.concurrent.atomic.AtomicLong;
 import java.util.function.Function;
+import java.util.stream.Collectors;
 
 import static java.io.File.separatorChar;
 import static org.openqa.selenium.OutputType.FILE;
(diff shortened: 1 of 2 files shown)


src/main/java/com/codeborne/selenide/impl/ScreenShotLaboratory.java
```diff
@@ -356,11 +358,7 @@ public String formatScreenShotPath(Driver driver) {
         String name = target.toFile().getName();
         screenshotUrl = driver.config().reportsUrl() + name;
       }
-      try {
-        screenshotUrl = new URL(screenshotUrl).toExternalForm();
-      } catch (MalformedURLException ignore) {
-        // ignored exception
-      }
+      screenshotUrl = normalizeURL(screenshotUrl);
```
vlsi: Should this be removed at all?

It looks like the problem is the URL is combined as string, then it is `normalized`.

Why don't we create the URL via the proper APIs in the first place?

For instance, `screenshotUrl = new URL(.., driver.config().reportsUrl(), ..., name, ...)` and so on.

dstekanov: Hi @vlsi! thanks for your review! One moment is... some URLs can contain special characters, for example, the space character, and to make these characters legal they need to be encoded. So the multi-argument URI constructor automatically takes care of the encoding...

vlsi: > So the multi-argument URI constructor automatically takes care of the encoding...

Well, I thought exactly behind those lines, and I wonder why do you need **both** `new URL` and `new URI` at the same time.

Isn't just one of them enough?

dstekanov: For example, new URL("http", "ci.org/with space/", "path with spaces/123.html") does not encode it... 

Creating URL first helps me to take its parts and pass it to URI because If I pass the whole string to URI (e.g. new URI("http://ci.org/with space/path with spaces/123.html")) it fails too... )) and I am not 100% sure that it is the best solution... So I hope that I explained why I did it

vlsi: How about `java.net.URLEncoder.encode(reportsURL, java.nio.charset.StandardCharsets.UTF_8.name()) + encode(screenShotPath, StandardCharsets.UTF_8.name())` ?

src/test/java/com/codeborne/selenide/impl/ScreenShotLaboratoryTest.java
```diff
@@ -96,6 +101,17 @@ void collectsAllScreenshots() {
       .hasToString("12356789.4");
   }
 
+  @Test
+  void canFormatReportsURLWithSpaces() {
+    Driver driver = new DriverStub(new SelenideConfig().reportsUrl("http://ci.org/with space/"),
```
vlsi: Is `"http://ci.org/with%20space/"` legal to be used in this context? Should it be different from `"http://ci.org/with space/"` or should it be the same?

Would you please add a test for `"http://ci.org/with%20space/"` as well?

src/main/java/com/codeborne/selenide/impl/ScreenShotLaboratory.java
dstekanov: yes, I tried it, and for example "http://ci.org" will result in "http%3A%2F%2Fci.org"

but I am not sure if it correct

vlsi: Oh, you are right.

So what should be the behavior if `reportsURL` does contain `?param=value` part?
Is it forbidden?

Where should `screenShotPath` be appended?

src/test/java/com/codeborne/selenide/impl/ScreenShotLaboratoryTest.java
dstekanov: Yes, it is a rare case when URL can contain space characters or when there is a folder on CI with spaces. But the main goal was to escape special characters when, for example, the test method has spaces, like described there #1072

vlsi: I see, however, it looks weird to concatenate strings with different `encoding` nature.

In other words, `reportsURL` is already an encoded URL, however, file path is not encoded.

So I think the approach should be **skip** encoding for `reportsURL` and **perform** encoding of the path added.

That, however, keeps the question of how the URL should look like if `reportsURL` is like `https://host:port/get?param=value`

src/main/java/com/codeborne/selenide/impl/ScreenShotLaboratory.java
dstekanov: Hm, seems it will not work... And previously it did not work too... I think that no one used it :) 

And I don't know if it should work.. for example, if the user provides `http://ci.org/atifacts?param=value` and when we append screenshot path, like `http://ci.org/atifacts/screenshotpath?param=value` it breaks the logic, isn't it?

dstekanov: As a solution, I think maybe we can document `reportsUrl` to provide baseUrl without query parameters

src/test/java/com/codeborne/selenide/impl/ScreenShotLaboratoryTest.java
asolntsev: The specification says that every part of URL should be encoded separately. 
I think we can assume that `reportsURL` is already encoded. 

So, the resulting code would be 
`reportsURL + "?" + encode(paramName) + "=" + encode(paramValue)`

See https://www.talisman.org/~erlkonig/misc/lunatech%5Ewhat-every-webdev-must-know-about-url-encoding/#DonotconstructURLswithoutencodingeachpart

dstekanov: I am really confused about this query in base reports URL. From my understanding, they are needed to send some parameters to a specific resource on the server. So as @vlsi said, where should we append it in the resulted URL?

For example:
if someone provides `http://ci.mycompany.com/job/666/artifact?some=value` and we append path `build/reports/tests/screenshot.html` how should look the resulting URL? 
Are these correct?
`http://ci.mycompany.com/job/666/artifact?some=value/build/reports/tests/screenshot.png` or
`http://ci.mycompany.com/job/666/artifact/build/reports/tests/screenshot.png?some=value`

At least (if without query) I can push another version where I encode the `path` without `reportUrl`

asolntsev: I assume that `reportsUrl` does NOT contain any parameters. 
* `http://ci.mycompany.com/job/666/artifact?some=value` is *invalid* value
* `http://ci.mycompany.com/job/666/artifact/path%20can%20contain%20spaces/` is a *valid* value.

Issue #996, opened 2019-10-17
Feature request: add Condition attributeContainsValue (java.lang.String attributeName, java.lang.String matchedExpectedAttributeValue)


sergeyrybakin:
Feature request
## The problem
Condition attribute requires **exact** text of attribute value.
Sample: $("#mydiv").shouldHave(attribute("fileId", "12345"));

It would be great to have Condition to verify attribute by matched text:
Example:
attribute "field" can contains value: "Job12345", "12345 job", etc. Where "12345" - is matched text.
$("#mydiv").shouldHave(attributeContainsValue("fileId", "12345"));

## My environment
* **Selenide Version:5.3.1
* **Chrome\Firefox\IE Version: 77.0.3865.120\69.0.2
* **Browser Driver Version:
* **Selenium Version:3.141.59
* **OS Version:Ubuntu 18.04.3 LTS

## Code To Reproduce Issue [ Good To Have ]
Please remember that with sample code it's easier to reproduce the bug.
It's not good example, but:
http://the-internet.herokuapp.com/dynamic_content?with_content=static

$("content img").shouldHave(attributeContainsValue("src", "/img/avatars/Original-Facebook-Geek-Profile-Avatar-"));

rosolko:
@sergeyrybakin Thanks for a report!

 In selenide we trying to hold only few _basic_ conditions. 
**But!** You can easily implement this condition by your own - just need to implement `Condition`.
Also you can find examples on our [wiki](https://github.com/selenide/selenide/wiki/Community-conditions) page.

Is this message helps you?
Write back please with answer and extend our wiki with you condition implementation.
Feel free to close this issue if no question left.

Thanks!

asolntsev:
@rosolko What about overloaded condition `attribute` with regexp parameter?

Something like this:

```java
$("#mydiv").shouldHave(attribute(Pattern.compile(".*12345.*")));

// or

$("#mydiv").shouldHave(attributeMathching(".*12345.*"));
```

PR #1100, opened 2020-03-26
Add MatchAttributeWithValue condition

dstekanov:
## Proposed changes

If resolves a feature request #996

## Checklist
- [x] Checkstyle and unit tests pass locally with my changes by running `gradlew check chrome_headless firefox_headless` command

### DIFF
--- a/src/main/java/com/codeborne/selenide/Condition.java
+++ b/src/main/java/com/codeborne/selenide/Condition.java
@@ -121,7 +121,7 @@ public abstract class Condition {
   /**
    * Assert that given element's attribute matches given regular expression
    *
-   * <p>Sample: <code>$("h1").should(attributeMatching(".*12345.*"))</code></p>
+   * <p>Sample: <code>$("h1").should(attributeMatching("fileId", ".*12345.*"))</code></p>
    *
    * @param attributeName  name of attribute
    * @param attributeRegex regex to match attribute value
@@ -275,7 +275,7 @@ public abstract class Condition {
   }
 
   /**
-   * Assert that element is exactly (case insensitive) given text
+   * Assert that element has exactly (case insensitive) given text
    * <p>Sample: <code>$("h1").shouldHave(exactText("Hello"))</code></p>
    *
    * <p>Case insensitive</p>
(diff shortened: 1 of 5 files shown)


src/main/java/com/codeborne/selenide/Condition.java
```diff
@@ -117,6 +118,18 @@ public static Condition attribute(String attributeName, String expectedAttribute
     return new AttributeWithValue(attributeName, expectedAttributeValue);
   }
 
+  /**
+   * Assert that given element's attribute matches given regular expression
+   *
+   * <p>Sample: <code>$("h1").should(attributeMatching(".*12345.*"))</code></p>
```
asolntsev: The Sample is invalid: `attributeMatching` has 2 parameters: `name` and `value`.

src/main/java/com/codeborne/selenide/Condition.java
```diff
@@ -262,7 +275,7 @@ public static Condition textCaseSensitive(String text) {
   }
 
   /**
-   * Assert that element is exactly (case insensitive) given text 
+   * Assert that element is exactly (case insensitive) given text
```
asolntsev: "is" -> "has"

src/main/java/com/codeborne/selenide/conditions/MatchAttributeWithValue.java
```diff
@@ -0,0 +1,40 @@
...
+import static java.util.regex.Pattern.DOTALL;
+
+public class MatchAttributeWithValue extends Condition {
+  private final String attributeName;
+  private final Pattern attributeRegex;
+
+  public MatchAttributeWithValue(String attributeName, String attributeRegex) {
+    super("match attribute");
+    this.attributeName = attributeName;
+    this.attributeRegex = Pattern.compile(attributeRegex, DOTALL);
```
asolntsev: Why DOTALL?
As I know, it's needed to support multiline Strings, which hardly can happen to attribute values.

src/main/java/com/codeborne/selenide/Condition.java
dstekanov: thanks, done!

src/main/java/com/codeborne/selenide/conditions/MatchAttributeWithValue.java
dstekanov: Thanks for review Andrei! 
You are right! I removed the flag, seems this is not needed there (at least, I did not find appropriate)

PR #1106, opened 2020-04-07
Add CheckReturnValue annotation for methods that only return value

jreznot:
## Proposed changes
Here we add `@CheckReturnValue` annotation from checker framework to methods that only return values and do not have any additional visible effects.
This change helps to spot a bug in IDE when someone uses method that does not have effects but only returns value, example from real tests:
```
$("div").text(); // the value is not used
```

## Checklist

FYI: For some reason checkstyle says that `Class Fan-Out Complexity is 30 (max allowed is 29).` for Condition class.

### DIFF
--- a/config/checkstyle/checkstyle.xml
+++ b/config/checkstyle/checkstyle.xml
@@ -54,7 +54,7 @@
     <!--<module name="MutableException"/>-->
     <module name="ClassFanOutComplexity">
       <property name="max" value="29"/>
-      <property name="excludeClassesRegexps" value=".*Condition.*"/>
+      <property name="excludedPackages" value="com.codeborne.selenide.conditions"/>
     </module>
     <module name="CyclomaticComplexity">
       <property name="max" value="11"/>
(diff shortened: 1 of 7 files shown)


BorisOsipov:
@jreznot does IntelliJ IDEA support @CheckReturnValue? I mean does it highlight such methods as warning\error in editor?

jreznot:
> @jreznot does IntelliJ IDEA support `@CheckReturnValue`? I mean does it highlight such methods as warning\error in editor?

Sure, it supports `@CheckReturnValue` )

With my changes:
image

asolntsev:
@jreznot Thank you for the PR!
I have two concerns:
1. Shouldn't we also add `@CheckReturnValue` to all methods of `SelenideDriver`? 
   I mean, the following code should also report a warning:
```java
  SelenideDriver driver = new SelenideDriver();
  driver.$("div");
```
2. When we add another method to Selenide, it's too easy to forget to mark it with `@CheckReturnValue`. It will happen. :(

jreznot:
I will add annotations for SelenideDriver too
>     1. When we add another method to Selenide, it's too easy to forget to mark it with `@CheckReturnValue`. It will happen. :(

It is not a big deal. Selenide is a mature library, we only need hints for heavily used APIs.

PR #1107, opened 2020-04-07
Use final for fields where it is possible

jreznot:
## Proposed changes

This small change marks fields as final so they cannot be changed by accident in the future.

## Checklist

### DIFF
--- a/src/main/java/com/codeborne/selenide/drivercommands/Navigator.java
+++ b/src/main/java/com/codeborne/selenide/drivercommands/Navigator.java
@@ -20,7 +20,7 @@ import static com.codeborne.selenide.logevents.LogEvent.EventStatus.PASS;
 import static java.util.regex.Pattern.DOTALL;
 
 public class Navigator {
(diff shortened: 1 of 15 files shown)


src/main/java/com/codeborne/selenide/drivercommands/Navigator.java
```diff
@@ -20,9 +20,9 @@
 import static java.util.regex.Pattern.DOTALL;
 
 public class Navigator {
-  protected Pattern absoluteUrlRegex = Pattern.compile("^[a-zA-Z]+:.*", DOTALL);
+  protected static final Pattern ABSOLUTE_URL_REGEX = Pattern.compile("^[a-zA-Z]+:.*", DOTALL);
```
asolntsev, line 23: My initial idea was to make all such fields protected and mutable to allow users to customize Selenide behaviour or fix some bugs on the fly.

But well, I guess nobody actually uses it, so let's make them final.
But I would also make it private then.

Issue #1119, opened 2020-04-15
Add option `load-extension` for Chrome (to avoid annoying popup)


asolntsev:
## Problem
After upgrading to Selenide 5.10.0 some folks see a popup in Chrome:
![image](https://user-images.githubusercontent.com/279773/79391813-e3095a00-7f7a-11ea-9b33-35a2ef08dc3c.png)

See https://ru.selenide.org/2020/03/18/selenide-5.10.0/#comment-4859162417

## Solution
Instead of current line 
```
options.setExperimentalOption("excludeSwitches", new String[]{"enable-automation"});
```

we should add 
```
options.setExperimentalOption("excludeSwitches", new String[]{"enable-automation", "load-extension"});
```
## Tell us about your environment
* **Selenide Version**:  5.10.1
* **Chrome**:

asolntsev:
Done in https://github.com/selenide/selenide/pull/1120

PR #1103, opened 2020-04-05
#1097 Allow to pass driver specific options from configuration

BorisOsipov:
## Proposed changes
Allow to change `*Options` rigth before Selenide creates driver by driver factories.
Resolve #1097 Allow to pass driver specific options from configuration
Fix #676

## Checklist
- [ ] Checkstyle and unit tests are passed locally with my changes by running `gradlew check chrome_headless firefox_headless` command
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have added necessary documentation (if appropriate)

PR #1120, opened 2020-04-15
#1119 disable annoying popup about extensions in Chrome

asolntsev:
## Proposed changes

## Checklist


dstekanov:
I've never used own extensions with Selenium, but I saw how it can be added:

```
ChromeOptions options = new ChromeOptions();
options.addExtensions(new File("/path/to/extension.crx"));
```

And let's consider the following because I am not sure for 100% and did not try it...
if we implement this #1103 then the user will be able to add extension without WebDriverProvider and his extension will be disabled

asolntsev:
@dstekanov It's true, out new setting will disable extensions. 
But I assume that
1. Most of users don't use Chrome extensions in tests, and
2. Those who use extensions create their custom `ChromeOptions` anyway. 

So it seems to me a safe change useful for most users.

dstekanov:
Agree with you!

PR #1121, opened 2020-04-16
Screenshot fullpage

jpratt2:
## Proposed changes
This adds the screenshotFullpage method for full-page screenshot functionality for Firefox and Chrome by using the Shutterbug plugin.

## Checklist
I have not. Frankly, I am not sure how to write at test, but I have tested it on my own machine.
Would someone be willing to help write a Mockito test?
I have not. I am willing to do so later, however.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## 5.11.0
 * #996 Add MatchAttributeWithValue condition -- thanks to Dmytro Stekanov for PR #1100
+* Add CheckReturnValue annotation for methods that only return value  --  thanks to Yuriy Artamonov for PR #1106
+* Add missing byTagName to Selectors to make it consistent with By  --  thanks to Yuriy Artamonov for PR #1104
 
 ## 5.10.0 (released 18.03.2020)
 * #1014 Add support for Shadow DOM  --  thanks to Dmytro Stekanov for #1090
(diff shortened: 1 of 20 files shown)


asolntsev:
@jpratt2 First of all, thank you for your contribution!

Concering this PR, I am in doubt.
1. I am not sure it's a good idea to add yet another dependency to Selenide (selenium-shutterbug). Especially of version 0.x 
What of this library will not be maintained? Why not some other similar library (for example, Ashot)? 
2. User need to expicitly call a new method ` screenshotFullpage()`. User can easily forget it.

I dream about creating a "plugin system" in Selenide. So that your solition is like a "plugin" or "extension" for Selenide. Anyone who needs "full-page" screenshot will just add your plugin, and Selenide starts taking full-page screenshots automatically. No need to call a separate method. What do you think?

jpratt2:
OK, what do you think of this?
We could start some GitHub pages, and one of them could be titled "Selenide Add-ons".
Contributors would be able to post recommendations (hopefully honest ones).

https://pages.github.com/

asolntsev:
@jpratt2 Definitely, it's a good idea. We already have a similar page for conditions: https://github.com/selenide/selenide/wiki/Custom-conditions

We should create the same page for add-ons. 
And we should promote it better :)

PR #1122, opened 2020-04-17
Speedup tests: use short timeout by default

asolntsev:
## Proposed changes
* use timeout 1 ms in most tests

## Checklist

### DIFF
--- a/src/test/java/integration/TabsTest.java
+++ b/src/test/java/integration/TabsTest.java
@@ -26,7 +26,7 @@ class TabsTest extends ITest {
   @Test
   @Video
   void userCanBrowseTabs_webdriver_api() {
-    openFile("page_with_tabs.html");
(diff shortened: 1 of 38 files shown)


src/test/java/integration/TabsTest.java
```diff
@@ -26,7 +26,7 @@ void setUp() {
   @Test
   @Video
   void userCanBrowseTabs_webdriver_api() {
-    openFile("page_with_tabs.html");
+    setTimeout(1000);
```
rosolko, line 29: Isn't `openFile("page_with_tabs.html");` removed by mistake?

asolntsev: No, the same line `openFile("page_with_tabs.html");` is executed in `setUp()` method.

Issue #1002, opened 2019-10-21
screenshots are duplicated when using ScreenShooterExtension and Allure


asolntsev:
See comment https://github.com/selenide/selenide/pull/946#issuecomment-507961860

## The problem
**Given** a project using
 * `com.codeborne.selenide.junit5.ScreenShooterExtension`
* Allure 

**When** a test fails
**Then** two screeenshots are takes.

## Details
1. `ScreenShooterExtension` takes a screenshot (and just logs it),
2. Then `ScreenShooterExtension` triggers Allure listener (via `SelenideLogger.commitStep`) which also takes a screenshot. 

It would be nice to avoid taking 2 screenshots.

## Tell us about your environment
* **Selenide Version**: 5.4.1

SeleniumTestAB:
@asolntsev would it be hard to check if logger with name "allure" is present and make condition from it?

asolntsev:
@SeleniumTestAB it's technically possible, but it doesn't seem to be a good idea in terms of design. Selenide core code should not depend on any listeners.
I would rather suggest that AllureListener could check if Selenide has already taken a screenshot during current test. Maybe we should even create a superclass for such listeners in Selenide.

SeleniumTestAB:
@asolntsev indeed it seems that fix should be on allure side rather. As for superclass, doesnt SelenideLogger fulfill such role?

dstekanov:
Hi @asolntsev, 
I was not able to reproduce this issue, and when I looked inside `ScreenShooterExtension` I did not find `SelenideLogger.commitStep` so seems this issue can be closed.

asolntsev:
@dstekanov I haven't tried to re-reproduce it yet, but my logic says that
1. When `ScreenShooterExtension` works without Allure, it takes screenshot.
2. When Allure works without `ScreenShooterExtension`, it also takes screenshot. 
3. When `ScreenShooterExtension and Allure are both enabled, they both must take their screenshots (they don't call each other to ask if there is already screenshot taken). 

Am I right?

dstekanov:
Now I see and seems it's not a problem with `ScreenShooterExtension` 

`ScreenShooterExtension` does not take (by default) screenshots for `UIAssertionError`.
So when such `non-UIAssertionError` happens, e.g. with `org.junit.jupiter.api.Assertions`,  `ScreenShooterExtension`takes a screenshot but it does not update status in LogEvent and Allure does not take a screenshot and there is no attachment in the report...

[...]

PR #1125, opened 2020-04-18
Add opportunity to get screenshots from current thread

### LINKED ISSUE #1029: Add opportunity to get screenshot from current thread via ScreenShot Laboratory
For https://github.com/selenide/selenide/blob/master/src/main/java/com/codeborne/selenide/impl/ScreenShotLaboratory.java add opportunity to get latest or all screenshots from current thread
Something like that:
```
  public File getLastThreadScreenshot() {
      return currentContextScreenshots.isEmpty() ? null : currentContextScreenshots.get(currentContextScreenshots.size() - 1);
  }
```
or
```
  public List<File> getLastThreadScreenshot() {
      return currentContextScreenshots.isEmpty() ? null : currentContextScreenshots.get();
  }
```
or add methods for working with **currentContext** variable without remove it.
Also don't forget add some logic into https://github.com/selenide/selenide/blob/master/statics/src/main/java/com/codeborne/selenide/Screenshots.java

dstekanov:
## Proposed changes
Added possibility to get all screenshots and the last screenshot from the current thread and context thread.

Getting screenshots from `currentContextScreenshots` fixes #1029
Getting screenshots from `threadScreenshots` I hope can be used to fix https://github.com/selenide/selenide/issues/1002 

## Checklist

### DIFF
--- a/src/main/java/com/codeborne/selenide/impl/ScreenShotLaboratory.java
+++ b/src/main/java/com/codeborne/selenide/impl/ScreenShotLaboratory.java
@@ -37,6 +37,7 @@ import java.util.function.Function;
 import java.util.stream.Collectors;
 
 import static java.io.File.separatorChar;
+import static java.lang.ThreadLocal.withInitial;
 import static org.openqa.selenium.OutputType.FILE;
 
 public class ScreenShotLaboratory {
@@ -52,6 +53,7 @@ public class ScreenShotLaboratory {
   protected AtomicLong screenshotCounter = new AtomicLong();
   protected ThreadLocal<String> currentContext = ThreadLocal.withInitial(() -> "");
   protected ThreadLocal<List<File>> currentContextScreenshots = new ThreadLocal<>();
+  protected ThreadLocal<List<File>> threadScreenshots = withInitial(ArrayList::new);
 
   protected ScreenShotLaboratory() {
   }
@@ -250,7 +252,14 @@ public class ScreenShotLaboratory {
     synchronized (allScreenshots) {
       allScreenshots.add(screenshot);
     }
+    addScreenshotToCurrentThread(screenshot);
     return screenshot;
+  }
+
+  private void addScreenshotToCurrentThread(File screenshot) {
+    List<File> files = threadScreenshots.get();
+    files.add(screenshot);
+    threadScreenshots.set(files);
   }
 
   protected File takeScreenshotInMemory(TakesScreenshot driver) {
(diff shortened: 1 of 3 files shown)


src/main/java/com/codeborne/selenide/impl/ScreenShotLaboratory.java
```diff
@@ -250,9 +252,16 @@ protected File addToHistory(File screenshot) {
...
       allScreenshots.add(screenshot);
     }
+    addScreenshotToCurrentThread(screenshot);
     return screenshot;
   }
 
+  private void addScreenshotToCurrentThread(File screenshot) {
+    List<File> files = threadScreenshots.get();
+    files.add(screenshot);
+    threadScreenshots.set(files);
```
asolntsev, line 262: I guess `threadScreenshots.set(files);` is not needed.

src/main/java/com/codeborne/selenide/impl/ScreenShotLaboratory.java
```diff
@@ -332,12 +341,37 @@ public void startContext(String context) {
...
+  public List<File> getThreadScreenshots() {
+    return Collections.unmodifiableList(threadScreenshots.get());
+  }
+
+  public List<File> getContextScreenshots() {
+    List<File> screenshots = currentContextScreenshots.get();
+    return Collections.unmodifiableList(screenshots == null
+      ? Collections.emptyList()
+      : screenshots);
+  }
```
asolntsev, line 353: I would change the order: there is no need to call `unmodifiableList` for `emptyList`:

```
return screenshots == null
      ? Collections.emptyList()
      : Collections.unmodifiableList(screenshots);
```

src/main/java/com/codeborne/selenide/impl/ScreenShotLaboratory.java
```diff
@@ -332,12 +341,37 @@ public void startContext(String context) {
...
   public File getLastScreenshot() {
     synchronized (allScreenshots) {
       return allScreenshots.isEmpty() ? null : allScreenshots.get(allScreenshots.size() - 1);
     }
   }
 
+  public File getLastThreadScreenshot() {
+    List<File> screenshots = threadScreenshots.get();
+    return screenshots.isEmpty()
+      ? null
```
asolntsev, line 364: probably return Optional instead of null?

src/main/java/com/codeborne/selenide/impl/ScreenShotLaboratory.java
```diff
@@ -332,12 +341,37 @@ public void startContext(String context) {
...
+      : screenshots.get(screenshots.size() - 1);
+  }
+
+  public File getLastContextScreenshot() {
+    List<File> screenshots = currentContextScreenshots.get();
+    return screenshots == null || screenshots.isEmpty()
+      ? null
```
asolntsev, line 371: probably return Optional instead of null?

src/main/java/com/codeborne/selenide/impl/ScreenShotLaboratory.java
dstekanov: yes, initially I thought to return Optional, but then decided to make it similar to existing getScreenthods()
but I agree, let's return Optional

Issue #1133, opened 2020-04-25
Please disable annoying popup "Save password?" in Chrome


asolntsev:
## Code to reproduce the issue:

```java
    open("https://the-internet.herokuapp.com/login");
    $("#username").val("tomsmith");
    $("#password").sendKeys("SuperSecretPassword!");
    $(".fa-sign-in").click();
```

## Tell us about your environment
* **Selenide Version**:  5.11.1
* **Chrome\Firefox\IE Version**: Chrome
* **Browser Driver Version**: 81

![image](https://user-images.githubusercontent.com/279773/80290485-e5736d00-874d-11ea-9739-b01c50dc102f.png)

asolntsev:
Done in https://github.com/selenide/selenide/pull/1134

PR #1134, opened 2020-04-25
#1133 disable "save password?" dialog

asolntsev:
## Proposed changes

## Checklist


src/test/java/com/codeborne/selenide/webdriver/ChromeDriverFactoryTest.java
```diff
@@ -64,8 +64,9 @@ void transferChromeOptionPreferencesFromSystemPropsToDriverNoAssignmentStatement
     ChromeOptions chromeOptions = new ChromeDriverFactory().createChromeOptions(config, proxy);
     Map<String, Object> prefsMap = getBrowserLaunchPrefs(ChromeOptions.CAPABILITY, chromeOptions);
 
-    assertThat(prefsMap).hasSize(2);
+    assertThat(prefsMap).hasSize(3);
```
vinogradoff, line 67: I think the lines 67,69,70 are "tests that are bound to implementation", which should be avoided in general.

asolntsev: I agree. We could check not the exact size, but onle the needed properties.

PR #1140, opened 2020-05-05
Nullity annotations for the most used APIs

jreznot:
## Proposed changes

At the moment, IntelliJ IDEA shows this warning in Kotlin because there is no nullity information in Selenide:

image

These changes improves IDE experience by providing exact type of parameters and return values including information on null/notnull values. This also makes Selenide work better with Kotlin as Kotlin now knows types and may omit excessive warnings and runtime checks.

image

## Checklist

Checkstyle warns about SelenideDriver for some reason.
```
[ant:checkstyle] [ERROR] /home/travis/build/selenide/selenide/src/main/java/com/codeborne/selenide/SelenideDriver.java:32:1: Class Fan-Out Complexity is 30 (max allowed is 29). [ClassFanOutComplexity]
```

### DIFF
--- a/src/main/java/com/codeborne/selenide/SelenideDriver.java
+++ b/src/main/java/com/codeborne/selenide/SelenideDriver.java
@@ -342,7 +342,7 @@ public class SelenideDriver {
   }
 
   @CheckReturnValue
-  @Nullable
+  @Nonnull
   public String url() {
     return getWebDriver().getCurrentUrl();
   }
--- a/src/main/java/com/codeborne/selenide/SelenideDriver.java
+++ b/src/main/java/com/codeborne/selenide/SelenideDriver.java
@@ -13,6 +13,9 @@ import org.openqa.selenium.WebDriver;
 import org.openqa.selenium.WebElement;
 import org.openqa.selenium.support.events.WebDriverEventListener;
 
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+import javax.annotation.ParametersAreNonnullByDefault;
 import java.io.File;
 import java.io.IOException;
 import java.net.URL;
@@ -26,6 +29,7 @@ import static java.util.Collections.emptyList;
 /**
  * "Selenide driver" is a container for WebDriver + proxy server + settings
  */
+@ParametersAreNonnullByDefault
 public class SelenideDriver {
   private static final Navigator navigator = new Navigator();
   private static final SelenidePageFactory pageFactory = new SelenidePageFactory();
@@ -53,11 +57,13 @@ public class SelenideDriver {
   }
 
   @CheckReturnValue
+  @Nonnull
   public Config config() {
     return config;
   }
 
   @CheckReturnValue
+  @Nonnull
   public Driver driver() {
     return driver;
   }
@@ -87,18 +93,21 @@ public class SelenideDriver {
   }
 
   @CheckReturnValue
+  @Nonnull
   public <PageObjectClass> PageObjectClass open(String relativeOrAbsoluteUrl,
                                                 Class<PageObjectClass> pageObjectClassClass) {
     return open(relativeOrAbsoluteUrl, "", "", "", pageObjectClassClass);
   }
 
   @CheckReturnValue
+  @Nonnull
   public <PageObjectClass> PageObjectClass open(URL absoluteUrl,
                                                 Class<PageObjectClass> pageObjectClassClass) {
     return open(absoluteUrl, "", "", "", pageObjectClassClass);
   }
 
   @CheckReturnValue
+  @Nonnull
   public <PageObjectClass> PageObjectClass open(String relativeOrAbsoluteUrl,
                                                 String domain, String login, String password,
                                                 Class<PageObjectClass> pageObjectClassClass) {
@@ -107,6 +116,7 @@ public class SelenideDriver {
   }
 
   @CheckReturnValue
+  @Nonnull
   public <PageObjectClass> PageObjectClass open(URL absoluteUrl, String domain, String login, String password,
                                                 Class<PageObjectClass> pageObjectClassClass) {
     open(absoluteUrl, domain, login, password);
@@ -114,11 +124,13 @@ public class SelenideDriver {
   }
 
   @CheckReturnValue
+  @Nonnull
   public <PageObjectClass> PageObjectClass page(Class<PageObjectClass> pageObjectClass) {
     return pageFactory.page(driver(), pageObjectClass);
   }
 
   @CheckReturnValue
+  @Nonnull
   public <PageObjectClass, T extends PageObjectClass> PageObjectClass page(T pageObject) {
     return pageFactory.page(driver(), pageObject);
   }
@@ -141,11 +153,13 @@ public class SelenideDriver {
   }
 
   @CheckReturnValue
+  @Nonnull
   public Browser browser() {
     return driver().browser();
   }
 
   @CheckReturnValue
+  @Nullable
   public SelenideProxyServer getProxy() {
     return driver().getProxy();
   }
@@ -155,11 +169,13 @@ public class SelenideDriver {
   }
 
   @CheckReturnValue
+  @Nonnull
   public WebDriver getWebDriver() {
     return driver.getWebDriver();
   }
 
   @CheckReturnValue
+  @Nonnull
   public WebDriver getAndCheckWebDriver() {
     return driver.getAndCheckWebDriver();
   }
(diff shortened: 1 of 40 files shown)


src/main/java/com/codeborne/selenide/SelenideElement.java
```diff
@@ -923,26 +995,27 @@
...
   boolean isImage();
 
   /**
    * Take screenshot of this element
-   * @return file with screenshot (*.png)
    *
+   * @return file with screenshot (*.png)
    * @see com.codeborne.selenide.commands.TakeScreenshot
    */
+  @Nonnull
```
BorisOsipov, line 1009: AFAIK it is Nullable 
see https://github.com/selenide/selenide/blob/master/src/main/java/com/codeborne/selenide/impl/ScreenShotLaboratory.java#L110

src/main/java/com/codeborne/selenide/SelenideElement.java
```diff
@@ -923,26 +995,27 @@
...
 
   /**
    * Take screenshot of this element
-   * @return buffered image with screenshot
    *
+   * @return buffered image with screenshot
    * @see com.codeborne.selenide.commands.TakeScreenshotAsImage
    */
   @CheckReturnValue
+  @Nonnull
```
BorisOsipov, line 1019: AFAIK it is Nullable. The same as for TakeScreenshot

src/main/java/com/codeborne/selenide/SelenideElement.java
```diff
@@ -106,95 +114,114 @@
...
    * Get the property value of the pseudo-element
+   *
    * @param pseudoElementName pseudo-element name of the element,
    *                          ":before", ":after", ":first-letter", ":first-line", ":selection"
-   * @param propertyName property name of the pseudo-element
+   * @param propertyName      property name of the pseudo-element
    * @return the property value or "" if the property is missing
    */
   @CheckReturnValue
+  @Nullable
```
BorisOsipov, line 204: javadoc says `@return the property value or "" if the property is missing`
is it really nullable? if it is true please fix javadoc then

src/main/java/com/codeborne/selenide/SelenideDriver.java
```diff
@@ -300,34 +336,41 @@ public boolean atBottom() {
     return executeJavaScript("return window.pageYOffset + window.innerHeight >= document.body.scrollHeight");
   }
 
+  @Nonnull
   public SelenideTargetLocator switchTo() {
     return driver().switchTo();
   }
 
   @CheckReturnValue
+  @Nullable
```
BorisOsipov, line 345: why is it nullable? Don't see in selenium client javadoc any info https://www.javadoc.io/doc/org.seleniumhq.selenium/selenium-api/3.141.59/org/openqa/selenium/WebDriver.html#getCurrentUrl--

also in w3c spec there is no info about return null
https://www.w3.org/TR/webdriver/#get-current-url

BorisOsipov:
>Checkstyle warns about SelenideDriver for some reason.
>SelenideDriver.java:32:1: Class Fan-Out Complexity is 30 (max allowed is 29). [ClassFanOutComplexity]

In fact it is valid warn - `ClassFanOutComplexity` counts all type references tokens and fails check when total count is more than threshold value.

Seems we have to add `SelenideDriver.java` to exclusion list for ClassFanOutComplexity rule or somehow change exclusion package patterns.

cc: @selenide/committers 

For more details https://checkstyle.sourceforge.io/apidocs/com/puppycrawl/tools/checkstyle/checks/metrics/ClassFanOutComplexityCheck.html

jreznot:
I've fixed checktyle config and all return value annotations mentioned in comments

asolntsev:
@jreznot Several chrome and firefox tests are failing:

```
java.lang.IllegalArgumentException: Unknown Selenide method: getAttribute
	at com.codeborne.selenide.commands.Commands.execute(Commands.java:146)
	at com.codeborne.selenide.impl.SelenideElementProxy.dispatchAndRetry(SelenideElementProxy.java:100)
	at com.codeborne.selenide.impl.SelenideElementProxy.invoke(SelenideElementProxy.java:66)
	at com.sun.proxy.$Proxy28.getAttribute(Unknown Source)
	at integration.AttributeTest.userCanFindElementByAttribute(AttributeTest.java:53)
```

rosolko:
Now tests throw such exception: 
```java
java.lang.IllegalArgumentException: object is not an instance of declaring class
        at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
        at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
        at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
        at java.base/java.lang.reflect.Method.invoke(Method.java:566)
        at com.codeborne.selenide.impl.SelenideElementProxy.dispatchAndRetry(SelenideElementProxy.java:108)
        at com.codeborne.selenide.impl.SelenideElementProxy.invoke(SelenideElementProxy.java:71)
        at com.sun.proxy.$Proxy29.getAttribute(Unknown Source)
        at com.codeborne.selenide.commands.GetSelectedValue.execute(GetSelectedValue.java:24)
        at com.codeborne.selenide.commands.GetSelectedValue.execute(GetSelectedValue.java:10)
        at com.codeborne.selenide.commands.Commands.execute(Commands.java:148)
        at com.codeborne.selenide.impl.SelenideElementProxy.dispatchAndRetry(SelenideElementProxy.java:105)
        at com.codeborne.selenide.impl.SelenideElementProxy.invoke(SelenideElementProxy.java:71)
        at com.sun.proxy.$Proxy29.getSelectedValue(Unknown Source)
        at integration.DynamicSelectsTest.waitsUntilOptionWithValueAppears(DynamicSelectsTest.java:37)

**Comment by jreznot** [comment]
Please do not check it, I am still in progress )

**Comment by jreznot** [comment]
Everything fixed now, Chrome still failed with `Caused by: java.net.ConnectException: Failed to connect to localhost/127.0.0.1:28901`

**Comment by asolntsev** [comment]
> Everything fixed now, Chrome still failed with `Caused by: java.net.ConnectException: Failed to connect to localhost/127.0.0.1:28901`

It was a flaky test: chrome crashes on Travis CI time-to-time.

**Comment by plagov** [comment]
@jreznot, thanks for your contribution to improving Kotlin support in Selenide.

I used to have such code:
```
fun getCustomerName(): String {
    return container.find("[data-testid='customer-name']").text
}
```
After upgrade to 5.12.0 it doesn't compile with:
```
Type mismatch.
Required: String
Found: String?
```
Of course, I can change the return type of the method from `String` to `String?`, but that will require to change all subsequent calls, which I don't want. Or I can do the not-null call, `find("...").text!!`, but that is not a recommended way, as far as I understand.
Can you please advise what will be the proper way to handle this?

**Comment by asolntsev** [comment]
@plagov thank you for the feedback.

@jreznot probably in this case we should change return type to `String` because (I guess) `getText()` never returns `null`.

**Comment by pavelpp** [comment]
I've never seen getText() return null.

**Comment by jreznot** [comment]
@asolntsev I've set text as nullable because `com.codeborne.selenide.commands.GetText` calls `GetSelectedText`for `select` tag and it may return null as it is stated here:
```
public class GetSelectedText implements Command<String> {
  ... 
  @Override
  public String execute(SelenideElement proxy, WebElementSource selectElement, Object[] args) {
    WebElement option = getSelectedOption.execute(proxy, selectElement, NO_ARGS);
    return option == null ? null : option.getText();
  }
}
```

**Comment by jreznot** [comment]
@pavelpp 

> Also, Selenide.executeJavaScript() returns String? now, though I'm not sure it can return null.

WebDriver JavaDoc for `org.openqa.selenium.JavascriptExecutor#executeScript`:
```
   * @return One of Boolean, Long, Double, String, List, Map or WebElement. Or null.
   */
  Object executeScript(String script, Object... args);
```

asolntsev:
1. I think we can change `GetSelectedText` to always return empty string.
2. But `executeJavascript` can return null for sure.

jreznot:
>     1. I think we can change `GetSelectedText` to always return empty string.

@asolntsev Isn't it a breaking change? In this case you cannot distinguish empty text of selected option and no selection in select.

pavelpp:
@jreznot Actually it is not possible to have no selection. It will fail much earierl. `  @Override
  public WebElement getFirstSelectedOption() {
    return getOptions().stream().filter(WebElement::isSelected).findFirst()
        .orElseThrow(() -> new NoSuchElementException("No options are selected"));
  }
`

Issue #1048, opened 2020-02-03
`$` inside PageObject should only search elements inside the SelenideDriver...


asolntsev:
in which the PageObject was created. 

### Code sample:
```java
var loginPage1 = browser1.open("/", LoginPage.class);
loginPage1.login();
```

### LoginPage.java

```java
class LoginPage() {
  public void login() {
    $("#log-me-in").click();  // this line fails because `$` refers to a "global" browser which was not opened in the test.
  }
}
```

### Suggested solution:
When used inside a page object, method `$` (and `$$`) could understand context and use the same `SelenideDriver` in which the page object was created.

asolntsev:
I guess it's not technically possible, so let's just cancel this feature request.

PR #1141, opened 2020-05-06
`$` inside PageObject should search elements inside SelenideDriver in…

jreznot:
… which PageObject was created #1048

## Proposed changes

This is very rough PoC of driver for PageObject instance.
Key changes:

- Javassist dependency to create sub-class proxies of PageObject classes
- DriverOrchestrator bridge from driver to static facade required
- Proxy can be created only for public and non-final classes
- PageObject created in context of the driver, fields set with correct driver
- On each method invocation of PageObject proxy correct driver set to main ThreadLocal

Only illustrates the general idea and should not be merged.

### DIFF
--- a/gradle/dependencies.gradle
+++ b/gradle/dependencies.gradle
@@ -14,6 +14,7 @@ dependencies {
     exclude group: 'net.bytebuddy', module: 'byte-buddy'
   }
   api('io.github.bonigarcia:webdrivermanager:3.8.1')
+  implementation('org.javassist:javassist:3.27.0-GA')
 
   compileOnly('org.checkerframework:checker:3.2.0')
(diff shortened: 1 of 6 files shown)


jreznot:
@asolntsev We are going crazy here

asolntsev:
It's not just crazy. It's lazy crazy proxy wrapper factory. :)

jreznot:
> It's not just crazy. It's lazy crazy proxy wrapper factory. :)

Jokes aside I believe that Selenide should provide convenient base class for Page classes with driver passed to constructor and with all Selenide `$` / `$$` methods in this base class. That would be easier and cleaner solution.

asolntsev:
Sorry, but it seems to be over-complicated. Let's think about a simpler solution, or leave it unsolved at all. The problem itself was not too big.

PR #1146, opened 2020-05-11
make possible to connect webdriver to already existed instance of chr…

sirdir:
## Proposed changes
Make possible to connect webdriver to already existed instance of chromuim/chrome/electron instead of creating new one (in my case it is electron application)
I want simply set it with `System.setProperty("chromeoptions.debuggerAddress", "localhost:1234");` 
 or `-Dchromeoptions.debuggerAddress=localhost:1234`

## Checklist

### DIFF
--- a/src/main/java/com/codeborne/selenide/webdriver/ChromeDriverFactory.java
+++ b/src/main/java/com/codeborne/selenide/webdriver/ChromeDriverFactory.java
@@ -71,6 +71,9 @@ class ChromeDriverFactory extends AbstractDriverFactory {
     }
     currentChromeOptions.setExperimentalOption("prefs", chromePreferences);
 
+    if (System.getProperty("chromeoptions.debuggerAddress") != null) {
+      currentChromeOptions.setExperimentalOption("debuggerAddress", System.getProperty("chromeoptions.debuggerAddress"));
+    }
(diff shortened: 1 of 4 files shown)


sirdir:
Quite a tricky thing.
Pull request is still valid, for most of the cases.

But i have separate issue with it.
I just tried go with custom `WebDriverProvider` but unfortunally, the electron application that I'm testing have not adjustable window size, at least in first login-window.
And because in selenide you explicitly adjust window size it fails in class `com.codeborne.selenide.webdriver.BrowserResizer.adjustBrowserSize()`
at 34 row
```
driver.manage().window().setSize(new org.openqa.selenium.Dimension(width, height));
```

BorisOsipov:
@sirdir 

Setting `browserSize=null` and `browserPosition=null` will resolve issue with autoresizing.

I'am not sure this changes are good. It should be done by capabilities or custom `WebDriverProvider`. If there are issue with `WebDriverProvider` let's resolve them.

sirdir:
> @sirdir
> 
> Setting `browserSize=null` and `browserPosition=null` will resolve issue with autoresizing.
> 
> I'am not sure this changes are good. It should be done by capabilities or custom `WebDriverProvider`. If there are issue with `WebDriverProvider` let's resolve them.

Thanks it worked.

BorisOsipov:
Ok close then

Issue #1055, opened 2020-02-04
Selenide takes multiple screenshots in case of chained locators


asolntsev:
## The problem
When running this line Selenide takes 4 equal screenshots:

```java
$(".unexisting").$("some").$("random").$("stuff").click();
```

But should take only one.

bereg2k:
Could it be linked to this behavior? Consider this:
```java
$("button").click();
$("div#results li").$("#result-1").shouldBe(visible, ofSeconds(20));
```
- Initially, `$("div#results li")` and `$("#result-1")` do not exist in the DOM
- After the button click, a long AJAX is happening
- `$("div#results li")` and `$("#result-1")` both appear after 10 seconds, and test passes successfully.

Here's Selenide's simple report on this case:
```
+-------------------------+----------------------------------------------------------------------+----------+----------+
|Element                  |Subject                                                               |Status    |ms.       |
+-------------------------+----------------------------------------------------------------------+----------+----------+
|open                     |http://localhost:1365/container.html                                  |PASS      |3276      |
|button                   |click()                                                               |PASS      |111       |
|div#results li           |should(exist)                                                         |FAIL      |4539      |
|div#results li           |should(exist)                                                         |FAIL      |4534      |
|div#results li           |should(exist)                                                         |FAIL      |4520      |
|div#results li           |should(exist)                                                         |PASS      |878       |
|div#results li/#result-1 |should be([visible, 20 s.])                                           |PASS      |15383     |
+-------------------------+----------------------------------------------------------------------+----------+----------+
```

Notice **FAIL** status for `"div#results li" should(exist)` steps. Here's how [AllureSelenide](https://github.com/allure-framework/allure-java/blob/master/allure-selenide/src/main/java/io/qameta/allure/selenide/AllureSelenide.java) class handles this for Allure report (notice a screenshot for the same steps):

<img width="310" alt="Screenshot 2021-03-25 at 20 15 30" src="https://user-images.githubusercontent.com/42677987/112515014-f33ed180-8da6-11eb-9ccb-5734429cfbd9.png">

**Bottom line: if it's possible to handle these `should(exist)` operations on container elements differently (not firing FAIL status under certain circumstances), then the multiple screenshots issue will be handled as well.**

bereg2k:
BTW, you can easily replicate this situation with existing Selenide's test codebase.
- Locate `src/test/resources/long_ajax_request.html`
- Increase this parameter `var commandDuration = 10000;` in the source
[...]

PR #1183, opened 2020-05-30
selenide issue #1055 fix code commit

ChrisZcu:
## Proposed changes
fix the issue where wrong number screenshot number


asolntsev:
@ChrisZcu Thank you for the PR. We value your contribution, but unfortunately this is not a correct solution.

When executing `$("#b1").$("#b2").click();`, this line:

```
if (method.getName().equals("$"))
            return Commands.getInstance().execute(null, webElementSource, method.getName(), args);
```

makes Selenide to search `"#b2` in the whole page, not only inside of `#b1`. And it can even find an element `#b2`, if it occasionally exists somewhere on the page (outside of `#b1`).

Issue #1166, opened 2020-05-22
Unexpected behavior of 'screenshot' functionality.


shubicus:
## The problem
Unexpected behavior of 'screenshot' functionality.

## Details
In the case of multithreading execution with custom browser - the shorthand screenshot approach `Selenide.screenshot(String fileName);` does not work. Due to this issue, there was found a workaround - to use 'ScreenShotLaboratory' with passing custom browser instance `ScreenShotLaboratory.getInstance().takeScreenShot(Driver driver, String fileName);`

## Solution
1. A short way -> Add method `Selenide.screenshot(Driver driver, String fileName);`
2. A long way -> Investigate and fix

## Tell us about your environment
* **Selenide Version**:  5.11.1
* **Chrome\Firefox**: Chrome 83.0.4103.61 (Official Build) (64-bit) \ Firefox 76.0.1 (64-bit)
* **Browser Driver Version**: Chrome 83.0.4103.39 \ Firefox v0.26.0
* **Selenium Version**: 3.141.59 (inside Selenide)
* **OS Version**: Windows 10 pro

## Code To Reproduce Issue:
https://github.com/shubicus/SelenideScreenshotBehaviour

asolntsev:
Well, yes: if you create your own instances of `SelenideDriver` and store them to local variables, Selenide cannot know about them. And of course, all those static Selenide methods like `$`, `$$`, `open`, `takeScreenshot` will never use your local variables. It seemed obvious for me. :)

I suggest another solution: let's add method `screenshot` to class `SelenideDriver`. 
This is consistent with other methods of `Selenide` vs `SelenideDriver`. 

This way you will be  able to use `driver.screenshot(String fileName)`

asolntsev:
Implemented by https://github.com/selenide/selenide/pull/1227

PR #1185, opened 2020-05-30
selenide issue #1166 fix code commit

ChrisZcu:
## Proposed changes
fix the issue 1166


asolntsev:
@ChrisZcu Thank you for the inspiration. 
I suggested another solution in https://github.com/selenide/selenide/issues/1166#issuecomment-659662633 and https://github.com/selenide/selenide/pull/1227

PR #1194, opened 2020-06-08
adding Item With Text collection condition

LuisOsv:
## Proposed changes
Added new collection condition to match if an item from the collection has the expected text
Example:
`$$("#user-table tbody tr  td.firstname").shouldBe(itemWithText("Bob"));`

## Checklist

### DIFF
--- a/src/test/java/com/codeborne/selenide/collections/ItemWithTextTest.java
+++ b/src/test/java/com/codeborne/selenide/collections/ItemWithTextTest.java
@@ -1,0 +1,61 @@
+package com.codeborne.selenide.collections;
+
+import com.codeborne.selenide.ElementsCollection;
+import com.codeborne.selenide.SelenideElement;
+import com.codeborne.selenide.ex.TextsMismatch;
+import com.codeborne.selenide.impl.WebElementsCollection;
+import org.assertj.core.api.WithAssertions;
+import org.junit.jupiter.api.Test;
+
+import java.util.Collections;
+
+import static com.codeborne.selenide.Mocks.mockCollection;
+import static com.codeborne.selenide.Mocks.mockElement;
+
+class ItemWithTextTest implements WithAssertions {
+  private SelenideElement element1 = mockElement("Test-One");
+  private SelenideElement element2 = mockElement("Test-Two");
+  private SelenideElement element3 = mockElement("Test-Three");
+  private WebElementsCollection collection = mockCollection("Collection description", element1, element2, element3);
+
+  @Test
+  void applyOnCorrectElementText() {
+    assertThat(new ItemWithText("Test-One")
+      .test(collection.getElements()))
+      .isTrue();
+  }
+
+  @Test
+  void applyOnWrongElementText() {
+    assertThat(new ItemWithText("Test-X")
+      .test(collection.getElements()))
+      .isFalse();
+  }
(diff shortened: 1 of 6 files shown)


src/test/java/integration/CollectionMethodsTest.java
```diff
@@ -464,4 +471,21 @@ void errorWhenSomeMatchedButNoneShould() {
...
+      .shouldBe(itemWithText("Bob"));
+  }
+
+  @Test
+  void errorWhenItemWithTextNotMatchedButShouldBe() {
+    String expectedText = "Luis";
+    assertThatThrownBy(()  -> $$("#user-table tbody tr  td.firstname").shouldBe(itemWithText(expectedText)))
+      .isInstanceOf(TextsMismatch.class)
+      .hasMessageContaining(
+        String.format(String.format("Texts mismatch" +
```
asolntsev, line 486: I think we should have another error message instead of "Texts mismatch". Something like "element with text not found".

src/test/java/com/codeborne/selenide/collections/ItemWithTextTest.java
```diff
@@ -0,0 +1,61 @@
...
+
+  @Test
+  void failOnMatcherError() {
+    String expectedText = "Won't exist";
+    assertThatThrownBy(() -> new ItemWithText(expectedText)
+      .fail(collection,
+        collection.getElements(),
+        new Exception("Exception message"), 10000)).isInstanceOf(TextsMismatch.class)
+      .hasMessageContaining(
+        String.format(String.format("Texts mismatch" +
```
asolntsev, line 57: I think we should have another error message instead of "Texts mismatch". Something like "element with text not found".

src/test/java/integration/CollectionMethodsTest.java
```diff
@@ -464,4 +471,21 @@ void errorWhenSomeMatchedButNoneShould() {
...
+  @Test
+  void shouldItemWithText() {
+    $$("#user-table tbody tr  td.firstname")
+      .shouldBe(itemWithText("Bob"));
+  }
+
+  @Test
+  void errorWhenItemWithTextNotMatchedButShouldBe() {
+    String expectedText = "Luis";
+    assertThatThrownBy(()  -> $$("#user-table tbody tr  td.firstname").shouldBe(itemWithText(expectedText)))
```
asolntsev, line 483: Please use `shouldHave(itemWithText)` instead of `shouldBe(itemWithText)`.

asolntsev:
@LuisOsv Thank you for the PR! It's good, just a few minor changes requested.

src/test/java/integration/CollectionMethodsTest.java
LuisOsv: Fixed, I have added a new exception class with "Element with text not found" as error message.

src/test/java/com/codeborne/selenide/collections/ItemWithTextTest.java
LuisOsv: fixed

src/test/java/integration/CollectionMethodsTest.java
LuisOsv: fixed

PR #1221, opened 2020-07-05
Unique downloads folder for every browser

asolntsev:
## Proposed changes
Solve issue ##1220: generate a unique downloads folder for every browser instance.  

The main change is _the downloads folder_. 
* before: `build/downloads` (for all browser instances)
* after: `build/downloads/<timestamp>_<pid>_<thread_id>"`

## API changes
* `com.codeborne.selenide.SelenideDriver` constructor gets new parameter `File browserDownloadsFolder`
* `com.codeborne.selenide.Driver` gets new method `File browserDownloadsFolder()`
* In all subclasses of `com.codeborne.selenide.webdriver.AbstractDriverFactory` method `create` gets new parameter `File browserDownloadsFolder`. Some of them ignore it (Safari and IE) because those browsers doesn't support custom downloads folder.

## Checklist

### DIFF
--- a/src/main/java/com/codeborne/selenide/SelenideDriver.java
+++ b/src/main/java/com/codeborne/selenide/SelenideDriver.java
@@ -36,7 +36,7 @@ import static java.util.Collections.emptyList;
 @ParametersAreNonnullByDefault
 public class SelenideDriver {
   private static final Navigator navigator = new Navigator();
-  private static ScreenShotLaboratory screenshots = ScreenShotLaboratory.getInstance();
+  private static final ScreenShotLaboratory screenshots = ScreenShotLaboratory.getInstance();
 
   private final Config config;
   private final Driver driver;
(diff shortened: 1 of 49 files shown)


src/main/java/com/codeborne/selenide/SelenideDriver.java
```diff
@@ -53,9 +53,9 @@ public SelenideDriver(Config config, Driver driver) {
     this.driver = driver;
   }
 
-  public SelenideDriver(Config config, WebDriver webDriver, @Nullable SelenideProxyServer selenideProxy) {
+  public SelenideDriver(Config config, WebDriver webDriver, @Nullable SelenideProxyServer selenideProxy, File browserDownloadsFolder) {
```
dstekanov, line 56: Should we have a constructor with the default browser downloads folder? 

I mean what if we brake the usage of `public SelenideDriver(Config config, WebDriver webDriver, @Nullable SelenideProxyServer selenideProxy)` constructor?

asolntsev: Right. I will restore the previous constructor too.

asolntsev: done in commit b23338445e7

Issue #1173, opened 2020-05-25
Add ability to click SelenideElement via JS/Selenium without changing Configuration.clickViaJs global parameter


artemryazantsev:
## Add ability to click SelenideElement via JS/Native(Selenium) click 
I'd like to have ability to click SelenideElement via JS/Selenium click without changing Configuration.clickViaJs global parameter.

As @asolntsev mentioned in Telegram chat it potentially can be:

> 
Да, возможно, стоит сделать. Но тогда, наверное, не методы $.click() и $.clickViaJS(), а что-нибудь более общее, типа $.click(ClickOptions options).

## Details
It is a common case when only JS clicks work for some elements and at the same moment these JS clicks don't work for other elements (e.g. react-js app in my case). And it would be great to have an ability to combine click types directly via Selenide without self-created JS/Selenium helper classes.

Current implementation doesn't provide an ability to click SelenideElement via JS/Selenium , user should select click type by setting `Configuration.clickViaJs = true/false `parameter

asolntsev:
implemented by https://github.com/selenide/selenide/pull/1226

PR #1226, opened 2020-07-15
Add new click via `ClickOptions`

dstekanov:
## Proposed changes
Feature for #1173 

## Checklist


src/main/java/com/codeborne/selenide/SelenideElement.java
```diff
@@ -954,6 +954,21 @@
...
+   */
+  void click(ClickOptions clickOption);
+
+  /**
+   * Click the element using {@link com.codeborne.selenide.ClickOptions}
+   * with a relative offset from the center of the element
+   *
+   * @see com.codeborne.selenide.commands.Click
+   */
+  void click(ClickOptions clickOption, int offsetX, int offsetY);
```
asolntsev, line 970: Ideally, we would like to avoid too many overloaded methods in class `Selenide`. 
Could we only have method `click(ClickOptions clickOption)`, and move parameters `int offsetX, int offsetY` into `ClickOptions` class?

dstekanov: Done. Changed to `click(ClickParams);`

PR #1233, opened 2020-07-28
Add new shadow root selector 'shadowRoot([shadowHost], [cssSelector])'

sakamoto66:
### I want to improve the selector of shadowCss.
- current style
shadowCss("p", "host1", "host2", "host3", "host4")
- new style
shadowRoot("host1//host4", "p")

### Support shadow-root elements
ex) shadow-root tree
```
host1
  + host2
     +p
  + host2
     +p
     +p
  + host2
     +p
```
$$(shadowRoot("host1/host2", "p")).shouldHaveSize(4);  

## Checklist

### DIFF
--- a/src/main/java/com/codeborne/selenide/selector/ByShadowRoot.java
+++ b/src/main/java/com/codeborne/selenide/selector/ByShadowRoot.java
@@ -14,6 +14,9 @@ import javax.annotation.ParametersAreNonnullByDefault;
 import java.io.Serializable;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.function.Function;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 @ParametersAreNonnullByDefault
 public class ByShadowRoot {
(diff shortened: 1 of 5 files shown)


asolntsev:
@dstekanov What do you think?

SeleniumTestAB:
I am not @dstekanov but i do see one problem here. ShadowRoot will only accept CSS when in fact Xpath is also very popular choice. I would either suggest getting another param there for type of selector, or not having that kind of thing as shadowRoot.

dstekanov:
Hi, @sakamoto66 thanks for your PR!

Initially, I thought about two cases:
`By shadowCss(String target, String shadowHost, String... innerShadowHosts)` - current implementation
`By shadowCss(String target, String shadowHost)` - similar to your proposal

But I decided to leave the first one because in most cases (my own analysis) there is one shadow host or one root shadow host.

Also if we introduce `shadowRoot("host1//host4", "p")` I am not quite sure if we will be able to deprecate `shadowCss(...)` because a lot of users can already use it. And having two methods seems a not so good idea...

src/main/java/com/codeborne/selenide/selector/ByShadowRoot.java
```diff
@@ -0,0 +1,186 @@
...
+      } else {
+        jsExecutor = (JavascriptExecutor) ((WrapsDriver) context).getWrappedDriver();
+      }
+      return jsExecutor;
+    }
+
+    @SuppressWarnings("unchecked")
+    private List<WebElement> getShadowRoots(final SearchContext host, String sh, final int limit) {
+      /*
+      console.info((function(){
```
dstekanov, line 126: personally, it could be hard to maintain this JS snippet in the future

sakamoto66: Thanks for your comment. @dstekanov
> But I decided to leave the first one because in most cases (my own analysis) there is one shadow host or one root shadow host.

I also think that is ideal.
However, I think that it may not be the case in reality.
It expects to build a deep shadow-root unknowingly when using vue.js etc.
(I would like to end the speculation story here...)

> Also if we introduce shadowRoot("host1//host4", "p") I am not quite sure if we will be able to deprecate shadowCss(...) because a lot of users can already use it. And having two methods seems a not so good idea...

OK!
If so, shadowCss has some problems.
I will raise the issue of the problem.

sakamoto66:
I'm late. The issue has been updated.
#1246

asolntsev:
@sakamoto66 Probably this problem was fixed by https://github.com/selenide/selenide/pull/1373 ?

asolntsev:
It seems that this PR was superseded by #1373. 

@sakamoto66 Anyway, thank you for your PR. We value your contribution. 
Sorry for our slow reaction and long thinking process.

Issue #1220, opened 2020-07-05
Selenide could generate a unique downloads folder for every browser instance


asolntsev:
## The problem
Currently all browsers opened by Selenide downloads files to the same folder (`Configuration.downloadsFolder`, by default 
equals to `build/downloads`).

It may cause conflicts when running tests in parallel.  
It would be better if each browser had a unique downloads folder. 

## Tell us about your environment
* **Selenide Version**: 5.12.2

PR #1238, opened 2020-08-01
#1220 delete empty downloads folders

asolntsev:
... to avoid creating tons of empty subfolders in "build/downloads"

## Checklist


rosolko:
It looks like we can easily remove standard system download folder. 
I'm not 100% sure about such behavior in general.

BorisOsipov:
@rosolko good catch...

asolntsev:
> It looks like we can easily remove standard system download folder.
Good catch, yes!
But only if it's empty (which is almost never empty)

> I'm not 100% sure about such behaviour in general.
Well, yes, my idea was to delete only folders created by Selenide itself. 
Probably I should change the implementation so that Selenide would track all created folders and delete them in the end (if empty).

asolntsev:
Ok, I replaced this PR by https://github.com/selenide/selenide/pull/1247

PR #1249, opened 2020-08-09
#1220 cleanup downloads folder before every `$.download()` call

asolntsev:
NB! We can do it only for folders created by Selenide (aka "unique"). In case of folders provided by user, we cannot cleanup them.


asolntsev:
Probably will be overseeded by https://github.com/selenide/selenide/pull/1252

Issue #1051, opened 2020-02-04
Implement plugin system for Selenide


asolntsev:
## The problem
Currently Selenide allows to override some parts of it and customize its behaviour. But probably this mechanism is not ideal. 

## Details
I don't have exact plans yet. This issue is for public discussion.

START BRAINSTORMING!

SeleniumTestAB:
That would be good if selenide was like allure if we speak about runtime, unless you mean plugins like pull requests from us and that selenide will know to use it. Plugins would be nice to organize. Or even better approach, make some base dependency like selenide-core and then add another ones?

rosolko:
I think good start pint will be:

- core
- commands
- drivers
- conditions

SeleniumTestAB:
> 
> 
> I think good start pint will be:
> 
>     * core
> 
>     * commands
> 
>     * drivers
> 
>     * conditions

I would add something like frameworks there so that ppl can make some kind of adaptors. It might give some ease of configuration if done right. Also what about static part of selenide? which is one of its trademark points?

PR #1264, opened 2020-09-07
Plugins system

asolntsev:
Implementation of #1051

## What problem does it solves:
* I have another implementation of `WebElementPrinter` in `selenide-appium` project which solves [issue with null tag names in Appium](https://github.com/selenide/selenide-appium/issues/38). See https://github.com/selenide/selenide-appium/pull/41

## What I am not sure about:
- The design of `Plugins`: is it ok to have all static methods? (though, I don't see any other options)
- The design of `Plugins`: it it ok to have links to all possible Selenide plugins (it's kind of bottleneck then)
- The name `WebElementPrinter`: it doesn't print anything. Something like "information provider", "describer" etc. would be better

### DIFF
--- a/src/main/java/com/codeborne/selenide/impl/Plugins.java
+++ b/src/main/java/com/codeborne/selenide/impl/Plugins.java
@@ -23,6 +23,10 @@ public class Plugins {
     return getPlugin(ElementDescriber.class, SelenideElementDescriber.class);
   }
 
+  public static Photographer getPhotographer() {
+    return getPlugin(Photographer.class, WebdriverPhotographer.class);
+  }
+
   @SuppressWarnings("unchecked")
   private static <T> T getPlugin(Class<T> klass, Class<? extends T> defaultImplementation) {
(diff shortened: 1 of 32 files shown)


dstekanov (approved):

rosolko (approved):

src/main/java/com/codeborne/selenide/impl/Plugins.java
```diff
@@ -0,0 +1,68 @@
...
+    String resource = "/META-INF/defaultservices/" + klass.getName();
+    URL file = Plugins.class.getResource("/META-INF/defaultservices/" + klass.getName());
+    if (file == null) {
+      throw new IllegalStateException("Resource not found in classpath: " + resource);
+    }
+
+    String className = readFile(file).trim();
+    try {
+      //noinspection unchecked
+      return (T) Class.forName(className).getConstructor().newInstance();
```
vlsi, line 53: Class.forName(className).asSubclass(T).getConstructor().newInstance();

asolntsev: thanks, refactored.

PR #1286, opened 2020-10-13
Add new collection condition - exactTextsCaseSensitiveInAnyOrder

plagov:
## Proposed changes
Add new collection condition `exactTextsCaseSensitiveInAnyOrder`. There's already a condition to check that a given elements collection contains expected texts in any order. But it doesn't check for the exact text and case. The proposed condition uses `equalsCaseSensitive` comparison.

## Checklist

### DIFF
--- a/src/main/java/com/codeborne/selenide/collections/ExactTextsCaseSensitiveInAnyOrder.java
+++ b/src/main/java/com/codeborne/selenide/collections/ExactTextsCaseSensitiveInAnyOrder.java
@@ -5,6 +5,7 @@ import com.codeborne.selenide.impl.Html;
 import org.openqa.selenium.WebElement;
 
 import java.util.List;
+import java.util.stream.Collectors;
 
 public class ExactTextsCaseSensitiveInAnyOrder extends ExactTexts {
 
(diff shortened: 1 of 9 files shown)


src/main/java/com/codeborne/selenide/collections/ExactTextsCaseSensitiveInAnyOrder.java
```diff
@@ -0,0 +1,43 @@
...
+  @Override
+  public boolean test(List<WebElement> elements) {
+    if (elements.size() != expectedTexts.size()) {
+      return false;
+    }
+
+    for (String expectedText : expectedTexts) {
+      boolean found = false;
+      for (WebElement element : elements) {
+        if (Html.text.equalsCaseSensitive(element.getText(), expectedText)) {
```
asolntsev, line 28: This double-iteration may be ineffective. Every call to `WebDriver.getText()` is slow. 

For better performance, we should first collect all texts to a `List<String>`, and then iterate this list.

asolntsev (changes requested):
@plagov Thank you for the PR! It's almost good. :) 

1. Performance issue
2. It would be good to add an integration test (similar to https://github.com/selenide/selenide/blob/master/statics/src/test/java/integration/ElementTextTest.java)

src/main/java/com/codeborne/selenide/collections/ExactTextsCaseSensitiveInAnyOrder.java
plagov: 1. Updated. Got texts from all elements first and then iterated over a list of Strings. 
Since `TextsInAnyOrder` condition is using the same approach, applied the same change there as well.

2. Added several integration tests.

asolntsev (approved):

Issue #1296, opened 2020-10-30
Feature: set active SelenideDriver


sanchek:
Create method for setting currenly active SelenideDriver with it's own SelenideConfig

## The problem
If user wants to open two browsers in test, he need to point browser explicitly.
For example (https://selenide.org/2018/10/10/selenide-5.0.0/): 
```
SelenideDriver browser1 = new SelenideDriver();
SelenideDriver browser2 = new SelenideDriver();

browser1.open("https://google.com");
browser2.open("http://yandex.ru");

browser1.$(h1).shouldHave(text("Google"));
browser2.$(h1).shouldHave(text("Yandeks"));
```

## Solution can might look like this:
```
    SelenideConfig browserConfig1 = new SelenideConfig()
      .baseUrl("http://google.ru");
    SelenideDriver selenideDriver1 = new SelenideDriver(browserConfig1);
    WebDriverRunner.setActiveSelenideDriver(selenideDriver1);
    Selenide.open("/signin");
    assertThat(selenideDriver1.url()).isEqualTo("http://google.ru/signin");
    
    SelenideConfig browserConfig2 = new SelenideConfig()
      .baseUrl("http://ya.ru");
    SelenideDriver selenideDriver2 = new SelenideDriver(browserConfig2);
    WebDriverRunner.setActiveSelenideDriver(selenideDriver2);
    Selenide.open("/signin");
    assertThat(selenideDriver2.url()).isEqualTo("http://ya.ru/signin");
```

asolntsev:
@sanchek While I may be mistaken, it seems to me an over-complication. 

If you really need to use different `SelenideDriver` instances in tests, why not just instantiate them and store into test field? It's much more simple and predictable than playing with static methods. 

```java
public class MyJUnitTest {
  private SelenideDriver web = new SelenideDriver();

  @Test 
  public void someTest() {
    web.open("https://google.com");
    web.$(h1).shouldHave(text("Google"));
  }
```

sanchek:
To me changeing currect active driver looks like more simple and reusable solution.

First of all, you can use and adopt any previously written testcases without any meaningful modifiactions.

Second, whithout changing webdriver you wont be able to use PageObject pattern without calling webdriver instance, and also PageObject classes should me modified.
Example:
![image](https://user-images.githubusercontent.com/6098833/97962742-9cc0fe80-1dd7-11eb-8deb-b3ff8f6fc079.png)
[...]

PR #1297, opened 2020-10-30
Feature: set active SelenideDriver

sanchek:
https://github.com/selenide/selenide/issues/1296

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Changelog
 
 ## 5.16.0 (planned to 2x.10.2020)
+* #1296 Add ability to set active SelenideDriver with own SelenideConfig
 * #1051 Selenide plugins system
 * #1261 Add actual own text to error message (when one of checks `ownText`, `exactOwnText` fails)
 
--- a/statics/src/main/java/com/codeborne/selenide/WebDriverRunner.java
+++ b/statics/src/main/java/com/codeborne/selenide/WebDriverRunner.java
@@ -18,7 +18,7 @@ import static com.codeborne.selenide.Configuration.headless;
  */
 public class WebDriverRunner implements Browsers {
   public static WebDriverContainer webdriverContainer = new WebDriverThreadLocalContainer();
-  private static final SelenideDriver staticSelenideDriver = new ThreadLocalSelenideDriver();
+  private static SelenideDriver staticSelenideDriver = new ThreadLocalSelenideDriver();
 
   /**
    * Use this method BEFORE opening a browser to add custom event listeners to webdriver.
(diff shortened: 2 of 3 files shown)


Tonicku:
Hi, i am interested about this feature and looking forward to use it 👍

asolntsev:
@Tonicku Could you please describe the context: why do you need this feature? I still think it should not be needed.

Tonicku:
> @Tonicku Could you please describe the context: why do you need this feature? I still think it should not be needed.

Hi Andrei, yeaaa i somehow sort out what i want to achieve.

Selenide 
- Configuration is not ThreadLocal
- WebDriver is theadlocal inside WebDriverRunner

Challenge
- setup selenide.configuration for specific tests set : PageLoadStrategy, hub url, browser, environment, local vs remote, 

Solution i put in place

TESTNG XML
- as i am using Testng xml to group the test by their configuration ( eager, normal )
<?xml version="1.0" encoding="UTF-8"?>
<!DOCTYPE suite SYSTEM "https://testng.org/testng-1.0.dtd">
<suite name="Test Suite" verbose="10" parallel="classes" thread-count="1" time-out="11080000">

    <parameter name="TYPE" value="local"/>
    <parameter name="HUB" value="http://mgs-qa-k8s-harbor201.be.monster.com:32444/wd/hub"/>
    <parameter name="ENV" value="TEST_ENV"/>

    <test name=" tests PageLoadStrategy EAGER">

        <parameter name="PAGE_LOAD_STRATEGY" value="eager"/>
       <parameter name="BROWSER" value="chrome"/>

        <packages>
            <package name="${path/package/*}"/>
        </packages>

    </test>

    <test name=" tests PageLoadStrategy NORMAL">

        <parameter name="PAGE_LOAD_STRATEGY" value="normal"/>
       <parameter name="BROWSER" value="chrome"/>

        <packages>
            <package name="${path/package/*}"/>
        </packages>

    </test>

    <test name=" tests PageLoadStrategy NORMAL">

        <parameter name="PAGE_LOAD_STRATEGY" value="normal"/>
       <parameter name="BROWSER" value="edge"/>

        <packages>
            <package name="${path/package/*}"/>
        </packages>

    </test>

</suite>

UITestBase
usage of the @Test testng annotation
reusing the parameters from the testng xml, to set the configuration for these test sets

Let me know if it 's accurate solution

asolntsev:
> * setup selenide.configuration for specific tests set : PageLoadStrategy, hub url, browser, environment, local vs remote,

Hi @Tonicku !
Why do you need to use different PageLoadStrategies or hubs in parallel tests?
I don't think it's really needed.

I think a better idea is to create different profiles (in Maven) or tasks (in Gradle) or test suites (in TestNG).
See for example https://github.com/selenide-examples/google/blob/master/pom.xml

Tonicku:
@asolntsev Thanks for your reply. we have few tests with images to be tested and without few. When we know they are no images to check : we want PageLoadStrategies to be set to eager, if not normal.
Thanks for sharing how to put in place the usage of Profile.

PR #1300, opened 2020-11-01
Improve toString for AND condition

fokinp:
## Proposed changes
Print all conditions that are combined with AND condition

## Checklist

### DIFF
--- a/src/test/java/com/codeborne/selenide/conditions/AndTest.java
+++ b/src/test/java/com/codeborne/selenide/conditions/AndTest.java
@@ -1,0 +1,15 @@
+package com.codeborne.selenide.conditions;
+
+import static com.codeborne.selenide.Condition.attribute;
+import static java.util.Arrays.asList;
+import static org.assertj.core.api.Assertions.assertThat;
+
+import org.junit.jupiter.api.Test;
+
+final class AndTest {
+  @Test
+  void verifyToString() {
+    And and = new And("checked", asList(attribute("checked", "true"), attribute("checked", "on")));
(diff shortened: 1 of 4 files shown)


src/test/java/com/codeborne/selenide/conditions/AndTest.java
```diff
@@ -0,0 +1,15 @@
+package com.codeborne.selenide.conditions;
+
+import static com.codeborne.selenide.Condition.attribute;
+import static java.util.Arrays.asList;
+import static org.assertj.core.api.Assertions.assertThat;
+
+import org.junit.jupiter.api.Test;
+
+final class AndTest {
```
fokinp, line 9: Do we need a separate Test class since we already had appropriate tests in `ConditionTest.java` ?

asolntsev: Good question. No, we don't need to duplicate test in `ConditionTest.java` and `AndTest.java`. It would be great to remove the duplication and leave only one of those tests (probably `AndTest.java`?)

asolntsev (approved):
Great, thank you!

src/test/java/com/codeborne/selenide/conditions/AndTest.java
fokinp: @asolntsev yes, I suppose be better to leave single test class and `AndTest.java` is the best candidate. But this work should be done in scope of another PR since `Or` condition tests can be refactored as well. So probably I'll prepare PR with this refactoring a bit later

asolntsev: Agree.

rosolko (approved):

Issue #1307, opened 2020-11-08
When downloading multiple files within a single request - Chrome security dialog can block the text execution


vinogradoff:
Sample code:

```java
@Test
    void downloadMultipleFiles() throws FileNotFoundException {
        Configuration.proxyEnabled = true;
        open("https://jsfiddle.net/vngr/b2s74tjp/10/");
        switchTo().frame("result");
        File text = $("#two-downloads").download(using(PROXY)
                .withTimeout(6000).withFilter(FileFilters.withName("hello_world.txt")));
        Assertions.assertEquals("hello_world.txt", text.getName());
        Assertions.assertEquals(6, text.length());

    }
```

PR #1308, opened 2020-11-08
Fix/1307 multiple download

vinogradoff:
## Proposed changes
This is a fix for #1307 which allows download multiple file by chrome as default setting

## Checklist

### DIFF
--- a/src/test/java/com/codeborne/selenide/webdriver/ChromeDriverFactoryTest.java
+++ b/src/test/java/com/codeborne/selenide/webdriver/ChromeDriverFactoryTest.java
@@ -44,9 +44,10 @@ final class ChromeDriverFactoryTest implements WithAssertions {
     Capabilities chromeOptions = factory.createCapabilities(config, browser, proxy, browserDownloadsFolder);
     Map<String, Object> prefsMap = getBrowserLaunchPrefs(ChromeOptions.CAPABILITY, chromeOptions);
 
-    assertThat(prefsMap).hasSize(3);
+    assertThat(prefsMap).hasSizeGreaterThanOrEqualTo(4);
     assertThat(prefsMap).containsEntry("credentials_enable_service", false);
     assertThat(prefsMap).containsEntry("plugins.always_open_pdf_externally", true);
+    assertThat(prefsMap).containsEntry("profile.default_content_setting_values.automatic_downloads", 1);
     assertThat(prefsMap).containsEntry("download.default_directory",
       new File(DOWNLOADS_FOLDER).getAbsolutePath());
   }
--- a/statics/src/test/java/integration/proxy/MultipleDownloadsTest.java
+++ b/statics/src/test/java/integration/proxy/MultipleDownloadsTest.java
@@ -17,7 +17,6 @@ import static org.junit.jupiter.api.Assertions.assertEquals;
 public class MultipleDownloadsTest extends IntegrationTest {
   @BeforeEach
   void setUp() {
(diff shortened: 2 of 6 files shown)


src/test/java/com/codeborne/selenide/webdriver/ChromeDriverFactoryTest.java
```diff
@@ -44,9 +44,7 @@ void defaultChromeOptions() {
     Capabilities chromeOptions = factory.createCapabilities(config, browser, proxy, browserDownloadsFolder);
     Map<String, Object> prefsMap = getBrowserLaunchPrefs(ChromeOptions.CAPABILITY, chromeOptions);
 
-    assertThat(prefsMap).hasSize(3);
-    assertThat(prefsMap).containsEntry("credentials_enable_service", false);
-    assertThat(prefsMap).containsEntry("plugins.always_open_pdf_externally", true);
+    assertThat(prefsMap).hasSizeGreaterThan(0);
```
asolntsev, line 47: I don't think it's a good change. 
The name of test is `defaultChromeOptions()`, and its intent is exactly to test those default options.

asolntsev (changes requested):

statics/src/test/java/integration/proxy/MultipleDownloadsTest.java
```diff
@@ -0,0 +1,27 @@
...
+import java.io.*;
+
+import static com.codeborne.selenide.DownloadOptions.using;
+import static com.codeborne.selenide.FileDownloadMode.PROXY;
+import static com.codeborne.selenide.Selenide.*;
+
+public class MultipleDownloadsTest extends IntegrationTest {
+  @BeforeEach
+  void setUp() {
+    closeWebDriver();
```
asolntsev, line 16: It's not needed to call `closeWebDriver();`: method `useProxy(true);` does it automatically if needed.

src/test/java/com/codeborne/selenide/webdriver/ChromeDriverFactoryTest.java
vinogradoff: I disagree. Unit tests should test behaviour and not the implementation details. Testing on implementation details is considered to be an anti-pattern, because it slows down the refactoring and just re-assuring that the concrete code lines exist in the docs.

We could argue for deleting this test all together, but these checks are really binding to the implementation.

vinogradoff: That change could be interpreted: we assert that some preferences are actually set, and that one concrete folder is setup for a certain property - this is at least behavour-ish, imho.

statics/src/test/java/integration/proxy/MultipleDownloadsTest.java
vinogradoff: fixed

src/test/java/com/codeborne/selenide/webdriver/ChromeDriverFactoryTest.java
asolntsev: Yes, unit tests should test behaviour and not the implementation details. 
But in this case, it's _the behaviour_. Method name is `createCapabilities`, test name is `defaultChromeOptions` - it's all about generating those capabilities. This test tests exactly that.

asolntsev (changes requested):

asolntsev (approved):

Issue #217, opened 2015-09-06
When test fails, Selenide should report **exact** actual value at the moment of failure


asolntsev:
Now it's possible that condition is not met at the moment when Selenide checks it. But at the moment of reporting error condition can be met. For example, this error message is misleading:

``` java
com.codeborne.selenide.ex.ElementShould: Element should have textCaseSensitive '09000525 FD: 1999-11-08' {.timeline-event-bubble-title}
Element: '<div class="timeline-event-bubble-title">09000525 FD: 1999-11-08</div>'
    at com.codeborne.selenide.impl.AbstractSelenideElement.checkCondition(AbstractSelenideElement.java:512)
    at com.codeborne.selenide.impl.AbstractSelenideElement.should(AbstractSelenideElement.java:476)
    at com.codeborne.selenide.impl.AbstractSelenideElement.invokeShould(AbstractSelenideElement.java:299)
    at com.codeborne.selenide.impl.AbstractSelenideElement.dispatchSelenideMethod(AbstractSelenideElement.java:179)
    at com.codeborne.selenide.impl.AbstractSelenideElement.dispatchAndRetry(AbstractSelenideElement.java:92)
    at com.codeborne.selenide.impl.AbstractSelenideElement.invoke(AbstractSelenideElement.java:67)
    at com.codeborne.selenide.impl.WaitingSelenideElement.invoke(WaitingSelenideElement.java:18)
I think that Selenide can be more "smart" and remember the stringified DOM element each time before checking the condition, so that when condition fails the error message is not contradictory.
```

PR #1313, opened 2020-11-11
Refactor text conditions

fokinp:
## Proposed changes

1. Move common text conditions logic to base class
   
- save/cache actual value/text that is used for compassing and for debug purpose in case of failure
- simplify TextCondition extension

2. Refactor unit test related to text conditions

## Checklist


asolntsev:
@fokinp Please fix the tests

asolntsev (changes requested):
It's a great initiative to refactor the duplicated code!

But there is a problem with storing actual value in `Condition`: it's not thread-safe. People can store a single `Condition` instance, say, to a static variable, and use it in multiple tests, and even in multiple parallel threads.

fokinp:
> But there is a problem with storing actual value in `Condition`: it's not thread-safe. People can store a single `Condition` instance, say, to a static variable, and use it in multiple tests, and even in multiple parallel threads.

@asolntsev that is very strange way to use conditions, but I got the point anyway :).
In this case I see two ways:

1.  do not cache actual value and just call `getText` second time for `actualValue`
2.  or use `ThreadLocal` storage for actual value per thread

If we go with 1st option:
 * potentially can be a case when value/text be changed after `apply` call (e.g. https://github.com/selenide/selenide/issues/217) and this behavior can confuse someone who will read/analyze test run
 * we unnecessary do an additional browser interaction (very minor issue)
 * we need do refactoring of: `SelectedText`

And small additional point: will be nice do not check comparing logic in condition tests (like we currently do) and cover this in `HtmlTest` and check only that we call correct method from `Html` class and pass correct values to it. What do you think?

asolntsev:
@fokinp Yes, #1 is not good (at least because of https://github.com/selenide/selenide/issues/217). Currently I don't see any other solutions except #2 (`ThreadLocal `). Let's do it.

src/main/java/com/codeborne/selenide/conditions/TextCondition.java
```diff
@@ -0,0 +1,42 @@
...
+import com.codeborne.selenide.Driver;
+import org.openqa.selenium.WebElement;
+
+import javax.annotation.Nullable;
+import javax.annotation.ParametersAreNonnullByDefault;
+
+@ParametersAreNonnullByDefault
+public abstract class TextCondition extends Condition {
+  private final String expectedText;
+  private final ThreadLocal<String> actualText = new ThreadLocal<>();
```
fokinp, line 13: @asolntsev jfyi, I get `"ThreadLocal" variables should be cleaned up when no longer used` sonar rule violation locally. I don't know exactly it is important for us or no

asolntsev:
because of merge conflicts, the work continues in a new PR https://github.com/selenide/selenide/pull/1566

@fokinp Thank you for the PR and sorry for so long delay. :(

PR #1319, opened 2020-11-15
Setup github actions

BorisOsipov:
## Proposed changes

Lost features in comparison with travis
- gradle cache
- Sonar integration
- slack notifications


asolntsev (approved):

rosolko:
@BorisOsipov Gradle cache an be setuped using `actions/cache@v2` action:
```
      - name: Setup cache
        uses: actions/cache@v2
        with:
          path: ~/.gradle/caches
          key: ${{ runner.os }}-gradle-${{ hashFiles('**/*.gradle') }}
          restore-keys: ${{ runner.os }}-gradle
```

rosolko:
@BorisOsipov Also, instead of sonar we can use github's `codeql` analysis. Documentation [reference](https://docs.github.com/en/free-pro-team@latest/github/finding-security-vulnerabilities-and-errors-in-your-code/configuring-code-scanning)

BorisOsipov:
For `codeql` analysis it seems work and for example looks like this https://github.com/BorisOsipov/selenide/pull/8/files#diff-5185a242c449ac7a9b97464b3093761343ca6b5e6592edc34bfe6c605d8ff8e9

rosolko (approved):
LGTM!

PR #1334, opened 2020-11-25
make SoftAssertsExtension thread-safe

dtuchs:
## Proposed changes
Fix [issue from Allure tg](https://t.me/allure_ru/7588)  
SoftAssertsExtension should use thread local ErrorsCollector

## Checklist
- [x] Checkstyle and unit tests are passed locally with my changes by running `gradlew check chrome_headless firefox_headless` command
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have added necessary documentation (if appropriate)

PR #1335, opened 2020-11-25
simplify SoftAssertsExtension

asolntsev:
...updated by https://github.com/selenide/selenide/pull/1334

## Proposed changes

## Checklist

### DIFF
--- a/src/main/java/com/codeborne/selenide/junit5/SoftAssertsExtension.java
+++ b/src/main/java/com/codeborne/selenide/junit5/SoftAssertsExtension.java
@@ -1,7 +1,6 @@
 package com.codeborne.selenide.junit5;
 
 import com.codeborne.selenide.logevents.ErrorsCollector;
-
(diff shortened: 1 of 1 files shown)


asolntsev:
@dtuchs Isn't this solution even more simple? 
(similar to https://github.com/selenide/selenide/blob/master/src/main/java/com/codeborne/selenide/junit/SoftAsserts.java)

rosolko (approved):

dtuchs:
@asolntsev Sorry, i haven't seen this PR before I did it.

Of course your solution more simply, but for my opinion that's not exactly junit5 way.
In case that Object (Resource) should be bounded to test lifecycle in Extension, i just thought it would be nice to have access to this Object from another Extensions (and yes, I know about YAGNI :)).  
Public Namespaces & storing in ExtensionContext will provide it.

asolntsev:
This PR didn't exist :)
I created it after your PR. 

I agree that it's better to use "JUnit way".

PR #1341, opened 2020-12-03
Fix tests

BorisOsipov:


asolntsev (approved):
Great, thank you!

FYI I started extracting TestNG support to a separate plugin: https://github.com/selenide/selenide-testng

Once it's done, we can upgrade to latest AssertJ again.

PR #1348, opened 2020-12-09
split the single gradle project to subprojects

asolntsev:
+ extract `junit4` subproject
+ extract `testng` subproject
+ extract `grid` subproject

## Proposed changes
extract JUnit4-related code to `modules/junit4` and TestNG-related code to `modules/testng`. 
The goal is to avoid mixing junit.jar and testng.jar in common classpath. For example, AssertJ `assumeThat` works incorrectly if encounters both jars in classpath.


gradle.properties
```diff
@@ -1,3 +1,7 @@
 org.gradle.parallel=true
 org.gradle.configureondemand=true
 org.gradle.jvmargs=-Xmx256m -XX:+HeapDumpOnOutOfMemoryError -Dfile.encoding=UTF-8
+
```
rosolko, line 4: If you're added `org.gradle.caching=true` can you please also add `org.gradle.vfs.watch=true` configuration?

gradle/dependencies.gradle
```diff
@@ -25,45 +34,36 @@ dependencies {
...
+    }
+    implementation('org.slf4j:slf4j-api:1.7.30')
+    testRuntimeOnly('org.slf4j:slf4j-simple:1.7.30')
+  }
 }
 
-task libsTest(type: Sync) {
-  from configurations.testRuntimeClasspath - configurations.compileClasspath
-  into "$buildDir/lib-test"
+subprojects {
```
rosolko, line 63: Maybe we can use or just think about it - [composite builds](https://docs.gradle.org/current/userguide/composite_builds.html)

rosolko (approved):

rosolko (commented):
Leave some comments, check them please.

gradle.properties
asolntsev: done

gradle/dependencies.gradle
asolntsev: yes, I will try composite builds.

PR #1355, opened 2020-12-29
Commands should return SelenideElement type instead of WebElement

BorisOsipov:
Useful when you want things like this:

```
class ScrollToCenter extends ScrollIntoView {
    @Override
    public SelenideElement execute(SelenideElement proxy, WebElementSource locator, Object[] args) {
        return super.execute(proxy, locator, new Object[] {"{block: \"center\"}"});
    }
}
```
...

```
$(".lupa").execute(new ScrollToCenter()).$(".pupa").click()
```


asolntsev (approved):
LGTM!

Issue #1356, opened 2020-12-30
Broken .setValue(null) in #1138


dzem:
## The problem
In selenide 5.17.0 `.setValue(null)` is fail with NPE.

## Details
Reason is call arguments.ofType(Duration.class); with arguments contains array with null value
`java.lang.NullPointerException
	at com.codeborne.selenide.impl.Arguments.ofType(Arguments.java:38)
	at com.codeborne.selenide.impl.SelenideElementProxy.getTimeoutMs(SelenideElementProxy.java:169)
	at com.codeborne.selenide.impl.SelenideElementProxy.invoke(SelenideElementProxy.java:75)`

## Tell us about your environment
* **Selenide Version**: 5.17.0
* **Chrome\Firefox\IE Version**:  87.0.4280.88
* **OS Version**: Win

## Code To Reproduce Issue
`$().setValue(null);`

asolntsev:
Fixed in https://github.com/selenide/selenide/pull/1357

PR #1357, opened 2020-12-30
#1356 fix set value null

dzem:
Fix failing .setValue(null)

## Proposed changes
Fix for https://github.com/selenide/selenide/issues/1356

## Checklist
- [✓] Checkstyle and unit tests are passed locally with my changes by running `gradlew check chrome_headless firefox_headless` command
- [✓] I have added tests that prove my fix is effective or that my feature works

### DIFF
--- a/src/main/java/com/codeborne/selenide/impl/Arguments.java
+++ b/src/main/java/com/codeborne/selenide/impl/Arguments.java
@@ -4,6 +4,7 @@ import javax.annotation.CheckReturnValue;
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
 import java.util.Arrays;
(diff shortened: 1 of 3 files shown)


BorisOsipov:
@dzem Thanks for reporting this issue and proposed fix.

It will be cool if fix covers similar cases:
```
  @Test
  public void foo() {
    open("https://www.jetbrains.com/search/");
    $("[data-test=search-field]").pseudo(":before", null);// still fail with weird NPE in Arguments class.
  }
```

dzem:
I changed fix implementation, covered `pseudo(,null)` too.
@BorisOsipov Please take a look

BorisOsipov (approved):
LGTM

PR #1370, opened 2021-01-15
Fix not and or

asolntsev:
## Proposed changes

## Checklist

### DIFF
--- a/src/main/java/com/codeborne/selenide/Condition.java
+++ b/src/main/java/com/codeborne/selenide/Condition.java
@@ -39,7 +39,7 @@ import javax.annotation.Nullable;
 import javax.annotation.ParametersAreNonnullByDefault;
 import java.util.function.Predicate;
 
-import static java.util.Arrays.asList;
+import static com.codeborne.selenide.conditions.ConditionHelpers.merge;
 
 /**
  * Conditions to match web elements: checks for visibility, text etc.
@@ -504,28 +504,34 @@ public abstract class Condition {
 
   /**
    * Check if element matches ALL given conditions.
+   * The method signature makes you to pass at least 2 conditions, otherwise it would be nonsense.
    *
    * @param name       Name of this condition, like "empty" (meaning e.g. empty text AND empty value).
-   * @param conditions Conditions to match.
+   * @param condition1 first condition to match
+   * @param condition2 second condition to match
+   * @param conditions Other conditions to match
    * @return logical AND for given conditions.
    */
   @CheckReturnValue
   @Nonnull
-  public static Condition and(String name, Condition... conditions) {
-    return new And(name, asList(conditions));
+  public static Condition and(String name, Condition condition1, Condition condition2, Condition... conditions) {
+    return new And(name, merge(condition1, condition2, conditions));
   }
 
   /**
    * Check if element matches ANY of given conditions.
+   * The method signature makes you to pass at least 2 conditions, otherwise it would be nonsense.
    *
    * @param name       Name of this condition, like "error" (meaning e.g. "error" OR "failed").
-   * @param conditions Conditions to match.
+   * @param condition1 first condition to match
+   * @param condition2 second condition to match
+   * @param conditions Other conditions to match
    * @return logical OR for given conditions.
    */
   @CheckReturnValue
   @Nonnull
-  public static Condition or(String name, Condition... conditions) {
-    return new Or(name, asList(conditions));
+  public static Condition or(String name, Condition condition1, Condition condition2, Condition... conditions) {
+    return new Or(name, merge(condition1, condition2, conditions));
   }
 
   /**
--- a/src/main/java/com/codeborne/selenide/conditions/Or.java
+++ b/src/main/java/com/codeborne/selenide/conditions/Or.java
@@ -4,9 +4,11 @@ import com.codeborne.selenide.Condition;
 import com.codeborne.selenide.Driver;
 import org.openqa.selenium.WebElement;
 
+import javax.annotation.Nonnull;
 import javax.annotation.ParametersAreNonnullByDefault;
 import java.util.List;
 
+import static com.codeborne.selenide.conditions.ConditionHelpers.negateMissingElementTolerance;
 import static java.util.stream.Collectors.joining;
 
 @ParametersAreNonnullByDefault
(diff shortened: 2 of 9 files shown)


src/main/java/com/codeborne/selenide/conditions/Or.java
```diff
@@ -4,19 +4,28 @@
...
+import static com.codeborne.selenide.conditions.ConditionHelpers.negateMissingElementTolerance;
 import static java.util.stream.Collectors.joining;
 
 @ParametersAreNonnullByDefault
 public class Or extends Condition {
 
   private final List<Condition> conditions;
 
-  public Or(String name, List<Condition> conditions) {
+  public Or(String name, Condition condition1, Condition condition2, Condition... conditions) {
```
fokinp, line 20: @asolntsev can we additionally keep previous signature too and just check argument amount? 
This change will break backward compatibility and sometimes more convenient to use `List`.
e.g.
```java
var expectedOptions = List.of("one", "two", "tree"); // can be initialized dynamically
var conditions = expectedOptions.stream().map(Condition::text).collect(Collectors.toList());
$("label").shouldHave(new Or("Label must contains any value", conditions));
```
The same for `Add` condition

src/main/java/com/codeborne/selenide/Condition.java
```diff
@@ -524,8 +522,8 @@ public static Condition and(String name, Condition... conditions) {
    */
   @CheckReturnValue
   @Nonnull
-  public static Condition or(String name, Condition... conditions) {
-    return new Or(name, asList(conditions));
+  public static Condition or(String name, Condition condition1, Condition condition2, Condition... conditions) {
```
fokinp, line 525: what about to update `javadoc` too?

src/main/java/com/codeborne/selenide/conditions/Or.java
asolntsev: Hm.. well, yes, we can keep the previous signature. Though, I don't understand why somebody should directly use classes And and Or. There were never meant to be used directly, but only via `and` and `or` methods.

fokinp: these constructors are public and everybody can use it . We use it in our project :), since we need to dynamically get all possible options from db query. Nevertheless we can get quick understanding what is behind `Condition.or` realization by looking into method details.

asolntsev: @fokinp Why do you use constructor `new And(...)` instead of static method `Condition.and(...)`?

fokinp: at this moment we use use only `new Or(...)` condition where we need to check dynamic parameters, and we don't use `new And(..)`. Will be cool if we will support both options via `Condition` level:
* `Condition.or(String, Condition, Condition, Condition...)`
* `Condition.or(String, List<Condition>)`
* .. the same for `and`

Even more these conditions might accept only one `Condition`, however for this case there is not so big sense :).
The main disadvantage of `Condition.<...>(String, List<Condition>)` is that we can't check arguments size on compile time, but again I would say sometimes will be nice to have it.

asolntsev: @fokinp Sorry, I still don't understand why you prefer `new Or(...)` over static method `Condition.or(...)`.  What means "we need to check dynamic parameters"? Both method check parameters equally. 

> Even more these conditions might accept only one Condition

Yes, my idea was to disable one Condition because it's nonsense.

fokinp: > Sorry, I still don't understand why you prefer new Or(...) over static method Condition.or(...)

Because static method usage with current signature is additional overhead to convert `List<Condition>` into `Condition[]` (on client side, something like ` conditionList.toArray(Condition[]::new)`) and then `Condition[]` to `List<Condition>` (on selenide, where we have `asList(conditions)`)

> What means "we need to check dynamic parameters"?

Let's imagine the case:
* we have **common** method (not in a test body, we want to have something generic) that performs search based on selected values (multi select field)
* then we check result record and want to verify that one field contains any of selected values

and if we talk about a code,  it will be like:
```java
public void search(List<String> values) {
   $("multi-select").selectOption(values.toArray(String[]::new)); // and here will be nice to support List as well
  //all other stuff like form submit and so
 }

public void verify(List<String> values) {
    List<Condition> conditionList = values.stream().map(Condition::text).collect(Collectors.toList());
    $("multi-select").shouldHave(Condition.or("description", conditionList.toArray(Condition[]::new)));
     // OR 
    $("multi-select").shouldHave(new Or("description", conditionList));
}
```
Existing method signature at least makes possible to implement something like:
`$("multi-select").shouldHave(Condition.or("description", conditionList.toArray(Condition[]::new)));`
with additional transformation effort but new one significantly increases test automation effort to use this common methods if we deal with `List` or `Arrays` (we need manually map first and second element, then prepare `array` without two first elements and so) 

Personally I like the idea how it is implemented in hamcrest matchers, please have a look into [AnyOf](http://hamcrest.org/JavaHamcrest/javadoc/1.3/org/hamcrest/core/AnyOf.html) where we have a choice to use `Iterable` or `varargs`

asolntsev: @fokinp Thanks for the explanation. I love to get more knowledge about how Selenide users use it. :)

Ok, I have restore the previous signature of `And` and `Or` constructors. 

But I have to mention that this test design seems a very over-engineered for me. Loading conditions from database? In test? Really? It sounds really strange. 

Test must be simple and stable. What if the loading failed? What if database contains wrong values? What is the loaded list of conditions is empty? How can you be sure that your tests are correct and checking the right things?

src/main/java/com/codeborne/selenide/Condition.java
asolntsev: done

src/main/java/com/codeborne/selenide/conditions/Or.java
fokinp: @asolntsev thanks a lot.

Loading test data from database it special case, I would say, more frequent case is when we use common `set` and `verify` methods based on method input arguments, and complex ones require extracting them into `bean`/`dto`/`pojo` structure. And for these cases we need to have a simple solution (signature) how to use conditions.

PR #1371, opened 2021-01-15
support for fail fast mechanism

pavelpp:
## Proposed changes

The feature allows us to avoid waiting if we can fail fast. This helps in situations when you want to perform additional checks while waiting for an element to appear or disappear. For example, you can check if there are any alerts or error messages displayed and fail fast after submitting a form instead of waiting for some element to appear. The condition is re-evaluated for the duration of should wait, unless triggered, which means that it may add additional network delay, especially if we are calling selenium methods. 

Example usage:
   
```
  $("#submit-button").click()
  $("#success").failIf(new FailFastCondition(() -> $("#error").exists(), "error text detected!"))
         .shouldBe(visible, Duration.ofSeconds(15));
```
 
In the example above the error could happen after a couple of seconds and if we didn't intercept it
we would have wasted 15 seconds while waiting for a success message.

In theory, you can pass any boolean supplier to FailFastCondition - for example, you could monitor network errors using BrowserUp Proxy.

`failIf()` method works the same way as the recently added `as()` method. You set fail fast condition once and it is re-used further whenever calling any of the selenide methods on that same element.

## Checklist
- [x] Checkstyle and unit tests are passed locally with my changes by running `./gradlew check chrome_headless firefox_headless` command

### DIFF
--- a/config/checkstyle/checkstyle.xml
+++ b/config/checkstyle/checkstyle.xml
@@ -59,7 +59,9 @@
     <module name="CyclomaticComplexity">
       <property name="max" value="11"/>
     </module>
-    <module name="NPathComplexity"/>
+    <module name="NPathComplexity">
+      <property name="max" value="336"/>
+    </module>
 
 
     <module name="ArrayTypeStyle"/>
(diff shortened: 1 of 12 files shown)


asolntsev:
We agreed that it's better to implement this feature in the user's project because it's too specific.

PR #1377, opened 2021-01-19
Replace long timeout with Duration for ElementsCollection 

ostap-oleksyn:
## Proposed changes
Since recently all 'should' methods in SelenideElement have a Duration argument for condition waiting. To make things consistent these changes update all 'should' methods for ElementsCollection to also use Duration instead of long for timeouts.

## Checklist

### DIFF
--- a/statics/debug.log
+++ b/statics/debug.log
@@ -1,0 +1,1 @@
+[0119/221532.796:ERROR:directory_reader_win.cc(43)] FindFirstFile: Системе не удается найти указанный путь. (0x3)
--- a/statics/debug.log
+++ b/statics/debug.log
@@ -1,1 +1,0 @@
-[0119/221532.796:ERROR:directory_reader_win.cc(43)] FindFirstFile: Системе не удается найти указанный путь. (0x3)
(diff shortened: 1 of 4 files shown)


asolntsev:
@ostap-oleksyn Thank you. It's a good idea to use `Duration`, but shouldn't we also keep the old methods with `long` argument for backward compatibility? At least for some time. We could mark them as deprecated.

ostap-oleksyn:
> @ostap-oleksyn Thank you.
> 
> It's a good idea to use `Duration`, but shouldn't we also keep the old methods with `long` argument for backward compatibility? At least for some time. We could mark them as deprecated.
> 
> P.S. Don't pay attention to the failing tests: they were also failing in master. Already fixed in master by now.

Good point. I'll update it.

statics/debug.log
```diff
@@ -0,0 +1 @@
+[0119/221532.796:ERROR:directory_reader_win.cc(43)] FindFirstFile: ������� �� ������� ����� ��������� ����. (0x3)
```
rosolko: Please do not commit debug file

ostap-oleksyn: already removed it

asolntsev (approved):

PR #1390, opened 2021-01-29
Pass noproxy options to outside proxy

BorisOsipov:
## The problem
```
Proxy proxy = new Proxy();
proxy.setHttpProxy("127.0.0.1:3128"); 
proxy.setNoProxy("localhost");
WebDriverRunner.setProxy(proxy);      
Configuration.proxyEnabled = true;
open("http://localhost")
```
it can not open localhost because noproxy settings gets lost. 

## Checklist


asolntsev (approved):
@BorisOsipov do i correctly understand that this change is not "Pass noproxy options TO outside proxy", but "Pass noproxy options FROM outside proxy TO selenide proxy"?

otherwise looks good.

BorisOsipov:
@asolntsev yep.

PR #1409, opened 2021-02-18
Added method for getting clipboard content

dbudim:
## Proposed changes
Added method to get clipboard content. Acts via java.awt.Toolkit
Support selenoid clipboard will be added via selenoid plugin in [PR](https://github.com/selenide/selenide-selenoid/pull/6)

Test provided but disabled due to needs configure X11 for github actions, not sure that it needs

## Checklist


BorisOsipov:
Down vote. It will not work for any remote browser and it will confuse users. Also I think the need for access to clipboard is rare.

BorisOsipov:
Ohh I haven't read PR message. 
Ok It will work for selenoid. What should I do with default selenium grid? Nothing to do.

dbudim:
> Ohh I haven't read PR message.
> Ok It will work for selenoid. What should I do with default selenium grid? Nothing to do.

You right, nothing to do with grid, close PR.

BorisOsipov:
>You right, nothing to do with grid, close PR.

Don't give up so fast. It is just my opinion. Maybe @selenide/committers have different thoughts at least.

dbudim:
> > You right, nothing to do with grid, close PR.
> 
> Don't give up so fast. It is just my opinion. Maybe @selenide/committers have different thoughts at least.

Okay) Thanks) 

Besides i found solution for grid but it looks not maintainable and not scalable)
[stackoverflow](https://stackoverflow.com/questions/27248952/accessing-data-on-remote-machines-clipboard-when-running-tests-using-selenium-w)

asolntsev:
@dbudim While I don't imagine why it's needed to use clipboard in tests, I like the idea of having _default_ methods for accessing/verifying clipboard in Selenide and providing custom implementations for Selenoid and Selenium Grid. 

btw, the solution for Selenium Grid seems ok for me, probably we could also include it in Selenide? @BorisOsipov

dbudim:
> @dbudim While I don't imagine why it's needed to use clipboard in tests,

@asolntsev 
In my current project we have few scenarios where need to use this.
E.g. there is a field with url for shared resource and button "copy" that copies it to clipboard and need to check that was copied same as was displayed for user in this field)

We use selenoid and I don't though about grid firstly)

BorisOsipov:
>btw, the solution for Selenium Grid seems ok for me, probably we could also include it in Selenide?

Don't know how. It is a custom servlet in grid. We can't predict how users implemented it.

asolntsev:
@dbudim Let's discuss the API. How users will use the method `Selenide.clipboard()`? Did you mean to verify the text? 

Then we should probably add `Condition` instead? Like 

1. `Selenide.clipboard().shouldHave(text("expected text"));`
2. `Selenide.clipboard().setValue("expected text");`

@dbudim Can't we use JavaScript to access the clipboard? I found one solution that looks promising (we only need to find webdriver options for browsers): https://developer.mozilla.org/en-US/docs/Web/API/Clipboard_API

dbudim:
> @dbudim Let's discuss the API. How users will use the method `Selenide.clipboard()`? Did you mean to verify the text?
> 
> Then we should probably add `Condition` instead? Like
> 
> 1. `Selenide.clipboard().shouldHave(text("expected text"));`
> 2. `Selenide.clipboard().setValue("expected text");`
> 
> @dbudim Can't we use JavaScript to access the clipboard? I found one solution that looks promising (we only need to find webdriver options for browsers): https://developer.mozilla.org/en-US/docs/Web/API/Clipboard_API

@asolntsev 
I've used simple assertions but your idea with `shouldHave` looks great and intuitive) I think it will be very usefull)

What about JS let me some time for research and trying this solution, if it works fine it will be cool) I'll try and back with response)

BorisOsipov:
>I found one solution that looks promising (we only need to find webdriver options for browsers): developer.mozilla.org/en-US/docs/Web/API/Clipboard_API

@asolntsev main limitation is `This feature is available only in secure contexts (HTTPS)`. You must serve your app via https to use this

dbudim:
@asolntsev 
I checked JS solution with `window.navigator.clipboard.readText()` and it works but require to grant permissions.
And wen need to add ChromeOptions as was described here [medium](https://sahajamit.medium.com/getting-clipboard-content-from-remote-selenium-chrome-nodes-67a4c4d862bd)
`profile.content_settings.exceptions.clipboard : {
  "[*.],*": {
    "last_modified": "1576491240619",
    "setting": 1
  }
}`

or use selenium 4 `executeCdpCommand()`, here are explanation [stackoverflow](https://stackoverflow.com/questions/58034127/emulating-a-clipboard-copy-paste-with-selinum-capybara)

Seems it is not universal solutions which will not work everywhere due to difference in setting permissions in different browsers

asolntsev:
@dbudim Ok, then we have no choices except using `Toolkit.getDefaultToolkit().getSystemClipboard()`. Let's just add a warning in javadoc that 
1. this method might cause flaky tests if someone else changes clipboard content at the same time. 
2. this method doesn't work in headless mode (right?)

dbudim:
> @dbudim Ok, then we have no choices except using `Toolkit.getDefaultToolkit().getSystemClipboard()`. Let's just add a warning in javadoc that
> 
> 1. this method might cause flaky tests if someone else changes clipboard content at the same time.
> 2. this method doesn't work in headless mode (right?)

@asolntsev 

1. Yes, i think it make sense to add warning that method works with common system clipboard)
But i think that in most cases shouldn’t be any problem, because in common runs people use selenoid where clipboard is owned by container of current test, and locally usually runs only single test for debug.

2. Method works in headless mode (i checked additionally by adding `Configuration.headless=true;` ) and it works ok.
I disabled test because while running on Github Actions i got error:
`GetClipboardTest > getClipboard()
 FAILED
    java.awt.HeadlessException: 
    No X11 DISPLAY variable was set,
    but this program performed an operation which requires it.
        at java.desktop/sun.awt.HeadlessToolkit.getSystemClipboard(HeadlessToolkit.java:216)
        at com.codeborne.selenide.Clipboard.getString(Clipboard.java:18)
        at integration.GetClipboardTest.getClipboard(GetClipboardTest.java:25)`

It's needed to set env variable in workflow. I'm not sure about my rights to change workflow script))

So, i need:
1) Add warning to java doc
2) add methods `getText() | shouldHave(text("expected text")) | setValue("expected text")`
right?)

asolntsev:
Yes.
We recently had a bug in build.gradle which disabled headless mode (fixed today in master). Probably this failure was caused by this bug?

BorisOsipov:
> But i think that in most cases shouldn’t be any problem, because in common runs people use selenoid

@dbudim do you have reliable statistic source? From my point of view selenoid is popular in CIS, but in other world cloud based providers like SauceLab\Bs\Testtingbot are more popular. Also there are official selenium grid and his forks. So I guess selenoid doesn't cover "most cases" :)

@asolntsev seems `getSystemClipboard()` need virtual or real xserver. We can install it as we did it earlier for non-headless test run, but mot sure it really need. I guess such test actually tests JDK methods...

dbudim:
> @dbudim do you have reliable statistic source? From my point of view selenoid is popular in CIS, but in other world cloud based providers like SauceLab\Bs\Testtingbot are more popular. Also there are official selenium grid and his forks. So I guess selenoid doesn't cover "most cases" :)


@BorisOsipov No, i don't have statistics) And you almost convinced me that this feature shouldn't be aded 😂
Too many constraints)) And it can become like some methods in appium which doesn't wotk and you don't know why)

@asolntsev Maybe @BorisOsipov  right and we'll not to add clipboard?)

asolntsev:
@dbudim @BorisOsipov  I am also not sure. But if we assume that
1. Many people use Selenoid and do need this feature,
2. We can provide solution for Grid (we will add the servlet to selenide.jar, and people will just need to add selenide.jar to Selenium Grid classpath)
3. Other cloud providers also provide API for accessing clipboard, and we (or somebody else) could create selenide plugins for each of those providers. 

then it seems quite a reasonable feature...

dbudim:
> @dbudim @BorisOsipov I am also not sure. But if we assume that
> 
> 1. Many people use Selenoid and do need this feature,
> 2. We can provide solution for Grid (we will add the servlet to selenide.jar, and people will just need to add selenide.jar to Selenium Grid classpath)
> 3. Other cloud providers also provide API for accessing clipboard, and we (or somebody else) could create selenide plugins for each of those providers.
> 
> then it seems quite a reasonable feature...

@asolntsev 
Okay) if feature is reasonable, let's do it as written before) Firstly for regular runs and selenoid (with clear explanation in documentation what variants are supported ) and then if will be requested for other variants)
Moved PR to WIP while implement all that was discussed )

src/main/java/com/codeborne/selenide/Clipboard.java
```diff
@@ -0,0 +1,59 @@
...
+    Toolkit.getDefaultToolkit().getSystemClipboard().setContents(new StringSelection(text), new StringSelection(text));
+  }
+
+  /**
+   * Check that value in clipboard equals to expected
+   *
+   * @param text expected value for compare
+   */
+  public void shouldBeText(String text) {
+    assertEquals(text, getText(), "Clipboard data doesn't match with expected!");
```
BorisOsipov, line 56: It should not be Clipboard method. it must be separate condition class extends `Condition.java`

BorisOsipov: Ohh it would be nice. But it seems we don't have any Conditions that doesn't depend on element `$("")`

dbudim: I also want to use Condition but Clipboard is separate instance and it not depends from element)

BorisOsipov: @dbudim I would suggest not add condition in this PR and create separate issue to discuss how we can implement such conditions. 
As I remembered we already had requests to have condition for e.g. current url like `smth.shouldHave(currentUrl("https://twitter.com/selenide"))`

dbudim:
@asolntsev @dbudim please review PR when you'll have time)

src/main/java/com/codeborne/selenide/SelenideDriver.java
```diff
@@ -440,6 +440,12 @@ public SessionStorage getSessionStorage() {
     return new SessionStorage(driver());
   }
 
+  @CheckReturnValue
+  @Nonnull
+  public Clipboard getClipboard() {
+    return new ClipboardService(driver()).load();
```
asolntsev, line 446: just use `inject(ClipboardService.class)` instead.

src/main/java/com/codeborne/selenide/ClipboardService.java
```diff
@@ -0,0 +1,38 @@
...
+  public Clipboard load() {
+    if (isNullOrEmpty(driver.config().remote())) {
+      return initClipboardWithDriver(getDefaultPlugin(Clipboard.class).getClass());
+    } else {
+      Clipboard implementation = StreamSupport.stream(ServiceLoader.load(Clipboard.class).spliterator(), false)
+        .findFirst()
+        .orElseThrow(() -> new IllegalStateException("Remote clipboard plugin not defined!"));
+      return initClipboardWithDriver(implementation.getClass());
+    }
+  }
```
asolntsev, line 28: this code can be simplified. 
1. let's just have class `ClipboardService` with default constructor and method `public Clipboard getClipboard(Driver driver)`.
2. project `selenide-selenoid` will have a subclass `SelenoidClipboardService` with overridden method `getClipboard`. 
3. whoever needs `ClipboardService` will just call `inject(ClipboardService.class)` to get it.

src/main/java/com/codeborne/selenide/impl/Plugins.java
```diff
@@ -42,7 +42,7 @@
     return implementation;
   }
 
-  private static <T> T getDefaultPlugin(Class<T> klass) {
+  public static <T> T getDefaultPlugin(Class<T> klass) {
```
asolntsev, line 45: no need to make it public.

asolntsev (changes requested):

src/main/java/com/codeborne/selenide/ClipboardService.java
dbudim: @asolntsev 
Do you mean `Plugins.inject()`? 
In my case i didn't use it because my logic was that user often has two types of running test:
1) Local - when writing or debug
2) Remote - running on CI etc.

And i don't want to change configuration (enable and disable plugin) every time when run tests in different modes. 
`Plugins.inject()` works as if there are no services in `META-INF.services` loads default from, but i want to load default during local run and use custom during remote run. So `Plugins.inject()` isn't suitable for this logic )
Are you agree, or maybe i'm not clearly understood you?)

src/main/java/com/codeborne/selenide/impl/Plugins.java
dbudim: need for my logic with loading default plugin for local run and custom plugin for remote

src/main/java/com/codeborne/selenide/ClipboardService.java
dbudim: Done as was requested:
- made `ClipboardService` with default constructor and method getClipboard(Driver driver)

src/main/java/com/codeborne/selenide/impl/Plugins.java
dbudim: reverted changes

src/main/java/com/codeborne/selenide/SelenideDriver.java
dbudim: done

asolntsev (approved):

Issue #1110, opened 2020-04-09
Implement Drag'nDrop using JS code


asolntsev:
This js library can be used:
https://github.com/Photonios/JS-DragAndDrop-Simulator

## The problem
Drag'nDrop doesn't work properly in most (all?) webdrivers. People need a workarohnd in Selenide.

dstekanov:
Hi @asolntsev, it works with Chrome.
But I cannot figure out why it does not want to work with Firefox...

kikxter:
I haven't been able to make it work on either Chrome, Firefox or Edge. 

Using Selenide v. 5.5.0
BrowserName=chrome Version=86.0.4240.111 Platform=MAC
Selenium WebDriver v. 3.141.59

My issue is that the DROP is done wherever my mouse pointer is located at the moment not where the TO (WebElement / SelenideElement ) is located at.

dbudim:
working on it in [PR](https://github.com/selenide/selenide/pull/1412)

dbudim:
done, please review [PR](https://github.com/selenide/selenide/pull/1412)

camohub:
So which version of Selenide is fixed? I have 5.19.0

asolntsev:
The feature was released in Selenide 5.19.0:

![image](https://user-images.githubusercontent.com/279773/111294722-49976c00-8653-11eb-8439-4526228c2975.png)

PR #1412, opened 2021-02-20
 Implemented drag and drop by JS  (issue 1110)

dbudim:
## Proposed changes
Implemented [issue 1110](https://github.com/selenide/selenide/issues/1110)
Tested both on chrome and firefox, works fine

## Checklist


asolntsev (changes requested):
@dbudim Thank you for the PR! People do need a working Drag'n'Drop. :)

1. Current implementation only works on pages with JQuery. But nowdays JQuery is not so popular anymore. Can't we rewrite the JS code to work without JQuery? Probably http://youmightnotneedjquery.com/ can help.
2. Instead of having two methods `$.dragAndDrop()` and `$.dragAndDropByJs()`, it's probably better to have one method `$.dragAndDrop()` with parameters.

Then it would look like `$.dragAndDrop(usingJavaScript())`. See for example similar api for clicks: https://ru.selenide.org/2020/09/26/selenide-5.15.0/

dbudim:
> @dbudim Thank you for the PR! People do need a working Drag'n'Drop. :)
> 
> 1. Current implementation only works on pages with JQuery. But nowdays JQuery is not so popular anymore. Can't we rewrite the JS code to work without JQuery? Probably http://youmightnotneedjquery.com/ can help.
> 2. Instead of having two methods `$.dragAndDrop()` and `$.dragAndDropByJs()`, it's probably better to have one method `$.dragAndDrop()` with parameters.
> 
> Then it would look like `$.dragAndDrop(usingJavaScript())`. See for example similar api for clicks: https://ru.selenide.org/2020/09/26/selenide-5.15.0/

@asolntsev done)
I found implementation on JS and used it, works fine)
Also moved solution to existed drag and drop command. 
As for drag function there are no many arguments as for ClickOptions, i've just added overloaded version with boolean param:
`dragAndDropTo(String targetCssSelector, boolean useJavaScript);`

So we have regular`$("#drag1").dragAndDropTo("#div2");` and JS `$("#drag1").dragAndDropTo("#div2", true);`

If it's ok, all is done, if need to make same as ClickOptions please tell, i'll do)

BorisOsipov:
>i've just added overloaded version with boolean param

it isn't smell good. I think `FooBarOptions` will be better.

dbudim:
> > i've just added overloaded version with boolean param
> 
> it isn't smell good. I think `FooBarOptions` will be better.

@BorisOsipov I've made this because i don't imagine what other options could be added to DragAndDrop command. So i thought that creating DragAndDropOptions for only one option is not needed) But if need make it for consistency with other commands I'll make it)

Should i move from boolean param to DragAndDropOptions ?)

asolntsev:
@dbudim I can imagine quite many options: speed of moving the cursor, pause before, pause after, shift from the center of the element etc.

dbudim:
> @dbudim I can imagine quite many options: speed of moving the cursor, pause before, pause after, shift from the center of the element etc.


> @dbudim I can imagine quite many options: speed of moving the cursor, pause before, pause after, shift from the center of the element etc.

@asolntsev 
Wow it's a great field for customisation) For me there was sufficient standart implementation )

Done)
Added `dragAndDropTo(String targetCssSelector, DragAndDropOptions options);`

asolntsev (approved):

asolntsev:
@dbudim Wow, it's pure JS and it works in all browsers! Bravo!

asolntsev:
@dbudim I think I will even enable JS mode by default. Because "actions" mode doesn't really work in any browsers.

dbudim:
> @dbudim I think I will even enable JS mode by default. Because "actions" mode doesn't really work in any browsers. 

Great)
Also I checked logs, even with spending time for reading js from file it works faster than regular actions)

vprystash:
Hi, when it will be possible to use this feature?

asolntsev:
I am going to release Selenide 5.19.0 soon. Tonight or tomorrow.

vprystash:
Sorry for disturbing you, but I really need this feature about drag&drop, when will it be possible to update?

rosolko:
> Tonight or tomorrow.

asolntsev:
@vprystash Well, will release it today. For sure. :)

PR #1414, opened 2021-02-22
Get element alias

pavelpp:
## Proposed changes
Added a method to get SelenideElement alias. It may be used to log SelenideElements with ugly xpath'es etc. Until now you could only set alias with as(), method, but not read it.

## Checklist
- [x ] Checkstyle and unit tests are passed locally with my changes by running `gradlew check chrome_headless firefox_headless` command
- [x ] I have added tests that prove my fix is effective or that my feature works
- [ x] I have added necessary documentation (if appropriate)


pavelpp:
committed unwanted changes

Issue #1406, opened 2021-02-15
Using ClickOptions on IE raises an JavascriptException "Object doesn't support this action" .


tmshft:
Hi . 
ClickOptions feature is so useful and I love it.
But JavascriptException raised on IE.

## The problem

I tried to click button tag by using `ClickOptions.usingJavaScript()` on IE11.

Then JavascriptException has raised.

If I use `#usingDefaultMethod` , the test is passed(we can  click element)

## environment

* **Selenide Version**: 5.18.0
* **Chrome\Firefox\IE Version**: IE11
* **Browser Driver Version**: IEDriverServer 3.150
* **Selenium Version**: 3.141.59
* **OS Version**: Windows10

## Code To Reproduce Issue [ Good To Have ]

source 

```
$("#mylocator").click(ClickOptions.usingJavaScript());
```

html source of target button.

```html
<button class="somthing" tabindex="0" type="submit" id="RacineSubmit" name="Submit"><span class="something">login</span></button>
```

log 

```log
2021-02-15 xx:28:59,793 [Test worker] [INFO]:com.codeborne.selenide.drivercommands.CreateDriverCommand#createDriver:Create webdriver in current thread 14: InternetExplorerDriver -> InternetExplorerDriver: internet explorer on WINDOWS (xxxxxx)

JavascriptException: Error from JavaScript: Object doesn't support this action
Screenshot: path
Page source: path
Timeout: 10 s.
Caused by: JavascriptException: Error from JavaScript: Object doesn't support this action
	at com.codeborne.selenide.ex.UIAssertionError.wrapToUIAssertionError(UIAssertionError.java:83)
	at com.codeborne.selenide.ex.UIAssertionError.wrapThrowable(UIAssertionError.java:74)
	at com.codeborne.selenide.ex.UIAssertionError.wrap(UIAssertionError.java:68)
	at com.codeborne.selenide.impl.SelenideElementProxy.invoke(SelenideElementProxy.java:90)
	at com.sun.proxy.$Proxy45.click(Unknown Source)
	at mypackage.WebTestUtils.login(WebTestUtils.java:93)
```

arunahk:
Is anyone checking on this?

asolntsev:
@arunahk I am going to find some windows laptop and debug this issue. But it takes some time. It would be great if you do it sooner.

BorisOsipov:
[...]

PR #1419, opened 2021-02-23
Fix click js in ie

asolntsev:
fixes https://github.com/selenide/selenide/issues/1406

had to use a different JS code to trigger "click" event in IE.

### DIFF
--- a/src/main/java/com/codeborne/selenide/impl/Plugins.java
+++ b/src/main/java/com/codeborne/selenide/impl/Plugins.java
@@ -26,8 +26,13 @@ public class Plugins {
   private static final Map<Class<?>, Object> cache = new ConcurrentHashMap<>();
 
   @SuppressWarnings("unchecked")
-  public static <T> T inject(Class<T> klass) {
-    return (T) cache.computeIfAbsent(klass, Plugins::loadPlugin);
+  public static synchronized <T> T inject(Class<T> klass) {
+    T plugin = (T) cache.get(klass);
+    if (plugin == null) {
+      plugin = loadPlugin(klass);
+      cache.put(klass, plugin);
+    }
+    return plugin;
   }
 
   private static <T> T loadPlugin(Class<T> klass) {
--- a/src/main/resources/click.js
+++ b/src/main/resources/click.js
@@ -1,0 +1,10 @@
+(function (element, offsetX, offsetY) {
+  const rect = element.getBoundingClientRect();
+  element.dispatchEvent(new MouseEvent('click', {
+    'view': window,
+    'bubbles': true,
+    'cancelable': true,
+    'clientX': rect.left + rect.width / 2 + offsetX,
+    'clientY': rect.top + rect.height / 2 + offsetY
+  }));
+})(arguments[0], arguments[1], arguments[2]);
--- a/src/main/resources/click.js
+++ b/src/main/resources/click.js
@@ -1,10 +1,25 @@
 (function (element, offsetX, offsetY) {
   const rect = element.getBoundingClientRect();
-  element.dispatchEvent(new MouseEvent('click', {
-    'view': window,
-    'bubbles': true,
-    'cancelable': true,
-    'clientX': rect.left + rect.width / 2 + offsetX,
-    'clientY': rect.top + rect.height / 2 + offsetY
-  }));
+
+  function mouseEvent() {
+    if (typeof (Event) === 'function') {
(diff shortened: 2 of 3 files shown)


src/main/java/com/codeborne/selenide/impl/Plugins.java
```diff
@@ -26,8 +26,11 @@
   private static final Map<Class<?>, Object> cache = new ConcurrentHashMap<>();
 
   @SuppressWarnings("unchecked")
-  public static <T> T inject(Class<T> klass) {
-    return (T) cache.computeIfAbsent(klass, Plugins::loadPlugin);
+  public static synchronized <T> T inject(Class<T> klass) {
```
BorisOsipov, line 29: looks like not relevant change...

BorisOsipov:
@asolntsev js click script changes LGTM

src/main/java/com/codeborne/selenide/impl/Plugins.java
asolntsev: @BorisOsipov Yes, it's not related by IE, but it's related to Windows. 
For some reason, this bug happened only on Windows where I was testing my changes on IE browser.

BorisOsipov: to be honest I don't understand why it needs. computeIfAbsent is CAS operations and should guarantee the same semantics...

BorisOsipov: but `loadPlugin` uses `ServiceLoader.load` and it is kind of black magic...

asolntsev: That's the point. This is the line from `java.util.concurrent.ConcurrentHashMap#computeIfAbsent` javadoc:
> the computation should be short and simple, and must not attempt to update any other mappings of this map.
> 
> Throws:
> IllegalStateException – if the computation detectably attempts a recursive update to this map that would otherwise never complete

Though, it doesn't fail every time, but occasionally. It seems to depend on order of elements in HashMap or something similar.

asolntsev: The problem is not with ServiceLoader.load(). The problem is with our code:
1. inject(Commands.class) calls `new Commands()`
2. which in turn calls `new DownloadFile()`
3. which in turn calls `inject(DownloadFileToFolder.class)`
4. which calls `Plugins.cache.computeIfAbsent()` - during the previous `computeIfAbsent()`. It _sometimes_ causes `IllegalStateException` inside `computeIfAbsent()`.

src/main/resources/click.js
```diff
@@ -0,0 +1,25 @@
...
+      return new MouseEvent('click', {
+        'view': window,
+        'bubbles': true,
+        'cancelable': true,
+        'clientX': rect.left + rect.width / 2 + offsetX,
+        'clientY': rect.top + rect.height / 2 + offsetY
+      });
+    }
+    else {
+      const event = document.createEvent('MouseEvent');
```
vladimirjegorov, line 15: could you please clarify? :) is this JS executed by browser or something else? 
if by browser (and since this is meant to fix issue in IE11) - IE11 (and some other potato/vintage browsers) does not support declaring variables via `const` and `let`

BorisOsipov: IE11 supports `const`. See https://caniuse.com/?search=const

asolntsev: ups, probably it would be safer to use `var` instead of `const`. :)

But I tested this code with IE11, and it worked. Do you know some browsers that are used for testing via webdriver and don't support `const`?

vladimirjegorov: Tested on IE11 that's (unfortunately) installed on my corporate PC, version: 
image
results:
image
then again - maybe that's just the browser's console that's just lying :) but most probably using the old fashioned `var` would've been safer :)
However if the existing solution fixes OP's problem then never mind ;)

asolntsev: But does `$("#mylocator").click(ClickOptions.usingJavaScript());` work in your IE?

vladimirjegorov: ok, most probably it's indeed related to how IE11's console treats variable scope :) further tests revealed that if block of code is run as multiple lines, then the result is as expected:
image
Thus I assume that the `click(usingJavaScript());` will indeed work in my IE :) Sorry to have bothered you :)

Issue #1422, opened 2021-03-04
Microsoft Edge Headless


mystygage:
## The problem
I want to run Microsoft Edge in headless mode and set `selenide.headless=true`.
The browser was opened anyway.

## Tell us about your environment
* **Selenide Version**: 5.19.0
* **Microsoft Edge**: 88.0.705.81
* **Selenium Version**: 3.141.59
* **OS Version**: macOS 11.2.2

## Code To Reproduce Issue [ Good To Have ]
https://github.com/mystygage/selenide/commit/9d41bd7f15bbc7f04716a0207c654b3137689a96

asolntsev:
@mystygage The problem is that EdgeDriver doesn't support headless mode (in Selenium 3).
It will support in Selenium 4, but it's not released yet...

But I found a workaround: https://github.com/selenide/selenide/pull/1424

mystygage:
Awesome. That would help a lo!

PR #1424, opened 2021-03-04
#1422 add headless mode to Microsoft Edge browser (chromium-based)

asolntsev:
P.S. After upgrading to Selenium 4 this code will not be needed.


BorisOsipov (approved):

rosolko (approved):
LGTM!

PR #1426, opened 2021-03-07
Add CollectionCondition.containTexts() method

bereg2k:
## Proposed changes
This is a method for asserting that `ElementsCollection` **contains** a given list of strings/texts. 
All the existing methods for checking collection's elements against list of strings assume that `expectedCollection.size() == expectedTexts.size()` (in order to succeed). 
Although, sometimes it's useful to assert that a collection contains a *baseline* of elements with a given strings/texts.  

### Example E2E scenario:
The application provides a list of available currencies for the client. The list comes from the external system.
Verify that the external system always provides "USD", "EUR" and "GBP" currencies for this client.
The list from the external system may vary (some currencies may be added, some may be dropped), but it should always contain 3 specific currencies - "USD", "EUR" and "GBP".

Possible output in DOM:
```html
<div test-id='availableCurrencies'>
    <div test-id='currency'>USD</div>
    <div test-id='currency'>GBP</div>
    <div test-id='currency'>EUR</div>
    <div test-id='currency'>RUB</div>
    <div test-id='currency'>CAD</div>
</div>
```

Currency collection may be validated like this:
`$$("[test-id='currency']").should(containTexts("USD", "EUR", "GBP"));`

Without this method, Selenide's current functionality allows doing such assertions like this:
```
$$("[test-id='currency']").shouldHave(itemWithText("USD"));
$$("[test-id='currency']").shouldHave(itemWithText("EUR"));
$$("[test-id='currency']").shouldHave(itemWithText("GBP"));
```
As far as I understand, this approach requires more code and slows down the test a bit. 
That's why I think the addition of `CollectionCondition.containTexts` might be useful.

## Checklist
- [x] I have added **unit** tests that prove my fix is effective or that my feature works

### DIFF
--- a/src/main/java/com/codeborne/selenide/CollectionCondition.java
+++ b/src/main/java/com/codeborne/selenide/CollectionCondition.java
@@ -180,11 +180,11 @@ public abstract class CollectionCondition implements Predicate<List<WebElement>>
    * Examples:
    * <pre code='java'>
    * // collection 1: [Tom, Dick, Harry]
-   * $$("collection 1 locator").should(containTexts("Tom", "Dick", "Harry); // success
+   * $$("collection 1 locator").should(containTexts("Tom", "Dick", "Harry"); // success
    * // collection 2: [Tom, John, Dick, Harry]
-   * $$("collection 2 locator").should(containTexts("Tom", "Dick", "Harry); // success
+   * $$("collection 2 locator").should(containTexts("Tom", "Dick", "Harry"); // success
    * // collection 3: [John, Harry, Tom]
-   * $$("collection 3 locator").should(containTexts("Tom", "Dick", "Harry); // fail
+   * $$("collection 3 locator").should(containTexts("Tom", "Dick", "Harry"); // fail
    * </pre>
    *
    * @param expectedTexts the expected texts that the collection should contain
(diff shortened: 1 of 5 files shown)


asolntsev (approved):

src/main/java/com/codeborne/selenide/CollectionCondition.java
```diff
@@ -173,6 +174,45 @@ public static CollectionCondition itemWithText(String expectedText) {
     return new ItemWithText(expectedText);
   }
 
+  /**
+   * Check that the given collection contains elements with given texts.
+   * <p></p>
+   * Examples:
+   * <pre code='java'>
+   * // collection 1: [Tom, Dick, Harry]
+   * $$("collection 1 locator").should(containTexts("Tom", "Dick", "Harry); // success
```
BorisOsipov, line 183: (minor) missed trailing quotation mark `"Harry);` here and next

BorisOsipov (commented):
LGTM please fix minor issue

src/main/java/com/codeborne/selenide/CollectionCondition.java
bereg2k: Thank you. 
Fixed all 6 occurrences.

PR #1428, opened 2021-03-08
Do not log extended "find" methods (e.g. parent, sibling and so)

fokinp:
"Find" methods might be used for Page Objects (PO) fields declaration
so in such cases they will be logged during PO initialization phase
no matter of they are used in real test or not

## Proposed changes

## Checklist


fokinp:
As proposal, to increase readability of such find methods we can use `as(String)` method call inside command methods.  What do you think ?

asolntsev (approved):

Issue #1389, opened 2021-01-28
Add method `$$.as(“name”)` for ElementCollection


asolntsev:
Similar to  `$.as(“name”)`

PR #1431, opened 2021-03-10
#1389 add method $$.as()

asolntsev:


statics/src/test/java/integration/errormessages/WrappedElementTest.java
```diff
@@ -0,0 +1,71 @@
...
+
+  static class PageObject {
+    @FindBy(tagName = "h2")
+    private WebElement header;
+
+    @FindBy(id = "invalid_id")
+    private WebElement categoryDropdown;
+
+    /**
+     * Classical Selenium-style page object :(
```
bereg2k, line 65: "Classic"

src/main/java/com/codeborne/selenide/impl/Alias.java
```diff
@@ -0,0 +1,53 @@
...
+  @CheckReturnValue
+  public String getOrElse(Supplier<String> defaultValue) {
+    return text;
+  }
+
+  /**
+   * As a rule, you don't need to use this method directly.
+   * @return text value of this alias (or empty text if alias is not defined)
+   */
+
```
bereg2k, line 26: extra line?

src/main/java/com/codeborne/selenide/ElementsCollection.java
```diff
@@ -469,13 +469,30 @@ public ElementsCollection snapshot() {
     return new ElementsCollection(new CollectionSnapshot(collection));
   }
 
+  /**
+   * Give this collection a human-readable name
+   *
+   * Caution: you probably don't need this method.
+   * It's always a good idea to have the actual selector instead of "nice" description (which might be misleading or even lying).
```
bereg2k, line 476: Maybe it's worth adding some `<p>` tags to make this JavaDoc look exactly as you meant it to be here.

rosolko (approved):
Looks pretty 👍

src/main/java/com/codeborne/selenide/ElementsCollection.java
asolntsev: I tried to add some `<p>`, but don't really see any difference. Probably I am just doing it wrong. 

@bereg2k Don't you want to send a PR with javadoc improvements?

src/main/java/com/codeborne/selenide/impl/Alias.java
asolntsev: thanks, fixed.

statics/src/test/java/integration/errormessages/WrappedElementTest.java
asolntsev: thanks, fixed.

src/main/java/com/codeborne/selenide/ElementsCollection.java
bereg2k: Without tags. Looks nice in the "edit" form.
image

But not as nice in the "read" form:
image
image

bereg2k: Even a single usage of tags makes it look as it was intended:

image
image

Or even this:
image

image

@asolntsev I'm up for the task btw, but I'm just saying, that any particular case of JavaDoc might benefit from some tagging.

asolntsev: Thank you for the knowledge sharing. 
But when I wrapped every clause to `<p></p>`, it still didn't look well in the resulting JavaDoc. 

Well, I see that you are not wrapping text to `<p>` tags, but using empty `<p></p>` essentially to create line break (like `<br>`). Seems strange for me...

Anyway, it seems that quite a lot of Selenide javadoc would need this formatting improvement, so probably it's a good idea to create a dedicated pull request.

PR #1433, opened 2021-03-12
override default Selenium http timeouts

asolntsev:
A temporary workaround to override default timeouts of OkClient used in Selenium.

Its default timeouts are incredibly long:
1. connectTimeout = 120000 ms
2. readTimeout = 10800000 ms

If it's fixed in Selenium 4, this workaround can be removed.

### DIFF
--- a/src/main/java/com/codeborne/selenide/webdriver/HttpClientTimeouts.java
+++ b/src/main/java/com/codeborne/selenide/webdriver/HttpClientTimeouts.java
@@ -1,0 +1,79 @@
+package com.codeborne.selenide.webdriver;
+
+import org.openqa.selenium.WebDriver;
+import org.openqa.selenium.remote.CommandExecutor;
+import org.openqa.selenium.remote.HttpCommandExecutor;
+import org.openqa.selenium.remote.RemoteWebDriver;
+import org.openqa.selenium.remote.http.HttpClient;
+import org.openqa.selenium.remote.internal.OkHttpClient;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import javax.annotation.ParametersAreNonnullByDefault;
+import java.lang.reflect.Field;
+import java.time.Duration;
+
+/**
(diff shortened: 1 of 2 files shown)


BorisOsipov:
So ironic that integration test in chrome fails with 
```
canUseSizzleSelectors_onTodoList_jquery()
org.openqa.selenium.WebDriverException: java.net.SocketTimeoutException: timeout
Build info: version: '3.141.59', revision: 'e82be7d358', time: '2018-11-14T08:17:03'
System info: host: 'fv-az203-371', ip: '10.1.0.4', os.name: 'Linux', os.arch: 'amd64', os.version: '5.4.0-1040-azure', java.version: '11.0.10'
Driver info: driver.version: SelenideDriver
selenide.url: https://todomvc4tasj.herokuapp.com/
selenide.baseUrl: https://127.0.0.1:18495
	at org.openqa.selenium.remote.service.DriverCommandExecutor.execute(DriverCommandExecutor.java:92)
	at org.openqa.selenium.remote.Remot
	```

BorisOsipov:
I don't see issue that connectTimeout is 120000 and suggest doesn't change it. It is not so long.
I don't know why readTimeout is so incredibly long and make it less seems reasonable for me.

Btw I walk through selenium codebase and find new default timeouts https://github.com/SeleniumHQ/selenium/blob/1d31b391b61837b62283b4b85e9d2eba55760350/java/client/src/org/openqa/selenium/remote/http/ClientConfig.java#L55

maybe use it as defaults for us?

asolntsev:
> maybe use it as defaults for us?

1. No, connectTimeout 120000 = 2 minutes is incredibly large timeout. If browser hangs, tests will fail only after 2 minutes. 
2. In new Selenium code, readTimeout is 3 minutes - which is not good at all. When my test tries to take a screenshot, and browser fails, the screenshot method will fail only after 3 minutes. Is it really ok?

BorisOsipov (approved):

PR #1435, opened 2021-03-14
Additional logging

fokinp:
## Proposed changes
Log `SelenideTargetLocator` methods except `alert` one
 _`alert` is skipped to avoid duplicated logging with Modal methods (`prompt/confirm/dismiss`)_

E.g.
Element | Subject
---------| --------
frame(index: 1) | switch to()
frame(mainFrame) | switch to()
parent frame | switch to()
default context | switch to()
active element | switch to()
window(index: 1) | switch to()
window(mainWindow) | switch to()

Log `Selenide.closeWindow` method call
_Especially helpful when we operate with several windows in a test_

E.g.
Element | Subject
---------| --------
current window | close()

Improve `Modal` methods log messages
_Specify `element` as `modal` and `subject` as action for the element. See:_

_was:_
Element | Subject
---------| --------
dismiss |
dismiss | Are you sure?
prompt |  ->
prompt | Sure do ->
prompt | Sure do -> You really want it?
confirm |    
confirm |You really want it?    

_now:_
Element | Subject
---------| --------
modal() | dismiss()
modal(Are you sure?) | dismiss()
modal() | prompt()  
modal() | prompt(Sure do)
modal(Are you sure?) | prompt(Sure do) 
modal() | confirm()
modal(Are you sure?) | confirm()

## Checklist


src/main/java/com/codeborne/selenide/SelenideTargetLocator.java
```diff
@@ -129,15 +138,17 @@ public Alert alert() {
    */
   @Nonnull
   public WebDriver innerFrame(String... frames) {
-    delegate.defaultContent();
+    defaultContent();
```
fokinp, line 141: just to log this action as well

asolntsev (approved):
@fokinp Code looks good. 
But I am not sure about the goal. Now we have a longer output (with lot of duplicate string "modal()") which doesn't give any more information. Is it worth it?

fokinp:
> But I am not sure about the goal. Now we have a longer output (with lot of duplicate string "modal()") which doesn't give any more information. Is it worth it?

@asolntsev I am agree that now we produce a bit long output and I am ready to any proposals how to make them more clear and shorter. Currently `Modal` logging is implemented by `SelenideTargetLocator` analogy (please see updated PR description). As options here we can:
* omit `modal` if `expectedDialogText` is set
* replace `modal` with `dialog`
* omit `()`

E.g.
Element | Subject
---------| --------
dialog | dismiss()
Are you sure? | dismiss()

Any thoughts?

asolntsev:
@fokinp In the beginning, I didn't realise we don't have any logging at all. I thought this PR is just an refactoring of an existing logging. 
Thanks, merged.

PR #1439, opened 2021-03-21
New naming and assertion error for CollectionCondition.containExactTextsCaseSensitive() (prev. containTexts())

bereg2k:
## Proposed changes
- Сhanged the name for `CollectionCondition.containTexts()` method - `containExactTextsCaseSensitive()`. Thanks to @fokinp for the suggestion.
- Changed the error message for `CollectionCondition.containExactTextsCaseSensitive()`.

Given:
```java
// collection has text elements [one, two, five, six, seven]
$$("#collection-id").should(containExactTextsCaseSensitive("one", "two", "three", "four"));
```
Assertion error before:
```
Texts mismatch
Actual: [one, two, five, six, seven]
Expected: [one, two, three, four]
Collection: #collection-id
```
Assertion error after:
```
The collection with text elements: [one, two, five, six, seven]
should contain all of the following text elements: [one, two, three, four]
but could not find these elements: [three, four]
Collection: #collection-id
```
- Added **negative UI test** for this feature
- Fixed **unit tests** for this feature
- Minor JavaDoc changes (examples)

## Checklist

### DIFF
--- a/src/main/java/com/codeborne/selenide/ElementsCollection.java
+++ b/src/main/java/com/codeborne/selenide/ElementsCollection.java
@@ -83,9 +83,8 @@ public class ElementsCollection extends AbstractList<SelenideElement> {
   /**
    * Check if a collection matches given condition(s).
    * <p> For example: </p>
-   * {@code $$(".text_list").should(containTexts("text1", "text2"));}
-   * {@code $$(".cat_list").should(allMatch("value==cat",
-   * el -> el.getAttribute("value").equals("cat")));}
+   * {@code $$("#text_list").should(containTexts("text1", "text2"));}
+   * {@code $$("#cat_list").should(allMatch("value==cat", el -> el.getAttribute("value").equals("cat")));}
    */
   @Nonnull
   public ElementsCollection should(CollectionCondition... conditions) {
--- a/src/main/java/com/codeborne/selenide/ElementsCollection.java
+++ b/src/main/java/com/codeborne/selenide/ElementsCollection.java
@@ -83,8 +83,10 @@ public class ElementsCollection extends AbstractList<SelenideElement> {
   /**
    * Check if a collection matches given condition(s).
    * <p> For example: </p>
-   * {@code $$("#text_list").should(containTexts("text1", "text2"));}
-   * {@code $$("#cat_list").should(allMatch("value==cat", el -> el.getAttribute("value").equals("cat")));}
+   * <pre code='java'>
+   * $$(".text_list").should(containExactTextsCaseSensitive("text1", "text2"));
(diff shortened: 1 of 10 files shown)


src/main/java/com/codeborne/selenide/collections/ContainTexts.java
```diff
@@ -45,14 +45,16 @@ public void fail(CollectionSource collection,
                    @Nullable List<WebElement> elements,
                    @Nullable Exception lastError,
                    long timeoutMs) {
+    List<String> actualTexts = ElementsCollection.texts(elements);
```
fokinp, line 48: what about to move this logic under `-else-` ?

src/main/java/com/codeborne/selenide/ElementsCollection.java
```diff
@@ -83,9 +83,8 @@ public ElementsCollection shouldHaveSize(int expectedSize) {
   /**
    * Check if a collection matches given condition(s).
    * <p> For example: </p>
-   * {@code $$(".text_list").should(containTexts("text1", "text2"));}
-   * {@code $$(".cat_list").should(allMatch("value==cat",
-   * el -> el.getAttribute("value").equals("cat")));}
+   * {@code $$("#text_list").should(containTexts("text1", "text2"));}
```
fokinp, line 86: I totally understand that potentially it's too late, but anyway, imho based on existing selenide naming it should be named as `containExactTexts`

src/main/java/com/codeborne/selenide/collections/ContainTexts.java
bereg2k: Good point. 
Initially, I had to put this line first because it was `actualTexts` also used in one of the `else if` branches (which is gone now).

It's fixed now, moved to `else` branch.

src/main/java/com/codeborne/selenide/ElementsCollection.java
asolntsev: @fokinp  good point!

bereg2k: Yes, it sounds absolutely reasonable. 

I changed the name to `containExactTextsCaseSensitive` for the main methods, and everywhere else in the dependent code. 
Now the result code should look like this:
`$$(".list").should(containExactTextsCaseSensitive("text1", "text2"));`

fokinp: I'm just worried a bit that this change breaks existing contract and we should additionally notify all users :)

bereg2k: ... and that the new version will follow soon =)

asolntsev (approved):

PR #1441, opened 2021-03-29
#1429 add overload methods to sibling()

oleksandr-diachenko:
## Proposed changes
Added overload methods to sibling()

SelenideElement sibling(String tag);
SelenideElement sibling(String tag, int index);
SelenideElement sibling(String tag, String attributeName, String attributeValue);

https://github.com/selenide/selenide/issues/1429

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Changelog
 
+## 5.20.2 (released 29.03.2021)
+* #1429 add overload methods to sibling() (by tag name, by tag and index, by tag attribute name and attribute value) 
+
 ## 5.20.1 (released 23.03.2021)
 * #1438 Added $$.should() method + JavaDoc fixes  --  thanks to Oleg Berezhnoy @bereg2k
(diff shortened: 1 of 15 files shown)


BorisOsipov:
Hey @oleksandr-diachenko thank you to help with this feature.

From my point of view we don't need it. These are the main reasons why:
1) We got only one request to add this methods. Seems it is not so important and useful.
2) It covers only limited amount of selector combinations cases but adds bunch of new methods to `SelenideElements` that are not concise and not easily understood. As I see it should not be 100500 overloads but like `click()` one method (sibling, parent, etc) that accepts params object with concise builder api e.g.
`$("lol").sibling(withTag("kek").index(2))`
3) When we add this we encourage writing such weird code like: 
`$("#lol).sibling("label", "name", "my-shitty-input").sibling("input", 1);`

oleksandr-diachenko:
@BorisOsipov i also think we don't need this, that's why i've closed this PR. Don't pay attention to my wasted time,  best way to understand how technology works from the inside is to write something :)

BorisOsipov:
>best way to understand how technology works from the inside is to write something :)
That's true. 👍
Usually I starts my contributions to OSS projects with the same motivation.

Issue #280, opened 2016-02-06
Add the possibility to wait for a HTML page title


peterkeller:
Actually it's possible to query the HTML page title with `title()`. However, this method returns a `String` and not an implementation of `SelenideElement`. Consequently, it is not possible to wait for a HTML page title such as:

    $(title("My Page Title")).shouldBe(present); // NOK

However, there are situations where the page content is not relevant and it is sufficient to test the HTML page title.

Please, add the possibility to wait for a HTML page title.

(I am aware of `byTitle(String)`, but this method returns elements having a title attribute with given value and not the HTML page title.)

asolntsev:
@peterkeller We can do it, but can you explain why it's needed?
I think it's not a good idea to depend on page title. First of all, because typical user doesn't see this title. You should better to check something on the page. Some text or image.

vinogradoff:
I think the question is reasonable, but just wondering…

if straightforward

$(„title“).shouldHave(exactText((„My Page Title“))

does the job?

vinogradoff:
If you just want make sure the page is loaded, you probably do not need any "wait", because ajax normally doesn't change title dynamically. But it can :) Anyway, the example above will wait if needed, and not wait if not needed.

peterkeller:
@vinogradoff: I tried your suggestion, however my little JUnit test fails:

    open("http://www.google.com");
    assertEquals("Google", title());   // OK
    $("title").shouldHave(exactText("Google"));  // NOK

Output:

> Element should have exact text 'Google' {title}
Element: '<title displayed:false></title>'
Screenshot: file:/Development/java/projects/projects-test/selenide/build/reports/tests/1455265141846.0.png
Timeout: 4 s.
	at com.codeborne.selenide.impl.WebElementSource.checkCondition(WebElementSource.java:66)
	at com.codeborne.selenide.commands.Should.should(Should.java:35)
	[...]

Testing the title could serve as a quicktest that a link target is valid without relying on a changing page content.

vinogradoff:
You are right, good or bad - <title> is considered "invisible" by Selenium.
So it's reasonable question for me.

Just curious does
assertEquals("Your Title", title()); 
ever fails in your project, because title is tampered by ajax?

RobertKielty:
Hi @asolntsev HOpe this finds you well. Here's a Selenide use case for checking the content of a title of a page:

[...]

PR #1478, opened 2021-05-29
Conditional wait for non WebElement entities

dbudim:
## Proposed changes
Added conditional wait for non WebElement entities like LocalStorage. Now we can wait for some condition by Predicate.

## Checklist


src/main/java/com/codeborne/selenide/ConditionWaiter.java
```diff
@@ -0,0 +1,42 @@
...
+  private static final Logger logger = LoggerFactory.getLogger(ConditionWaiter.class);
+
+  public static void waitFor(Conditional conditional, Predicate predicate, String message) {
+    waitFor(conditional, predicate, message, 10);
+  }
+
+  public static void waitFor(Conditional conditional, Predicate predicate, String message, int timeout) {
+    try {
+      await()
+        .pollInterval(1, SECONDS)
```
rosolko, line 24: In any way this case should be configurable.
You can reload method with default value, but add an ability to set this parameter with custom value as well as `timeout`.

src/main/java/com/codeborne/selenide/ConditionWaiter.java
```diff
@@ -0,0 +1,42 @@
...
+        .timeout(timeout, SECONDS)
+        .pollInSameThread()
+        .until(() -> checkUnThrowable(conditional, predicate));
+    } catch (ConditionTimeoutException e) {
+      throw new ConditionTimeoutException(message);
```
rosolko, line 29: Re-wrap this exception with selenide one.

src/main/java/com/codeborne/selenide/ConditionWaiter.java
```diff
@@ -0,0 +1,42 @@
...
+    }
+  }
+
+  private static boolean checkUnThrowable(Conditional conditional, Predicate<Conditional> predicate) {
+    try {
+      return predicate.test(conditional);
+    } catch (Exception e) {
+      logger.info("Fail to check condition." + System.lineSeparator() + e);
```
rosolko, line 37:
```suggestion
      logger.info("Fail to check condition", e);
```

rosolko (changes requested):

src/main/java/com/codeborne/selenide/ConditionWaiter.java
dbudim: The most appropriate that i found was SoftAssertionError, i used it. 
But i think we can create new on like ConditionNotMatchException)

src/main/java/com/codeborne/selenide/Conditional.java
```diff
@@ -6,6 +6,6 @@
 
   void should(Predicate<T> predicate, String message);
 
-  void should(Predicate<T> predicate, String message, int timeout);
+  void should(Predicate<T> predicate, String message, long timeout);
```
rosolko, line 9: I don't think that `int`/`long` meet selenide api.
For example methods with any sort of wait such `shouldBe` or `shouldBe` using `Duration` type of `timeout`.
Also we already deprecate methods `waitUntil` or `waitWhile` with `long` type `timeout` parameter.
Is is possible to use `Duration` instead of `long` in awaility library methods

src/main/java/com/codeborne/selenide/Conditional.java
```diff
@@ -0,0 +1,12 @@
+package com.codeborne.selenide;
+
+import java.time.Duration;
+import java.util.function.Predicate;
+
+public interface Conditional<T> {
+
+  void should(Predicate<T> predicate, String message);
+
+  void should(Predicate<T> predicate, String message, Duration timeout);
```
rosolko, line 10: Look like you're missing condition with `polling`

src/main/java/com/codeborne/selenide/LocalStorage.java
```diff
@@ -1,10 +1,24 @@
...
+
+  @Override
+  public void should(Predicate<LocalStorage> predicate, String message) {
+    ConditionWaiter.waitFor(this, predicate, message);
+  }
+
+  @Override
+  public void should(Predicate<LocalStorage> predicate, String message, Duration timeout) {
+    ConditionWaiter.waitFor(this, predicate, message, timeout);
+  }
```
rosolko, line 23: ditto

src/main/java/com/codeborne/selenide/LocalStorageConditions.java
```diff
@@ -0,0 +1,15 @@
...
+
+  public static Predicate<LocalStorage> itemPresent(String item) {
+    return localStorage -> localStorage.containsItem(item);
+  }
+
+  public static Predicate<LocalStorage> itemHasValue(String item, String value) {
+    return localStorage -> localStorage.getItem(item).equals(value);
+  }
+
+}
```
rosolko, line 15: ditto

src/main/java/com/codeborne/selenide/Conditional.java
dbudim: I thought that polling is not so frequent configurable but if you say, i'll do it

src/main/java/com/codeborne/selenide/LocalStorageConditions.java
dbudim: Here is no polling, only predicate. What are you mean?

rosolko: Opps, miss comment

src/main/java/com/codeborne/selenide/Conditional.java
rosolko: There is no sense to make ability to configure pooling and not extend  interface with it

BorisOsipov:
@dbudim Thank you for help with this feature. That's awesome.

I have only one concern - you've added `awatility` deps for this.
We already have the same retry feature in https://github.com/selenide/selenide/blob/master/src/main/java/com/codeborne/selenide/impl/SelenideElementProxy.java#L130 with simple `do..while`. What do you think about to avoid adding `awatility` deps and make it works with our own retry impl?

src/main/java/com/codeborne/selenide/ConditionWaiter.java
BorisOsipov: yes we have to create new exception like it is for element conditions.

dbudim: aded

BorisOsipov:
>While starting work on this PR i've asked about this moment here and we agreed that awaitility can be used so i've used it)

Sure you can. My post is not about to restrict\allow smth. It is my opinion :)

BorisOsipov:
One thing we forget. Will we take screen\pagesource on fail?

Users say that is a must - https://github.com/selenide/selenide/issues/280#issuecomment-858523685

dbudim:
> One thing we forget. Will we take screen\pagesource on fail?
> 
> Users say that is a must - [#280 (comment)](https://github.com/selenide/selenide/issues/280#issuecomment-858523685)

I've checked) Yes)
image

BorisOsipov:
>I've checked) Yes)

nope) it is screenshots from https://github.com/selenide/selenide/blob/f9553715609644a74f99c6643afe91371219ed8f/statics/src/test/java/integration/IntegrationTest.java#L34 junit extension.

Also it will be nice to embed screen\source path to exception as we do for element condition

dbudim:
> > I've checked) Yes)
> 
> nope) it is screenshots from
> 
> https://github.com/selenide/selenide/blob/f9553715609644a74f99c6643afe91371219ed8f/statics/src/test/java/integration/IntegrationTest.java#L34
> 
> junit extension.
> Also it will be nice to embed screen\source path to exception as we do for element condition

You mean 

> > I've checked) Yes)
> 
> nope) it is screenshots from
> 
> https://github.com/selenide/selenide/blob/f9553715609644a74f99c6643afe91371219ed8f/statics/src/test/java/integration/IntegrationTest.java#L34
> 
> junit extension.
> Also it will be nice to embed screen\source path to exception as we do for element condition

Am i on right way?) 
I used in `UIAssertionError.wrap` as like in `ElementsCollection`)
Now Exception looks like:
image

But for it we need to put Driver into ConditionWaiter, so now usage will be `new ConditionWaiter(driver).waitFor();`


Is it ok?)

BorisOsipov:
>Am i on right way?)
>I used in UIAssertionError.wrap as like in ElementsCollection)

I guess yes

src/main/java/com/codeborne/selenide/LocalStorageConditions.java
```diff
@@ -0,0 +1,15 @@
+package com.codeborne.selenide;
+
+import java.util.function.Predicate;
+
+public class LocalStorageConditions {
+
+  public static Predicate<LocalStorage> itemPresent(String item) {
```
bereg2k, line 7: For the sake of readability, it'd better rename it to `containItem()` or `item()` so that the user's code would read like this:

`localStorage().should(containItem("mouse"),"Item 'mouse' doesn't exist")`

`localStorage().shouldHave(item("mouse"),"Item 'mouse' doesn't exist")`

src/main/java/com/codeborne/selenide/LocalStorageConditions.java
```diff
@@ -0,0 +1,15 @@
...
+    return localStorage -> localStorage.containsItem(item);
+  }
+
+  public static Predicate<LocalStorage> itemHasValue(String item, String value) {
```
bereg2k, line 11: For the sake of readability, it'd better rename it to `itemWithValue()` so that the user's code would read like this:

`localStorage().shouldHave(itemWithValue("mouse", "Jerry"),"Item 'mouse' value doesn't match")`

src/main/java/com/codeborne/selenide/ConditionWaiter.java
```diff
@@ -0,0 +1,56 @@
...
+
+import java.time.Duration;
+import java.util.function.Predicate;
+
+import static org.awaitility.Awaitility.await;
+
+/**
+ * Class provide conditional polling wait by predicate with ignore exceptions.
+ */
+
```
bereg2k, line 17: is it ok that there's an empty line between JavaDoc and the class's declaration?

dbudim: removed blank line

src/main/java/com/codeborne/selenide/LocalStorageConditions.java
dbudim: you right)
renamed itemPresent -> containItem

src/main/java/com/codeborne/selenide/LocalStorageConditions.java
dbudim: not sure. We have only `should(predicate)` method method,` so should(itemWithValue)` looks not pretty)

src/main/java/com/codeborne/selenide/LocalStorageConditions.java
bereg2k: `containItem()`. 

`containsItem()` reads grammatically incorrect if you consider it as a part of *"should containS item"*.

src/main/java/com/codeborne/selenide/LocalStorageConditions.java
bereg2k: If we can only utilize "should", then I'm suggesting `should(containItemWithValue(...))` or something like that. It will also look good next to the suggested above `should(containItem(...))`.

dbudim: Thanks) looks nice) 
done)

src/main/java/com/codeborne/selenide/LocalStorageConditions.java
dbudim: sure?
what about `containsKey` in Map?

dbudim: You right, i followed `containsKey` in `Map` convention, but in case of `should` it looks better)
done

dbudim:
@asolntsev @dstekanov Waiting for your review)

asolntsev (changes requested):
1. It seems a better English name for `ConditionNotMatchException` is `ConditionNotMetException` or `ConditionNotFulfilled`. See https://stackoverflow.com/questions/63751794/what-does-the-condition-is-met-exactly-mean-in-programming-languages
2. We still have hard-coded polling interval 500ms in `ConditionWaiter`. Should take it from `driver.config().pollingInterval()`.
3. Do we really need `Awaitility` dependency in `ConditionWaiter`? It seems to be a simple do..while loop.
4. I have found Selenide already has class very similar to `ConditionWaiter`: `com.codeborne.selenide.impl.Waiter`. 
  * It doesn't catch exceptions. I think we can add try/catch to `Waiter`.

src/main/java/com/codeborne/selenide/ConditionWaiter.java
```diff
@@ -0,0 +1,46 @@
...
+    try {
+      await()
+        .pollInterval(polling)
+        .timeout(timeout)
+        .pollInSameThread()
+        .until(() -> checkUnThrowable(conditional, predicate));
+    } catch (ConditionTimeoutException e) {
+      throw new SoftAssertionError(message);
+    }
+  }
```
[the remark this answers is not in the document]
dbudim: We have discussed it few times before)
I think that awaitility is more convenient and smooth rather than making own "bicycle" with do while)

src/main/java/com/codeborne/selenide/Conditional.java
```diff
@@ -0,0 +1,14 @@
+package com.codeborne.selenide;
+
+import java.time.Duration;
+import java.util.function.Predicate;
+
+public interface Conditional<T> {
```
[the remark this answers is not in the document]
dbudim: Not sure)
As for many programmers choosing good name is challenging for me)

What about `Awaitable` or `CanWaitCondition` ?)

dbudim:
> 1. It seems a better English name for `ConditionNotMatchException` is `ConditionNotMetException` or `ConditionNotFulfilled`. See https://stackoverflow.com/questions/63751794/what-does-the-condition-is-met-exactly-mean-in-programming-languages
agree, will make

> 2. We still have hard-coded polling interval 500ms in `ConditionWaiter`. Should take it from `driver.config().pollingInterval()`.
agree will add

> 3. Do we really need `Awaitility` dependency in `ConditionWaiter`? It seems to be a simple do..while loop.
I prefer to use Awaitility it more convenient and smooth ) but if you say i'll make it with do while

> 4. I have found Selenide already has class very similar to `ConditionWaiter`: `com.codeborne.selenide.impl.Waiter`.
> 
> * It doesn't catch exceptions. I think we can add try/catch to `Waiter`.
Do you mean add to existed methods or add waitUnthrowable?

src/main/java/com/codeborne/selenide/ConditionWaiter.java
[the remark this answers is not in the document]
BorisOsipov: @dbudim 

Dmitriy let's remove awaitlity. At least two mainteiners have concerns about it. Simple solution like `do..while` or `com.codeborne.selenide.impl.Waiter` are better than new dependency :)

BorisOsipov:
@dbudim another thing I realized today we forgot about events for reporters.
Commands and asserts execution produces events that reporter like allure uses to enrich in report.

for example, for element it done here
https://github.com/selenide/selenide/blob/b8b91aea105a754e611111fab144ed3f5d436568/src/main/java/com/codeborne/selenide/impl/SelenideElementProxy.java#L83

it will be nice if it works for not element asserts too.

src/main/java/com/codeborne/selenide/ConditionWaiter.java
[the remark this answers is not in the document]
dbudim: Okay) 
We can improve `Waiter` class, as i see it just waits and doesn't throw any exception?
should we add separate method which will be throw if condition not met until timeout? Or we can modify existed method?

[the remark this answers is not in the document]
asolntsev: It seems we can modify the existing method. I am surprised it doesn't throw an exception. :)

src/main/java/com/codeborne/selenide/Conditional.java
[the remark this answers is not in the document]
asolntsev: Then Conditional is better. :)
One more option is `Subject`, but I am not sure if it's better.

dbudim:
> @dbudim another thing I realized today we forgot about events for reporters.
> Commands and asserts execution produces events that reporter like allure uses to enrich in report.
> 
> for example, for element it done here
> https://github.com/selenide/selenide/blob/b8b91aea105a754e611111fab144ed3f5d436568/src/main/java/com/codeborne/selenide/impl/SelenideElementProxy.java#L83
> 
> it will be nice if it works for not element asserts too.

You right)
Agreed with @asolntsev that he will make it)

dbudim:
@asolntsev @BorisOsipov 
I removed awaitility))

Used `Waiter` class, but for getting default values from config and making screenshots it is needed to put `Driver` into `Waiter`.
I checked usages and didn't see ability to get driver in that places, so i decided do not change constructor and just create separate `wait` methods which puts Driver as argument.
Also i modified check predicate for unThrowable check)

Please let me know what you think about it)

asolntsev (approved):

rosolko (approved):
Looks pretty!

PR #1483, opened 2021-06-07
add getSessionId method to Driver

petroOv-PDFfiller:
## Proposed changes
I faced a bug with selenide-selenoid. In case when added listener to driver(class which implements WebDriverEventListener).  
Method getWebDriver() returns instance of EventFiringWebDriver which cannot be casted to RemoteWebDriver. It produce ClassCastException [here](https://github.com/selenide/selenide-selenoid/blob/2758043b06b90407ab5f65103a1caeee1173a0da/src/main/java/org/selenide/selenoid/DownloadFileInSelenoid.java#L55) and in the other places of selenide-selenoid project

## Checklist

### DIFF
--- a/src/main/java/com/codeborne/selenide/Driver.java
+++ b/src/main/java/com/codeborne/selenide/Driver.java
@@ -4,6 +4,9 @@ import com.codeborne.selenide.proxy.SelenideProxyServer;
 import org.openqa.selenium.JavascriptExecutor;
 import org.openqa.selenium.WebDriver;
 import org.openqa.selenium.interactions.Actions;
+import org.openqa.selenium.remote.RemoteWebDriver;
+import org.openqa.selenium.remote.SessionId;
+import org.openqa.selenium.support.events.EventFiringWebDriver;
 
 import javax.annotation.CheckReturnValue;
 import javax.annotation.Nonnull;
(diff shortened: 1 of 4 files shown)


src/main/java/com/codeborne/selenide/Driver.java
```diff
@@ -95,4 +98,17 @@ default SelenideTargetLocator switchTo() {
...
+  @Nullable
+  default SessionId getSessionId() {
+    WebDriver driver = getWebDriver();
+    if (getWebDriver() instanceof EventFiringWebDriver) {
+      driver = ((EventFiringWebDriver) getWebDriver()).getWrappedDriver();
+    }
+    if (driver instanceof RemoteWebDriver) {
+      return ((RemoteWebDriver) driver).getSessionId();
+    }
+    return null;
```
BorisOsipov, line 112: As I know all drivers implements `RemoteWebDriver` interface, so instead of return `null` it should always return `((RemoteWebDriver) driver).getSessionId()`, isn't it?

BorisOsipov (changes requested):

src/main/java/com/codeborne/selenide/Driver.java
BorisOsipov: I mean it is impossible by webdriver protocol to have driver implementations that don't have `sesisonId`

petroOv-PDFfiller: Yes you are right. fixed

BorisOsipov (approved):
LGTM. Thank you!

PR #1488, opened 2021-06-22
add support for okhttp 4.9.1

asolntsev:
* in okhttp 3, fields are named "connectTimeout" and "readTimeout"
* in okhttp 4, fields are named "connectTimeoutMillis" and "readTimeoutMillis"


rosolko (approved):
Looks good

Issue #1500, opened 2021-07-05
Improve switchTo().innerFrame() API


cbxp:
## The problem
* `innerFrame(vararg)` actually compiles without any args: `innerFrame()` - the expectation being that it could switch to the first (or only) available frame, on the page, but in reality it does nothing, not even an exception
*  `innerFrame()` accept frame id or names a parameter, which is inconsistent with `$()`, which accepts selectors. It would be nice to make two versions: 1. With instances of FrameByIdOrName 2. With string selectors. Maybe also rename the method to `frame()` and deprecate the old one because of behaviour change.
* Maybe switchTo() could also be a singleton constant, so that parentheses would not be needed?

## Details
If necessary, describe the problem you have been experiencing in more detail.

## Tell us about your environment
* **Selenide Version**: 5.22.2

asolntsev:
implemented in PR #1509

PR #1509, opened 2021-07-24
#1500 Improve switchTo().innerFrame() API

asolntsev:


src/main/java/com/codeborne/selenide/SelenideTargetLocator.java
```diff
@@ -134,13 +135,19 @@ public Alert alert() {
   }
 
   /**
-   * Switch to the inner frame (last child frame in given sequence)
+   * Switch to the inner frame (last child frame in given sequence).
+   * <p>
+   * This method
+   * <ol>
+   *  <li> first switches to the root frame (aka "default content"),</li>
```
bereg2k, line 142: - For the ordered list it's unnecessary to use "first...", "then..."
- `firstFame` -> `firstFrame`, `otherFrames` -> `nextFrames`
- JavaDoc doesn't recognize backticks, it's better to use `<code>`, `{@code}`, or simply regular `""`
- `3. consecutively switches to "nextFrames" until it reaches the last one`

asolntsev: thanks, fixed.

Issue #1508, opened 2021-07-19
Implement `ObjectCondition<WebDriver>` for the "number of windows" condition


bereg2k:
Implement ObjectCondition<WebDriver> for the number of opened windows.

## Details
One of the rather useful expected conditions in Selenium's native library `ExpectedConditions` is `numberOfWindowsToBe(int)`. It's possible to call it using Selenide like this:

`Selenide.Wait().withTimeout(ofSeconds(60)).until(numberOfWindowsToBe(2))`

It works well but in light of [the recent feature](https://github.com/selenide/selenide/pull/1478), it's better to use it via new approach. It adds reporting, built-in waits, etc. 

Possible usage:

```java
webdriver().shouldHave(numberOfWindows(2));
webdriver().shouldHave(numberOfWindows(2), ofSeconds(60));
```

Possible implementation:
- `NumberOfWindows` class that implements `ObjectCondition<WebDriver>`
- add `numberOfWindows` method to `WebDriverConditions`

bereg2k:
I'd like to provide this feature if it's considered useful.

asolntsev:
Sure, we can do it. 

I am not sure it's so much useful because typically it's enough to just switch to the new window:
```java
$("button").click();  // click opens a new tab
switchTo(window(1)); // wait for the new tab and switches to it.
```

But probably there might be cases...

bereg2k:
It's a pretty ordinary thing on my current project. It's useful in E2E scenarios with integrations. 

1. User presses a button on a page.
2. User sees a spinner/loading dialog. 
3. The browser automatically opens the 2nd tab to continue the scenario there (normally, it's some remote service). That's for positive cases. 

This becomes useful because we can't tell how much time does the system need to send a user to the 2nd tab:
```java
$("button").click(); // click opens a new tab
Selenide.Wait().withTimeout(ofSeconds(60)).until(numberOfWindowsToBe(2)); 
switchTo(window(1)); // it's only reasonable to switch when the the 2nd tab is actually opened 
```
After all, `numberOfWindowsToBe` condition is in Selenium library for a reason.

Considering that Selenide is actively used for integration/e2e testing, this seems useful for such cases.

bereg2k:
@asolntsev I've looked into `com.codeborne.selenide.SelenideTargetLocator`...

Is it possible to have the same functionality with this:

`switchTo(window(1), ofSeconds(60));`

If so, then this feature request loses most of its value :)

asolntsev:
[...]

PR #1511, opened 2021-07-25
#1508 `numberOfWindows` condition for `webdriver()`

bereg2k:
## Proposed changes
The feature similar to `FluentWait` condition `numberOfWindowsToBe()`.

```java
webdriver().shouldHave(numberOfTabs(2));
webdriver().shouldHave(numberOfTabs(2), ofSeconds(60));
```

### Note
@asolntsev I'm not so sure about the naming here. `numberOfWindows` is closer to the original condition in Selenium, but it's a bit misleading (we're dealing with tabs of the browser; "windows" can be attributed to the several browsers/browser's windows) and long. `tabs()` is a bit short and doesn't sound that grammatically correct ("web driver should have tabs...") or accurate. That's why I went with `numberOfTabs`, but I'm open to suggestions.

## Checklist

### DIFF
--- a/src/main/java/com/codeborne/selenide/WebDriverConditions.java
+++ b/src/main/java/com/codeborne/selenide/WebDriverConditions.java
@@ -1,11 +1,6 @@
 package com.codeborne.selenide;
 
-import com.codeborne.selenide.conditions.webdriver.CurrentFrameUrl;
-import com.codeborne.selenide.conditions.webdriver.CurrentFrameUrlContaining;
-import com.codeborne.selenide.conditions.webdriver.CurrentFrameUrlStartingWith;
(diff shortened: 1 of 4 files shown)


asolntsev:
@bereg2k There is almost no difference between "tabs" and "windows", at least in Selenium. 
* Every tab can be "moved" to a separate window and vice versa. 
* For webdriver, term "window" means both "windows" and "tabs", no difference. 
* Method `Selenide.closeWindow()` closes the active tab.
 
So it's probably better to use term "window" as it's closer to the original Selenium terminology.

One alternative for `numberOfFoos` is `foosCount`.

bereg2k:
@asolntsev renamed all "tabs" references to "windows".

asolntsev (approved):

Issue #1525, opened 2021-08-18
Add execute(Command, Duration) method to SelenideElement


evpl:
Hi. I faced a problem when i tried to execute custom wait command with custom timeout. I checked with `InvokationHandler` that it will work well if you just add method `execute(Command, Duration)` to `SelenideElement`.

Optionally, but it will great if you add opportunity to set custom polling interval like `execute(Command, Duration, Duration)` or `execute(Command, Duration, long)`. But in this case you need to edit `SelenideElementProxy`.

asolntsev:
@evpl Good point! We definitely need `execute(Command, Duration)`. 

But can you explain the real need for custom polling interval?

evpl:
This is just a guess. Users with custom commands will probably need to set a custom polling interval :)
For example, to exclude spam clicking (or another command execution) on an element which may break by this actions.

evpl:
I am working on it (`execute(Command, Duration)` method)

PR #1531, opened 2021-08-20
#1525 add SelenideElement execute(Command, Duration) method

evpl:
## Proposed changes
Issue #1525 - Add execute(Command, Duration) method to SelenideElement

## Checklist


asolntsev (changes requested):
I don't see where the timeout is used. Am I blind?..

evpl:
@asolntsev it's not easy to understand how the tests should look :)
I made a fix

Issue #1074, opened 2020-02-28
Problem with fastSetValue on Material UI Components 


rusakovStas:
## The problem
Material UI is one of the most popular React Components library. 
And when i'm trying to use Selenide with fastSetValues (setting values via js) it's not working correctly. 

## Details
When Selenide sets value via JS Material UI component ( Autocomplete ) do not call hooks like onInputChange and others. 

## Tell us about your environment
* **Selenide Version**: 5.8.0
* **Yandex browser**: 20.2.0.1145
* **Browser Driver Version**: 79
* **Selenium Version**: embedded in Selenide
* **OS Version**: mac OS

## Code To Reproduce Issue [ Good To Have ]
To reproduce the bug you may run the test: 
```
@Test
    void check() {
        Configuration.fastSetValue = true;
        open("https://rusakovstas.github.io/material-autocomplete/");
        $(byText("Вбейте что нибудь или скопируйте")).closest("div").$("input").setValue("123");
        $(byText("Поймали логику!")).shouldBe(visible);
    }
```
For now it's failed. If you remove Configuration.fastSetValue = true; it works fine.

rusakovStas:
Hi! 
I've finally found the solution of this question! It's really simple) 
To set value via js properly for Material UI you have to do this by `setAttribute('value', 'some text'); `
Not just `element.value = `
I think the reason in some JS Proxy or something like that.

I even have changed the method setValue in Selenide and it's working (after local install)! 
But now: 
1. i can't push branch (getting 403). Maybe there are some rules for that? 
2. i'm not sure how can i write the test for this? 

Just in case code for fix is pretty simple: 

```
private String setValueByJs(Driver driver, WebElement element, String text) {
    return driver.executeJavaScript(
        "return (function(webelement, text) {" +
            "if (webelement.getAttribute('readonly') != undefined) return 'Cannot change value of readonly element';" +
            "if (webelement.getAttribute('disabled') != undefined) return 'Cannot change value of disabled element';" +
            "webelement.focus();" +
            "var maxlength = webelement.getAttribute('maxlength') == null ? -1 : parseInt(webelement.getAttribute('maxlength'));" +
            "webelement.setAttribute('value', " +
            "maxlength == -1 ? text " +
            ": text.length <= maxlength ? text " +
            ": text.substring(0, maxlength));" +
            "return null;" +
            "})(arguments[0], arguments[1]);",
        element, text);
  }
```

rusakovStas:
[...]

PR #1537, opened 2021-08-26
#1074 fix fastSetValue for material design

asolntsev:
fix for https://github.com/selenide/selenide/issues/1074

let's try https://github.com/selenide/selenide/pull/1076 once again :)

NB! It seems to be a bad idea: setting `value` property is the right way, while changing attribute `value` is the wrong.


asolntsev:
Again, let's close this PR. :)
* In theory, setting `value` _property_ is the right way, while changing _attribute_ `value` is the wrong way.
* In practice, this PR didn't work for TextAreas.

Issue #980, opened 2019-10-02
Add current URL on the error message


gtiwari333:
## The problem
The error message on a test failure doesn't display the URL where we are getting the error. 

## Details

The following message is helpful. It would be great to add the URL as well.

```
com.codeborne.selenide.ex.ElementNotFound: Element not found {#dashboardButton}
Expected: visible or transparent: visible or have css value opacity=0
Screenshot: file:/C:/_work/app/build/reports/tests/1570048960084.0.png
Page source: file:/C:/_work/app/build/reports/tests/1570048960084.0.html
Timeout: 4 s.
Caused by: NoSuchElementException: no such element: Unable to locate element: {"method":"css selector","selector":"#dashboardButton"}

	at com.codeborne.selenide.impl.WebElementSource.createElementNotFoundError(WebElementSource.java:37)
	at com.codeborne.selenide.impl.ElementFinder.createElementNotFoundError(ElementFinder.java:100)
	at com.codeborne.selenide.impl.WebElementSource.checkCondition(WebElementSource.java:65)

```
 
## Possible Solution:

We may need to update com.codeborne.selenide.ex.UIAssertionError#uiDetails() and  com.codeborne.selenide.ex.ErrorMessages to pull the current url from driver object:

`driver.getWebDriver().getCurrentUrl();`

asolntsev:
@gtiwari333 Why not, good idea. Let's do it!

BorisOsipov:
Should be done carefully. Url can be quite large and in this case it will make error message hideous

asolntsev:
Good point. Probably some substring could help?

BorisOsipov:
Ok lets think.

> Scenario:
> As software test automation I want to find root cause of failed test quicker.
> 
> Given I see current app url in stacktrace
> Then I open url?
> Or I attach to bug report?

I don't know for what we need url in stacktrace.

Pros
1) Common reason - make failure message more rich.
2) Non technical team members can easily open app and compare sceen from  error and page itself (and don't open test code lol)
3) Can quickly open application from stacktrace I look at what is wrong.

Cons
1) For modern SPA application it is useless. They often use one route for serving bunch of pages.
2) For enterprise apps url can be very large enormous and will make error hideous(or if we trim url we won't have any cons I mentioned above)
3) Often you can just open url and server returns corresponding page... So for what we need such url?
[...]

PR #1538, opened 2021-08-26
#980 add current page URL to the error message in case of test failure

asolntsev:
implementation of https://github.com/selenide/selenide/issues/980

NB! We need to make it configurable (pluggable): in many projects this URL is not needed


asolntsev:
supeseded by https://github.com/selenide/selenide/pull/2097

PR #1552, opened 2021-09-03
Proposed collection condition: check if collection elements contain / do not contain given attribute with value

plagov:
Proposing a new collection condition that asserts that all collection elements contain or do not contain a given attribute with value.
There's already a condition `CollectionCondition.allMatch()` and `noneMatch()`, but I see two downsides with it. First, it accepts the `Predicate<WebElement>` and it might not be easy and clear for many users how to use it. Second, when the above conditions fail, the error message is generic, it says `Collection matcher error` and doesn't give information on the predicate that was applied. That is understandable, why it doesn't provide more info - because you can't extract more information from the predicate.

With the proposed new condition - the usage would be much simpler, the user would need just to pass in an attribute name and value that all collection elements are expected to have. And it would be possible to specify a more descriptive and informative error message under the hood with both the attribute and name that are expected but not met.

@asolntsev what do you think about this? Is it needed or is unnecessary?
I have quickly created a rough prototype in this PR just to initiate a discussion. Not ready for review yet. Just a draft, so that we all won't waste time if the decision would be not doing it.

### DIFF
--- a/src/main/java/com/codeborne/selenide/CollectionCondition.java
+++ b/src/main/java/com/codeborne/selenide/CollectionCondition.java
@@ -2,6 +2,7 @@ package com.codeborne.selenide;
 
 import com.codeborne.selenide.collections.AllMatch;
 import com.codeborne.selenide.collections.AnyMatch;
+import com.codeborne.selenide.collections.Attribute;
 import com.codeborne.selenide.collections.ContainExactTextsCaseSensitive;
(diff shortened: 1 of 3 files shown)


asolntsev:
@plagov I would rather suggest to improve `allMatch` method. Can we modify it to throw the proper error with readable error message etc.?

plagov:
@asolntsev I think you're right and it's really not needed. I think we can close/reject it.

PR #1554, opened 2021-09-07
Enhance `closest()` and `ancestor()` methods to search by attribute and attribute w/ value

plagov:
## Proposed changes
Enhance `closest()` and `ancestor()` methods to search by attribute and attribute w/ value.
There's an existing command `closest` that will locate an ancestor with a given tag or class. That's not enough in some cases. With modern UI frameworks (React, for example), classes are often autogenerated and the same class appears on many elements. For example:

```html
<div test-attribute="test-value" class="React_class_zxcvbn">
  <div test-attribute="test-value" id="container" class="React_class_qwerty">
    <div>
      <div id="first-div">First div element</div>
      <div id="second-div">Second div element</div>
    </div>
  </div>
</div>
```

If I want to find a container element of the `#first-div` element, then `closest()` won't help. The container element doesn't have a unique and reliable class name to search for. The tag name is also not unique, all elements are divs, so the `closest()` will find the first `div`.

Thus, enhanced `closest()` or `ancestor()` commands will also search for an ancestor by the attribute and attribute's value in addition to an existing functionality to search by the tag or CSS class.

## Checklist


asolntsev:
I understand people do need to find elements that are hard to find :) 

But it seems to me that there might be a plenty of such methods. In the end, it will be hard for people to find the right method among all these "find*", "closest*", "nearest*", "parent*", "grandparent*", "sibling*" etc. In the end, is it really simpler than just using xpath?

Compare:

1. `$("#first-div"). closestWithAttributeAndValue("test-attribute", "test-value")`
2. `$x("//*[@id='first-div']//anccesstor::*[@test-atribute='test-value']")`

The xpath implementation is even shorter. :) 
And if you write a wrong xpath, IDEA will highlight it. 

I don't dislike this feature, just not sure. Open for discussions. 

P.S. Probably we should just introduce some nice API for building XPath in Java code?

plagov:
@asolntsev I understand your concern. New commands and conditions indeed may get lost when there are many of them
My motivation for this was to simplify working with XPath. If the one doesn't construct XPaths that often in his project, then it will take him some time to make up the one that will find the target element. XPath's syntax is not that obvious, so it might be confusing whether it should have a `//` or `/` or `./` when I call a `find(By.xpath)` on the element and so on...

vinogradoff:
Thank you for PR!

What about this. Could you instead of adding two new methods to API, just extend the functionality the closest(String) method in this way:

closest("[test-attribute]") - locates the first up-tree element containing the attribute  
closest("[test-attribute=test-value]") - locates the first up-tree element containing the attribute with value

I thinks @asolntsev will approve this solution.

plagov:
@vinogradoff @asolntsev yes, I can extend the existing `closest()` as you propose. 
The very first idea of how to do it is to check if the string starts with `[` and doesn't contain `=`, then it is just the attribute. If it does contain the `=`, then everything after that is a value. But I see two concerns here. What if the attribute name will contain the `=`? Yes, it might be weird to have an attribute containing a `=` character, but who knows? :) 
And second, if the user will forget to surround the attribute with square brackets? Selenide will think it's a regular tag.
Also, not sure about the value wrapped in single or double quotes...

What do you think?

vinogradoff:
I don't think "="
allowed in attribute names. If user forgets to close - nothing will be found, right? So no problem too. If you throws an exception
In case attribute is malformed - its okay too

plagov:
@vinogradoff ok, I will update this PR and will submit the updated implementation.

plagov:
@asolntsev @vinogradoff ready for review. Changed as per Alexei's comments to enhance the existing `closest()` method to handle attribute and attribute value. I decided to implement this using the rule engine pattern. Because there are 4 conditions (tag, class, attr, attr w/ value), then the straightforward approach with if-else wouldn't look good. With the rule engine it looks clean and would be very easy to enhance the method further should we add another way to find the closest element.

asolntsev:
@plagov Ooops, it seems we have conflicts with master branch. Would it be hard to resolve those?

plagov:
@asolntsev I see that #1567 PR has updated the former GetClosest class and the current implementation is not compatible with my implementation. I will have to refactor my PR in order to incorporate new changes. But before doing this, could you please review the current implementation and give some feedback on whether it's good or not and if it is suitable to go forward with it? If all good, then I will continue and will resolve the merge conflict.

Just would like to get some feedback and wouldn't want to waste time on future work if something is not OK already at this point :)

src/main/java/com/codeborne/selenide/commands/closest/ClosestRuleEngine.java
```diff
@@ -0,0 +1,28 @@
+package com.codeborne.selenide.commands.closest;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+
+public class ClosestRuleEngine {
+
+  private static final List<ClosestRule> rules = new ArrayList<>();
```
asolntsev, line 9: It may be simplified: just use `Arrays.asList(...)` instead of `new ArrayList<>()`

asolntsev (changes requested):
I like the approach, looks nice!

bereg2k:
@plagov sorry for the inconvenience.

plagov:
@asolntsev I've resolved all merge conflicts. Ready for review.

src/main/java/com/codeborne/selenide/commands/ancestor/AncestorWithAttributeAndValueRule.java
```diff
@@ -0,0 +1,25 @@
+package com.codeborne.selenide.commands.ancestor;
+
+import static java.lang.String.format;
+
+public class AncestorWithAttributeAndValueRule extends SelectorValidation implements AncestorRule {
+
+  private String xpath;
```
asolntsev, line 7: **Problem:**
`AncestorWithAttributeAndValueRule` holds the found `xpath` as a field which can be concurrently written by parallel threads (instance of `AncestorWithAttributeAndValueRule` itself is stored as a static variable in `AncestorRuleEngine`). 

**Proposed solution:**
Instead of holding the found `xpath` in a field and returning a `boolean`, method `evaluate` could **return**  `Optional<AncestorResult>`.

src/main/java/com/codeborne/selenide/commands/ancestor/AncestorWithAttributeAndValueRule.java
```diff
@@ -0,0 +1,25 @@
...
+
+  @Override
+  public boolean evaluate(String selector, int index) {
+    boolean evaluationResult = false;
+    if (isAttribute(selector) && containsAttributeValue(selector)) {
+      String attribute = selector.substring(1, selector.indexOf("="));
```
asolntsev, line 13: For better performance, calculate `selector.indexOf("=")` only once a put to a local variable.

asolntsev (changes requested):
A few more comments :)

src/main/java/com/codeborne/selenide/commands/ancestor/AncestorWithAttributeAndValueRule.java
plagov: I'm not sure I understood. I change `evaluate()` method in rules like this:

```java
  public Optional<AncestorResult> evaluate(String selector, int index) {
    if (isAttribute(selector) && containsAttributeValue(selector)) {
      int endIndex = selector.indexOf("=");
      String attribute = selector.substring(1, endIndex);
      String attributeValue = selector.substring(endIndex + 1, selector.length() - 1);
      this.xpath = format("ancestor::*[@%s='%s'][%s]", attribute, attributeValue, index);
    }
    return Optional.of(new AncestorResult(xpath));
  }
```

Then there rule engine class:

```java
  public AncestorResult process(String selector, int index) {
    AncestorRule ancestorRule = rules
      .stream()
      .filter(rule -> rule.evaluate(selector, index).isPresent())
      .findFirst()
      .orElseThrow(() -> new IllegalArgumentException("Selector does not match any rule"));
    return ancestorRule.getAncestorResult();
  }
```

But the `xpath` is `null`.
Could you please help me?

asolntsev: Yes, just remove field `xpath` and method `getAncestorResult()` from `AncestorRule`. Use only the value returned by method `evaluate`.

src/main/java/com/codeborne/selenide/commands/ancestor/AncestorWithAttributeAndValueRule.java
plagov: Thanks, updated.

src/main/java/com/codeborne/selenide/commands/ancestor/AncestorWithAttributeAndValueRule.java
plagov: Thanks. Updated.

asolntsev (approved):
LGTM!

Issue #1560, opened 2021-09-15
Remove @Deprecated on ElementsContainer.getSelf()


qky666:
## The problem
There are some use cases (IMO) where method `ElementsContainer.getSelf()` is quite useful.

## Details
For example: I have a page with a banner in it. I want to test that, if I click on "close" button, the Banner disappears. My code could be something like this:

```
public class MyPage {
    @FindBy(css = "div#some-banner") public BannerWidget banner;
    // Maybe lots of other SelenideElement and ElementsContainer.
    
    public void closeBanner() {
        banner.close.click();
        banner.getSelf().should(disappear);
    }
}

public class BannerWidget extends ElementsContainer {
    public SelenideElement close = $("button.close");
    // Maybe lots of other SelenideElement.
}
```

I know that there are many other ways to do the same thing, and many of these will not use `getSelf()`... But I find this one quite intuitive for me, and I really think that are some other use cases like this one where `getSelf()` is very useful.

Thanks!

asolntsev:
@qky666 Thank you for the realistic example of why this method can be useful. I thought nobody really uses it. :)

Well, we'll un-deprecate it.

PR #1564, opened 2021-09-17
Improve page objects

asolntsev:
implementation of https://github.com/selenide/selenide/issues/1560


asolntsev:
it broken `FieldOfGenericTypeTest`. Don't want to figure it out.

PR #1569, opened 2021-09-21
Added link to the correct dependency module (core)

vinogradoff:
You need to add browserup-proxy-core, not browserup-proxy dependency, so this link could be helpful to find the right dependency faster.

### DIFF
--- a/src/main/java/com/codeborne/selenide/drivercommands/CreateDriverCommand.java
+++ b/src/main/java/com/codeborne/selenide/drivercommands/CreateDriverCommand.java
@@ -58,7 +58,7 @@ public class CreateDriverCommand {
       }
       catch (NoClassDefFoundError e) {
         throw new IllegalStateException("Cannot initialize proxy. " +
-          "Probably you should add BrowserUpProxy dependency to your project.", e);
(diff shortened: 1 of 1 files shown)


vinogradoff:
stylecheck fix

asolntsev (approved):
The change is not bad, but it doesn't really solve the problem.
If one doesn't understand clause "add a dependency", he will also not understand the URL as well.

vinogradoff:
@asolntsev it was from my experience, I've read the old message, googled for browserupproxy, added dependecy to browserup-proxy. Didn't work:-). Googled in Selenide project- and recalled it must be not browserup-proxy, but browserup-proxy-core. 

So new message would save me time to get to -core immediately.

asolntsev:
@vinogradoff Probably it might be link to wiki post like https://github.com/selenide/selenide/wiki/Lazy-loading ?

vinogradoff:
@asolntsev yes, actually the best would be 1) create the wiki article about Selenide Proxy 2) link it at a few spot like this one.

this was a quick-fix, I create an issue reminder for wiki.

Issue #1577, opened 2021-09-23
Drag'n'Drop between frames


asolntsev:
## The problem

Hello, community.
I'm struggling to perform `dragAndDrop` where element to drag is in iframe.

All my experiments failed. 
As potential solution would be calculate point to drop, but **there must be an easier solution before I get to that**...

## Details

1. When trying to perform
```java
elementToDrop.dragAndDropTo(switchTo().frame($(By.cssSelector(".gjs-frame"))).findElement(By.cssSelector("body .layout-column-component--edit")));
```

result:
```
Element not found {$$(5 elements)[0]}

Caused by: StaleElementReferenceException: stale element reference: element is not attached to the page document
```

2. When use
```java
elementToDrop.dragAndDropTo("body .layout-column-component--edit", usingJavaScript());
```

result:
```
Caused by: NoSuchElementException: no such element: Unable to locate element: {"method":"css selector","selector":"body .layout-column-component--edit"}
```

3. When trying to use Actions(), it drops only where my mouse is (even if it's on 2nd screen).

## Tell us about your environment
* **Selenide Version**: 5.24.4

PR #1578, opened 2021-09-23
#1577 add code to reproduce the problem with Drag'n'Drop between frames

asolntsev:
Code to reproduce the issue https://github.com/selenide/selenide/issues/1577

(I still don't know how to fix the issue)


ervuks:
@asolntsev perhaps the issue is not in the framework itself, but just an edcase to perform?

asolntsev:
@ervuks It seems to be a limitation of webdriver.

Issue #1573, opened 2021-09-22
add check for title of current page


asolntsev:
I want to check the title using code like this:

```java
Selenide.webdriver().shouldHave(WebDriverConditions.title("All Developer Tools and Products by JetBrains"));
```

ervuks:
I can try to take this and make PR

PR #1579, opened 2021-09-23
#1573 - Ability to assert page title

ervuks:
## Proposed changes
Implemented feature, requested by https://github.com/selenide/selenide/issues/1573

## Checklist


asolntsev (changes requested):
1. Isn't it better to rename `title` to `titleContaining` because it check for the substring?
2. Don't we want to have two methods: `title` (for exact match) and `titleContaining` (for substring)?

ervuks:
Perhaps then just leave the `exact` option without the possibility to check against partial match?

asolntsev:
@ervuks yes, I like the exact match.

asolntsev (approved):

Issue #1572, opened 2021-09-22
Add click method with custom timeout


asolntsev:

evpl:
@asolntsev Hi,i'll do it with ClickOptions refactoring

evpl:
@asolntsev look at #1580 PR please. If it's ok, I will add tests.

PR #1580, opened 2021-09-26
#1572 click methods refactoring, click with given timeout

evpl:
From #1572 

## Proposed changes
- `com.codeborne.selenide.ClickOptions` refactoring (now it's an interface), the `com.codeborne.selenide.ClickMethod` enum has been removed
- `com.codeborne.selenide.commands.Click` command refactoring
- new `SelenideElement` methods with given timeout: `click(Duration)`, `click(int, int, Duration)`, `click(ClickOptions, Duration)`
- related changes
- tests

## Pros:
- code extensibility
- dependency inversion
- `usingJavaScript(111, 222)` is just shorter than `usingJavaScript().offsetX(111).offsetY(222)`
- the native way to execute command with a timeout

## Checklist


src/main/java/com/codeborne/selenide/commands/Click.java
```diff
@@ -15,76 +14,27 @@
...
+    clickOptions.click(driver, webElement);
     return null;
   }
-
-  protected void click(Driver driver, WebElement element) {
-    if (driver.config().clickViaJs()) {
-      click(driver, element, 0, 0);
-    }
-    else {
-      element.click();
```
dstekanov, line 46: Do we want to remove WebElement click()? Is it the same as Actions click()?

evpl: Fixed a bug by creating `ClickOptions.ZeroOffsetDefaultMethod` class

src/main/java/com/codeborne/selenide/commands/Click.java
```diff
@@ -4,87 +4,27 @@
...
-    }
-    else if (args.length == 2) {
-      Integer offsetX = firstOf(args);
-      Integer offsetY = (Integer) args[1];
-      click(driver, webElement, offsetX, offsetY);
+      ClickOptions.clickUsingConfigMethod(driver, webElement);
+    } else if (args[0] instanceof Integer) {
+      ClickOptions.clickUsingConfigMethod(driver, webElement, (Integer) args[0], (Integer) args[1]);
+    } else {
+      Util.<ClickOptions>firstOf(args).click(driver, webElement);
```
dstekanov, line 26: I guess it will fail when we call `void click(Duration timeout);`? If I am not mistaken, we need to ignore Duration inside Click command. Could you re-check it, please?

evpl: Yes, I fixed it with new mechanism `OverloadedMethodCommandBuilder`. I'm sure we can reuse it for other commands with method overloading.

asolntsev (changes requested):
Hi.
This PR seems over-engineered to me. 

I don't see why it's needed to split `ClickOptions` to interface/subclasses.

I don't see the need for `OverloadedMethodCommandBuilder` with its rules. I guess the intention was to create a "nice" code, which is nice. But unfortunately the resulting code seems more complicated and harder to understand. :(

PR #1595, opened 2021-10-07
Validate element when assert if `disabled` or `enabled`

plagov:
## Proposed changes
Add validation to check if the element is applicable for the disabled/enabled condition. The `Condition.enabled/disabled` is calling the Selenium's `WebElement.isEnabled()` that should be called on the `input` element. Without knowing that enabled/disabled should be called on the `input`, the Selenide's user can call this assertion on an any other element that has `disabled` attribute, and it will either fail with an ambiguous error message or even worse - will pass, which will be a false positive.
With this validation, the test will fail if called on a non-input element. Thus, the user will know exactly what went wrong.

For example, given an element:
```html
<label disabled id="disabled-label">Label with disabled attribute</label>
```
When asserting it with `$("#disabled-label").shouldBe(disabled)`, then it will fail with an ambiguous error message:
```
Element should be disabled {#disabled-label}
Element: '<label disabled id="disabled-label">Label with disabled attribute</label>'
Actual value: enabled
```
When asserting it with `$("#disabled-label").shouldBe(enabled)`, then the test will pass. Which is technically correct from Selenium's perspective. But likely is not correct from the user's experience who doesn't know that this assertion should be used only on `input`. 

In modern web frameworks, front end developers can set a `disabled` attribute on any element. So, the proposed change will help testers to use a disabled/enabled condition properly.

## Checklist

### DIFF
--- a/src/main/java/com/codeborne/selenide/Condition.java
+++ b/src/main/java/com/codeborne/selenide/Condition.java
@@ -457,14 +457,21 @@ public abstract class Condition {
   public static final Condition focused = new Focused();
 
   /**
-   * Checks that element is not disabled
+   * Checks that element is not disabled.
+   * <br>
+   * Will throw {@code IllegalArgumentException} if called on an element which is not {@code input}
    *
    * @see WebElement#isEnabled()
    */
   public static final Condition enabled = new Enabled();
 
   /**
-   * Checks that element is disabled
+   * Checks that element is disabled.
+   * <br>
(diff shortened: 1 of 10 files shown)


plagov:
@asolntsev @BorisOsipov I don't know how to do it. Given the logic specified in the HTML spec by the above link, I'm not even sure it should be done in Selenide.

asolntsev:
@plagov So what's next? Should we close this PR?

plagov:
@asolntsev yes, I think we can close this PR.

Issue #1593, opened 2021-10-06
5.25.0-selenium-4.0.0-rc-2 not working with browserup-proxy-core (dependency issues)


thomas-eaa:
## The problem
Using `selenide 5.25.0-selenium-4.0.0-rc-2` does not work in combination with `browserup-proxy-core 2.1.2` because of dependency conflicts (`selenium-api`, `netty`).

## Details
`selenide 5.25.0-selenium-4.0.0-rc-2` requires `selenium-api 4.0.0-rc2`, whereas `browserup-proxy-core  2.1.2` (latest version) requires `selenium-api 3.141.59`. Even after exclusion of the older dependency, issues with conflicting versions of `netty` dependencies appear and cannot be resolved:

`Caused by: java.lang.NoSuchMethodError: 'java.lang.String io.netty.util.internal.ObjectUtil.checkNonEmptyAfterTrim(java.lang.String, java.lang.String)'
	at io.netty.handler.codec.http.HttpMethod.<init>(HttpMethod.java:123)
	at io.netty.handler.codec.http.HttpMethod.<clinit>(HttpMethod.java:36)
	at org.asynchttpclient.netty.request.NettyRequestFactory.newNettyRequest(NettyRequestFactory.java:114)
	at org.asynchttpclient.netty.request.NettyRequestSender.newNettyRequestAndResponseFuture(NettyRequestSender.java:204)
	at org.asynchttpclient.netty.request.NettyRequestSender.sendRequestWithCertainForceConnect(NettyRequestSender.java:136)
	at org.asynchttpclient.netty.request.NettyRequestSender.sendRequest(NettyRequestSender.java:114)
	at org.asynchttpclient.DefaultAsyncHttpClient.execute(DefaultAsyncHttpClient.java:259)
	at org.asynchttpclient.DefaultAsyncHttpClient.executeRequest(DefaultAsyncHttpClient.java:228)
	at org.asynchttpclient.DefaultAsyncHttpClient.executeRequest(DefaultAsyncHttpClient.java:249)
	at org.openqa.selenium.remote.http.netty.NettyHttpHandler.makeCall(NettyHttpHandler.java:57)
	at org.openqa.selenium.remote.http.AddSeleniumUserAgent.lambda$apply$0(AddSeleniumUserAgent.java:42)
	at org.openqa.selenium.remote.http.Filter.lambda$andFinally$1(Filter.java:56)
[...]

PR #1612, opened 2021-10-20
#1593 Publish Selenide as separate artifacts

asolntsev:
* selenide.jar (the old good Selenide, incl. JUnit5 support)
* selenide-proxy.jar
* selenide-junit.jar (for JUnit4 support)
* selenide-testng.jar (for TestNG support)

It solves #1593 and allows using just `selenide-proxy` or `selenide-testng` dependency without any other dependencies.


asolntsev:
@BorisOsipov ping

rosolko (approved):
Nice, lets try this out!

PR #1625, opened 2021-10-27
avoid using Selenium 3* transitive dependencies

asolntsev:
Maven prefers an older "selenium-api-3*.jar" instead of a newer "selenium-api-4*.jar" because it's located "higher" in a dependency tree (because it's a transitive dependency of browserup proxy, allure etc.).

This commit is a workaround against this wrong Maven behaviour.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Changelog
 
+## 6.0.3 (released 27.10.2021)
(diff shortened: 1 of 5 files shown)


egraumann:
I am still experiencing this issue...selenium 3 dependencies are being imported, even in version 6.0.3. Any information on this?

asolntsev:
@egraumann Please run `mvn dependency:tree` or `gradle dependencies`, and you will see the reason.

Issue #1650, opened 2021-12-01
[New feature] Add new selectors "ByTagAndText" and "WithTagAndText"


maurizio-lattuada:
## The problem
As today, Selenide offers the following selectors aimed to look for elements based solely on their text:

- `ByText`
- `WithText`

Would be nice to locate elements based yes on their text, but also on their tag.

This feature is inspired to the [wildcard locators of Karate](https://github.com/karatelabs/karate/tree/master/karate-core#wildcard-locators), albeit with a simple implementation (aka find _exactly_ a tag).

## Details
I'm going to implement this new feature and to provide a pull request.

## Tell us about your environment
* **Selenide Version**: 6.1.1
* **Chrome\Firefox\IE Version**: Chrome 96.0.4664.55
* **OS Version**: macOS Monterey 12.0.1

asolntsev:
@maurizio-lattuada Thank you for the suggestion. 
It's not bad at all, the PR is good by all means.

But it made me think that it covers only once case (tag+text), but we would probably need more generic approach allowing us to combine any locators. For example, I prefer to find elements by CSS class, not html tag (tag can change during a refactoring/redesign, but class name is often meaningful). 

I mean, instead of `ByTagAndText("div", "Hello")` we could have `byTag("div").and.byText("Hello")`. 
It would allow us to write `byClass(".greeting").and.byText("Hello")` as well.
It would also allow us to combine multiple locators: `byTag("div").and.byClass(".greeting").and.byText("Hello")`. 

In some sense, probably we just need a convenient builder for an XPath? Like fluent API for building an xpath from Java code. 

What do you think?

maurizio-lattuada:
@asolntsev your proposal sounds legit as well. Indeed I created this PR since I need that specific feature (aka search elements having a tag and a text) and, to be honest, was rather easy to extend the behavior of the existing classes `ByText` and `WithText`.

If you agree, maybe you can merge this initial feature (of course, no need to have a new Selenide version being released _just_ for that) and I can open a new ticket as an improvement/enhancement of this one according to your suggestion.

Sounds good to you?

asolntsev:
@maurizio-lattuada The problem is that if we release it as is, then somebody will start using it, and we will need to support it for a long time, even if we create a more generic api. 

Isn't it simple enough to just create classes `ByTagAndText` and `WithTagAndText` right in your project?

maurizio-lattuada:
You got a point there.

[...]

PR #1651, opened 2021-12-01
New selector tag and text

maurizio-lattuada:
## Proposed changes
See #1650.

## Checklist


asolntsev (approved):

asolntsev (commented):
@maurizio-lattuada For your information, we usually don't use all these classes like `ByText` directly. Instead, Selenide has class `Selectors` with bunch of static methods. 

So I added two methods `Selectors.byTagAndText` and `Selectors.withTagAndText` after merging your PR.

PR #1658, opened 2021-12-09
fixed #1656 checking for empty string

### LINKED ISSUE #1656: Selenide should not throw exceptions if selenide.remote is set, but empty
`-Dselenide.remote=` throws an exception "invalid remote parameter" now.

Typical case for CI automation is to set different parameters for Selenium Grid/Selenoid execution and no parameter for local execution.

But `-Dselenide.remote=$grid_address` doesn't work for local execution even if $grid_address is empty.

`-Dselenide.remote=` should work the same way, when it is not set at all.

vinogradoff:
Fix for #1656. if config is empty string, it should be ignored.


asolntsev:
Overceded by https://github.com/selenide/selenide/pull/1663

PR #1659, opened 2021-12-09
fix for #1656 another solution

vinogradoff:
another solution (perhaps a better one)


vinogradoff:
perhaps we could extend this for other properties as well....

It looks like it is not a bad idea to process -Dselenide.anything= the same way as not setting it at all.

PR #1673, opened 2021-12-22
Override default Selenium http timeouts for RemoteWebDriver

IrinaStyazhkina:
Earlier in pr https://github.com/selenide/selenide/pull/1433 default Selenium http timeouts were overriden.

But it didn't work for RemoteWebDriver

Now we also set shorter timeouts (connectTimeout = 1 minute, readTimeout = 2 minutes) for remote.

### DIFF
--- a/src/main/java/com/codeborne/selenide/webdriver/HttpClientTimeouts.java
+++ b/src/main/java/com/codeborne/selenide/webdriver/HttpClientTimeouts.java
@@ -4,9 +4,11 @@ import org.openqa.selenium.WebDriver;
 import org.openqa.selenium.remote.CommandExecutor;
 import org.openqa.selenium.remote.HttpCommandExecutor;
 import org.openqa.selenium.remote.RemoteWebDriver;
+import org.openqa.selenium.remote.TracedCommandExecutor;
 import org.openqa.selenium.remote.http.ClientConfig;
 import org.openqa.selenium.remote.http.HttpClient;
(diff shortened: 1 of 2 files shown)


BorisOsipov:
Not sure that is the best solution. It seems you can use RemoteWebDriver contructor to set custom timeouts in https://github.com/selenide/selenide/blob/master/src/main/java/com/codeborne/selenide/webdriver/RemoteDriverFactory.java#L15 by smth like that - 
```
        ClientConfig clientConfig = ClientConfig
                .defaultConfig()
                .baseUrl(url)
                .readTimeout(Duration.ofMinutes(1))
                .connectionTimeout(Duration.ofMinutes(2));

        Tracer tracer = OpenTelemetryTracer.getInstance();

        CommandExecutor httpCommandExecutor = new HttpCommandExecutor(
                Collections.emptyMap(),
                clientConfig,
                new TracedHttpClient.Factory(tracer, HttpClient.Factory.createDefault()));
        TracedCommandExecutor tracedCommandExecutor = new TracedCommandExecutor(httpCommandExecutor, tracer);
        RemoteWebDriver wd = new RemoteWebDriver(tracedCommandExecutor, capabilities);
  ```

asolntsev:
@IrinaStyazhkina I agree: setting timeouts in constructor of `RemoteDriverFactory` seems to be better than using reflection. What do you think?

asolntsev:
Superseded by https://github.com/selenide/selenide/pull/1703 and released as Selenide 6.2.1

PR #1678, opened 2022-01-04
upgrade to BrowserUpProxy 2.1.3

asolntsev:
It's a fork of BUP, see https://github.com/browserup/browserup-proxy/issues/388#issuecomment-1004097733


rosolko (approved):
Sounds reasonable!
I think next we need to think about migration onto mitm-proxy

asolntsev:
@rosolko Yes, `mitm-proxy` seems to be a better alternative.

PR #1696, opened 2022-01-16
ci: add randomized test matrix for better test coverage

vlsi:
This enables significantly improve test coverage while still keeping test feedback fast and easy configuration.
    
For example, this matrix captures locale-dependent issues like https://github.com/spockframework/spock/issues/1414
See how the matrix looks in the fork: https://github.com/vlsi/selenide/actions/runs/1704145433
    
The idea is that you declare test axis like `os`, `java`, `java_distribution`, `timezone`, `locale`, etc, and `matrix.js` produces several jobs for you.
    
See https://github.com/vlsi/github-actions-random-matrix

### DIFF
--- a/.github/workflows/matrix.js
+++ b/.github/workflows/matrix.js
@@ -1,0 +1,140 @@
+// The script generates a random subset of valid jdk, os, timezone, and other axes.
+// You can preview the results by running "node matrix.js"
+// See https://github.com/vlsi/github-actions-random-matrix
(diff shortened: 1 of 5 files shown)


asolntsev:
@vlsi Thank you for the suggestion, but the change itself seems quite complex. 
I don't think we really need to test Selenium/Selenide code with multiple Java versions. We never experienced problems because of Java-specific bugs.

PR #1702, opened 2022-01-18
Ignore whitespaces for filename in Content-Disposition header

yevgeniy-mikhailov:
## Proposed changes
Current behavior: Selenide doesn't detect filenames in the `download()` method if this one has whitespaces after `filename`. For instance, `attachment; filename = testName.csv`. In that case, a filename is detected as `csv`

After that fix Selenide ignores whitespaces around `=` after `filename` in that header

### DIFF
--- a/src/main/java/com/codeborne/selenide/impl/HttpHelper.java
+++ b/src/main/java/com/codeborne/selenide/impl/HttpHelper.java
@@ -22,7 +22,7 @@ import static org.apache.commons.lang3.StringUtils.left;
 public class HttpHelper {
 
   private static final Pattern FILENAME_IN_CONTENT_DISPOSITION_HEADER =
-    Pattern.compile(".*filename\\*?=\"?((.+)'')?([^\";?]*)\"?(;charset=(.*))?.*", CASE_INSENSITIVE);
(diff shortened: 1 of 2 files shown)


src/main/java/com/codeborne/selenide/impl/HttpHelper.java
```diff
@@ -22,7 +22,7 @@
 public class HttpHelper {
 
   private static final Pattern FILENAME_IN_CONTENT_DISPOSITION_HEADER =
-    Pattern.compile(".*filename\\*?=\"?((.+)'')?([^\";?]*)\"?(;charset=(.*))?.*", CASE_INSENSITIVE);
+    Pattern.compile(".*filename\\*? *= *\"?((.+)'')?([^\";?]*)\"?(;charset=(.*))?.*", CASE_INSENSITIVE);
```
asolntsev, line 25: Please use `\\s` instead of just ` ` - it means "any space symbol" in regexp.

yevgeniy-mikhailov: Looks like it could be dangerous to use it in this way, because `enter` is also could be in `/s` in the regex

asolntsev: Agree. No changes needed.

PR #1737, opened 2022-02-15
moving merge of user defined capabilities to match other factories

andrematosfundao:
## Proposed changes
Some user defined capabilities get replaced due the the merge of predefined capabilities hapening before setting download preferences.

noticed other browser factories allways merge user difined capabilities after browser defined capabilities.

## Checklist
- [ nope ] Checkstyle and unit tests are passed locally with my changes by running `gradlew check chrome_headless firefox_headless` command
-> was not able to run tests onmy env not sure why.
- [ nope ] I have added tests that prove my fix is effective or that my feature works
-> since this is not a new feature is only making functionality the same across browser factories.
-> did add a line to change log


asolntsev:
Some tests are failing:

```
FirefoxDriverFactoryTest > keepConfigurationFirefoxProfileWhenTransferPreferencesFromSystemPropsToDriver() FAILED
    org.opentest4j.AssertionFailedError: 
    expected: 25
     but was: 0)
        at app//com.codeborne.selenide.webdriver.FirefoxDriverFactoryTest.keepConfigurationFirefoxProfileWhenTransferPreferencesFromSystemPropsToDriver(FirefoxDriverFactoryTest.java:74)

```

andrematosfundao:
@asolntsev fixed the issues with the unit tests

asolntsev:
@andrematosfundao Let me first double-check: why do you want to override "pdfjs.disabled" setting?
Selenide intentionally set `"pdfjs.disabled"` to `true` to be able to download and verify PDF files. How are you going to verify the PDF by your automated test if it's opened in the browser?

andrematosfundao:
@asolntsev if you want validate the normal behaviour of the browser that is to open the pdf in a new tab. 
since selenide by default deactivates the pdf reader. 
i did the change to allow the user to run a test with the normal behaviour of the browser.

asolntsev:
@andrematosfundao My question is still in the air: how exactly can you verify the pdf if it's opened in a tab? What exactly do you want to assert in the test?

1. Actually "the normal" behaviour of the browser depends on user settings. Different browsers with different settings can either open pdf or download it. You never know.
2. There is no need to download the browser behaviour (unless you are working in a team which develops the browser itself). You only need to verify your application.

andrematosfundao:
@asolntsev thanks for you quick feedback. wonderfull to see that selenide is so well maintened.

Regarding the default pdf viewer:
if one allows the browser to use the default viewer we can quickly check that the file can be downloaded and processed by the browser, we would also have a confirmation that a user with default config on the browser can read the pdf.

regarding the unit test:
i want to assert that a tester can change the browser to use browser default pdf reader using Configuration.browserCapabilities, which currently is possible on all factories except firefox.

Since i noticed all factories allow the tester to use pdf viewer in the browser with exception of firefox, i made this change. 
It could be that the unit test does not make much sence, I can remove it or replace with you recomendation.

please advise.

asolntsev:
@andrematosfundao Sorry, for my weird communication. :)
Your unit-test is good, and you argumentation is strong ("all factories except firefox allow..."). We can merge this PR, no problems.

I only want to ask what exactly will you check when this PR gets merged. How will look your code which "quickly checks that the file can be downloaded and processed by the browser"? What asserts are you going to use?

andrematosfundao:
@asolntsev currently is used in chrome for exactly that, quick dirty test that the pdf loads, and then we check the url in the viewer tab is the correct one.

in future we may improve on that and actually download the file and check it agaist test data. 

i am tasked with converting a bunch of tests done in "selenium" to a new test framework made with selenide, hence I found the different behavior between chrome and firefox.

asolntsev:
@andrematosfundao Yes! This is what I am trying to say: instead of your "dirty test", I recommend to improve your test right now. Instead of just url check, use method `$.download()`:

```
File file = $.download();
// optionally, you can check the name or content of the file
```

andrematosfundao:
@asolntsev yes we are in agreement on the best practice. 

however as you know we have limited time and need to focus on important tasks first, this pdf feature is not the main feature of my product, I deemed that a little test is better than no test.

I would like selenide framework to allow me to run very basic fast test.

will it be possible to merge this change?

asolntsev:
@andrematosfundao Honestly, I don't understand you. You are saying that this feature is minor and a simple test is enough. But I am suggesting you even more simple test - why don't you want to use it? 

P.S. The PR can be merged at any moment, but it has merging conflicts. Could you resolve them?
image

asolntsev:
nevermind, I have resolved it.

andrematosfundao:
@asolntsev thanks for all the support.

Issue #1768, opened 2022-04-08
[security] Potential Credential leak in report


dacloutier-logmein:
## The problem
When the report is printed at the end, every "text field" that was entered via selenide will display the value that was set in the field.

## Details
While helpful in a local development scenario, once we've passed that, there is potential to leak sensitive information.

## proposed solution
  either add a new method : `$("ELEMENT").setSentiveValue` that would display the value as `*****` in the report.

## Tell us about your environment
* **Selenide Version**:  6.3.4
* **Chrome\Firefox\IE Version**:  Irrelevant to the issue
* **Browser Driver Version**:  Irrelevant to the issue
* **Selenium Version**: Irrelevant to the issue
* **OS Version**: Irrelevant to the issue

## Code To Reproduce Issue [ Good To Have ]
```
$("#emailAddress").setValue("LEAKED_USERNAME@DOMAIN.COM");
$("#password").setValue("LEAKED_CREDENTIALS");
....
rest of the test...
```

output:
(cleaned up for brevity)

```
20:13:13      +-----------------------------+------------------------------------------------------------------------+------------+------------+
20:13:13      | Element                     | Subject                                                                | Status     | ms.        |
20:13:13      +-----------------------------+------------------------------------------------------------------------+------------+------------+
20:13:13      | webdriver                   | create()                                                               | PASS       | 3553       |
20:13:13      | open                        | https://someURL/resource/                                              | PASS       | 6288       |
20:13:13      | html                        | get wrapped driver()                                                   | PASS       | 60         |
20:13:13      | #credentials                | should([exist, 30 s.])                                                 | PASS       | 213        |
20:13:13      | #emailAddress               | set value(LEAKED_USERNAME@DOMAIN.COM)                                  | PASS       | 333        |
20:13:13      | #password                   | should be([visible, 30 s.])                                            | PASS       | 118        |
20:13:13      | #password                   | set value(LEAKED_CREDENTIALS)                                          | PASS       | 313        |
20:13:13      | #password                   | press enter()                                                          | PASS       | 185        |
20:13:13      +-----------------------------+------------------------------------------------------------------------+------------+------------+
```

asolntsev:
We can do this feature in next release. In progress.

asolntsev:
Implemented in PR https://github.com/selenide/selenide/pull/1770

PR #1770, opened 2022-04-10
#1768 add method to mask passwords etc. in reports

asolntsev:


src/main/java/com/codeborne/selenide/SetValueOptions.java
```diff
@@ -0,0 +1,93 @@
...
+  @CheckReturnValue
+  @Nonnull
+  public SetValueOptions sensitive() {
+    return new SetValueOptions(method, value, mask(value.toString()));
+  }
+
+  @Nonnull
+  @CheckReturnValue
+  String mask(String text) {
+    return REGEX_ANY_CHAR.matcher(text).replaceAll("*");
```
dacloutier-logmein, line 83: This leaves a "hint" on how long the entered string was.
```
"potatoes" => "********"
"apples" => "******"
```
Not a big deal though.

statics/src/test/java/integration/FastSetValueTest.java
```diff
@@ -58,4 +60,15 @@ void fastSetValue_insideFrame() {
...
     $("h2").should(appear).shouldHave(text("john in frame"));
   }
+
+  @Test
+  void fastSetValue_usingOptions() {
+    Configuration.fastSetValue = false;
+    $("#username").setValue(withText("john").withDisplayedText("J* username").usingMethod(JS));
+    $("#usernameHint").should(appear);
+
+    $("#password").setValue(withText("admin").sensitive().usingMethod(JS));
```
dacloutier-logmein, line 70: Love the implementation, didn't thought of doing it that way.

dacloutier-logmein (approved):
Looking forward to use this !

src/main/java/com/codeborne/selenide/SetValueOptions.java
asolntsev: yes, that is a challenge. 
But it also gives a hint when the QA engineer misconfigured a wrong password. Or maybe the password is even empty or null (because of wrong setup etc.).

dacloutier-logmein: yep, that is also why I think this is acceptable.

rosolko: It looks like this mthod shouldn't be exposed as public api. Is it reasonable to mark it as private?

asolntsev: I made it package-private just for unit-test.

asolntsev: Now I made it private (I found that the unit-test didn't really called this method explicitly :) ).

PR #1800, opened 2022-05-10
feat: Add an ability to take a full-screen screenshot using CDP

### LINKED ISSUE #1799: Selenide could take a full-screen screenshot using CDP
see https://github.com/diemol/selenium-4-demo/blob/master/src/test/java/com/saucelabs/demo/BiDiTest.java#L107

rosolko:
Closes #1799 

## Checklist

### DIFF
--- a/src/main/java/com/codeborne/selenide/impl/WebdriverPhotographer.java
+++ b/src/main/java/com/codeborne/selenide/impl/WebdriverPhotographer.java
@@ -29,22 +29,6 @@ public class WebdriverPhotographer implements Photographer {
   @CheckReturnValue
   @Override
   public <T> Optional<T> takeScreenshot(Driver driver, OutputType<T> outputType) {
-    if (driver.config().fullPageScreenshots()) {
-      return takeFullScreenshot(driver, outputType);
-    } else {
-      return takeVisibleScreenshot(driver, outputType);
-    }
-  }
-
-  public <T> Optional<T> takeVisibleScreenshot(Driver driver, OutputType<T> outputType) {
-    if (driver.getWebDriver() instanceof TakesScreenshot) {
-      T screenshot = ((TakesScreenshot) driver.getWebDriver()).getScreenshotAs(outputType);
-      return Optional.of(screenshot);
-    }
-    return Optional.empty();
-  }
-
-  public <T> Optional<T> takeFullScreenshot(Driver driver, OutputType<T> outputType) {
     if (driver.getWebDriver() instanceof ChromiumDriver chromiumDriver) {
       Options options = getOptions(chromiumDriver);
       HashMap<String, Object> captureScreenshotOptions = new HashMap<>() {{
@@ -68,7 +52,7 @@ public class WebdriverPhotographer implements Photographer {
     if (driver.getWebDriver() instanceof RemoteWebDriver remoteWebDriver) {
       String browserName = remoteWebDriver.getCapabilities().getBrowserName();
       if (!(browserName.equalsIgnoreCase(CHROME) || browserName.equalsIgnoreCase(EDGE))) {
-        return Optional.empty();
+        return takeDefaultScreenshot(driver, outputType);
       }
 
       WebDriver webDriver = new Augmenter().augment(remoteWebDriver);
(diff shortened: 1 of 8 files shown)


src/main/java/com/codeborne/selenide/impl/WebdriverPhotographer.java
```diff
@@ -15,10 +18,42 @@ public class WebdriverPhotographer implements Photographer {
...
+
+      Map<String, Object> result = chromeDriver.executeCdpCommand("Page.captureScreenshot", new HashMap<>());
+
+      chromeDriver.executeCdpCommand("Emulation.clearDeviceMetricsOverride", new HashMap<>());
+
+      String base64 = (String) result.get("data");
+      T screenshot = outputType.convertFromBase64Png(base64);
+      return Optional.of(screenshot);
+    }
+    return Optional.empty();
```
rosolko, line 57: Not sure about behaviour in situation when this option is enabled but driver is not a chrome. Should we just return an empty value as well as selenium style screenshot or log/throw an error?

src/main/java/com/codeborne/selenide/impl/WebdriverPhotographer.java
```diff
@@ -15,10 +18,42 @@ public class WebdriverPhotographer implements Photographer {
...
+  public <T> Optional<T> takeVisibleScreenshot(Driver driver, OutputType<T> outputType) {
     if (driver.getWebDriver() instanceof TakesScreenshot) {
       T screenshot = ((TakesScreenshot) driver.getWebDriver()).getScreenshotAs(outputType);
       return Optional.of(screenshot);
     }
     return Optional.empty();
   }
+
+  public <T> Optional<T> takeFullScreenshot(Driver driver, OutputType<T> outputType) {
+    if (driver.getWebDriver() instanceof ChromeDriver chromeDriver) {
```
BorisOsipov, line 37: `ChromeDriver =>ChromiumDriver` to support full screenshots for Edge

BorisOsipov (commented):
It looks good for first time feature implementation. It will be nice if it can work for se4 grid\selenoid\moon as well - not only for local chromium.

rosolko:
> It looks good for first time feature implementation. It will be nice if it can work for se4 grid\selenoid\moon as well - not only for local chromium.

The main problem is devtool dependency inside selenium. The are 4 of it as for now. What should we use as implementation?
Screenshot 2022-05-10 at 21 04 50


Because RemoteWebDriver work with devtools another way - [selenium doc](https://www.selenium.dev/documentation/webdriver/bidirectional/chrome_devtools/#emulate-geo-location-with-the-remote-webdriver)
While trying to use it I have to make an versioned imports: 

```java
import org.openqa.selenium.devtools.v100.page.Page;
import org.openqa.selenium.devtools.v101.emulation.Emulation;
```

BorisOsipov:
>The main problem is devtool dependency inside selenium. The are 4 of it as for now. 

That's true. Issue is `captureBeyondViewport` devtools command flag is experimental and as it described in docs `This may be changed, moved or removed`.

rosolko:
> > The main problem is devtool dependency inside selenium. The are 4 of it as for now.
> 
> That's true. Issue is `captureBeyondViewport` devtools command flag is experimental and as it described in docs `This may be changed, moved or removed`. [chromedevtools.github.io/devtools-protocol/tot/Page/#method-captureScreenshot](https://chromedevtools.github.io/devtools-protocol/tot/Page/#method-captureScreenshot)

As far as I test, if user approach with `Emulation` and `setDeviceMetricsOverride` there are no needs to set `captureBeyondViewport` with true value.

I've add a POC for `RemoteWebDriver` case

BorisOsipov:
@rosolko hmm. I am worried that `setDeviceMetricsOverride` can lead to page re-rendering due changes in page width\height and significant changes in page DOM. is not it?

rosolko:
> @rosolko hmm. I am worried that `setDeviceMetricsOverride` can lead to page re-rendering due changes in page width\height and significant changes in page DOM. is not it?

Yes, it's true. By the right after `captureScreenshot` I've execute a `clearDeviceMetricsOverride` method

asolntsev (approved):

asolntsev:
Ok, I will merge this PR to Selenide branch `feature/#1799-full-screen-screenshots` and fix failing tests there.

asolntsev:
re-worked in https://github.com/selenide/selenide/pull/1858.

PR #1824, opened 2022-05-28
#1818 Devtools support

### LINKED ISSUE #1818: Allow clear localstorage\cookies for all domains in Chromium via devtools.
## The problem
```
Selenide.clearBrowserCookies();
// and
Selenide.clearBrowserLocalStorage();
```
clear caches only for current domain.

## Details
Chromium allows us to clear all caches via devtools
```
WebDriver driver = new Augmenter().augment(WebDriverRunner.getWebDriver());
DevTools devTools = ((HasDevTools) driver).getDevTools();
devTools.createSessionIfThereIsNotOne();
devTools.send(Network.clearBrowserCookies());
devTools.send(Network.clearBrowserCache());
```

Probably we can add them as build-in selenide methods or replace clear Browser Cookies\clearBrowser LocalStorage behaviour at all(not sure).

BorisOsipov:


rosolko:
Is it possible to works with dev tools for chromium based browsers like edge or opera also?

BorisOsipov:
@rosolko yes it is. There is guard here : https://github.com/selenide/selenide/blob/dfbf806788f54347e9de347d142084d94a02e23c/src/main/java/com/codeborne/selenide/impl/CdpProvider.java#L19

I also tested different cases for getting devtools from plain WebDriver instance.
- local\non-local browsers
- drivers enchanced with bytebuddy to support webdriver listeners
- etc

src/main/java/com/codeborne/selenide/drivercommands/DevTools.java
```diff
@@ -0,0 +1,32 @@
...
+  public DevTools(Driver driver) {
+    this.driver = driver;
+  }
+
+  public org.openqa.selenium.devtools.DevTools seleniumDevTools() {
+    return cdpProvider.getCdp(driver);
+  }
+
+  public void clearCookies() {
+    try(var cdp = cdpProvider.getCdp(driver)) {
```
asolntsev, line 22: spaces!

src/main/java/com/codeborne/selenide/drivercommands/DevTools.java
```diff
@@ -0,0 +1,32 @@
...
+      cdp.send(Storage.clearDataForOrigin("*", "cookies"));
+    }
+  }
+
+  public void clearLocalStorage() {
+    try(var cdp = cdpProvider.getCdp(driver)) {
```
asolntsev, line 28: spaces!

src/main/java/com/codeborne/selenide/impl/CdpProvider.java
```diff
@@ -0,0 +1,31 @@
...
+
+@ParametersAreNonnullByDefault
+public interface CdpProvider {
+  @Nonnull
+  @CheckReturnValue
+  DevTools getCdp(Driver driver);
+
+  default void requireChromium(WebDriver driver) {
+    if (driver instanceof HasCapabilities hasCapabilities) {
+      boolean isChromium = new Browser(hasCapabilities.getCapabilities().getBrowserName(), false).isChromium();
```
asolntsev, line 23: only Chromium?
I thought Firefox also supports CDP?

statics/src/main/java/com/codeborne/selenide/Selenide.java
```diff
@@ -911,6 +912,23 @@ public static Clipboard clipboard() {
...
+  /**
+   * Provide access to Chromium DevTools.
+   * Default implementation supports local or Selenium grid 4 sessions only.
+   * <p>
+   * Remote runs support can be implemented via plugins.
+   * Plugin for Selenoid will be released soon.
+   * <p>
+   *
+   * @return DevTools
+   * @since 6.X.0
```
asolntsev, line 924: 6.6.0, I guess?

statics/src/main/java/com/codeborne/selenide/Selenide.java
```diff
@@ -911,6 +912,23 @@ public static Clipboard clipboard() {
     return getSelenideDriver().getClipboard();
   }
 
+  /**
+   * Provide access to Chromium DevTools.
```
asolntsev, line 916: What about Firefox?

asolntsev (changes requested):

statics/src/main/java/com/codeborne/selenide/Selenide.java
BorisOsipov: Probably. Added

statics/src/main/java/com/codeborne/selenide/Selenide.java
BorisOsipov: According the Mozilla docs, Firefox supports CDP only subset of CDP protocol. Actually I haven't seen any CDP commands that works in Firefox.
As I see tools like Playwright patches Firefox sources and distribute its own Firefox binaries versions. I believe we can't tell that our API will support Firefox out of the box without additional impact to research how to allow Firefox actually works with CDP. This PR is a first implementation and Firefox support is out of scope.

PR #1831, opened 2022-06-01
Feature/#1799 full screen screenshots

asolntsev:
fix for #1799, continuation of https://github.com/selenide/selenide/pull/1800

### DIFF
--- a/src/main/java/com/codeborne/selenide/impl/WebdriverPhotographer.java
+++ b/src/main/java/com/codeborne/selenide/impl/WebdriverPhotographer.java
@@ -20,9 +20,6 @@ import java.util.HashMap;
 import java.util.Map;
 import java.util.Optional;
 
-import static com.codeborne.selenide.Browsers.CHROME;
-import static com.codeborne.selenide.Browsers.EDGE;
-
 @ParametersAreNonnullByDefault
 public class WebdriverPhotographer implements Photographer {
   private static final JavaScript js = new JavaScript("get-screen-size.js");
@@ -52,12 +49,10 @@ public class WebdriverPhotographer implements Photographer {
     }
 
     if (driver.getWebDriver() instanceof RemoteWebDriver remoteWebDriver) {
-      String browserName = remoteWebDriver.getCapabilities().getBrowserName();
-      if (!(browserName.equalsIgnoreCase(CHROME) || browserName.equalsIgnoreCase(EDGE))) {
+      WebDriver webDriver = new Augmenter().augment(remoteWebDriver);
+      if (!(webDriver instanceof HasDevTools)) {
         return takeDefaultScreenshot(driver, outputType);
       }
-
-      WebDriver webDriver = new Augmenter().augment(remoteWebDriver);
       DevTools devTools = ((HasDevTools) webDriver).getDevTools();
       devTools.createSession();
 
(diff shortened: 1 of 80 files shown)


src/main/java/com/codeborne/selenide/impl/WebdriverPhotographer.java
```diff
@@ -1,24 +1,94 @@
...
 
 @ParametersAreNonnullByDefault
 public class WebdriverPhotographer implements Photographer {
+  private static final JavaScript js = new JavaScript("get-screen-size.js");
+
   @Nonnull
   @CheckReturnValue
   @Override
   public <T> Optional<T> takeScreenshot(Driver driver, OutputType<T> outputType) {
+    if (driver.getWebDriver() instanceof ChromiumDriver chromiumDriver) {
```
BorisOsipov, line 31: You cannot rely on such type guard. Webdriver instance can be decorated e.g. from webdriver listeners.
example:
```
public class MainPageTest {
    @Test
    public void test() {
        WebDriverRunner.addListener(new WebDriverListener() {
            @Override
            public void beforeAnyCall(Object target, Method method, Object[] args) {

            }
        });
        Selenide.open();
        WebDriver driver = WebDriverRunner.getWebDriver();
        System.out.println(driver.getClass()); // net.bytebuddy.renamed.java.lang.Object$ByteBuddy$gM4k6yJv
        System.out.println(Arrays.toString(driver.getClass().getInterfaces())); // all ChromeDriver interfaces
        assertTrue(driver instanceof ChromiumDriver); // fail :/
    }
}
```

asolntsev:
@rosolko I measured how slow the screenshooting is.

Generating 100 screenshoots took:

|       | Chrome | Chrome headless | Firefox |
| ----------- | ----------- | ----------- | ----------- |
| Old good screenshots      | 30s   |  22s       | 15s |
| Full-screen screenshots   | 35s   |  27s       | 3min |

In Chrome:
* The full-screen screenshots are a bit slower, but the difference is minor. 

In Firefox:
* **The full-screen screenshots are a much slower!**

statics/src/test/java/integration/ScreenshotsTest.java
```diff
@@ -52,4 +52,15 @@ void canTakeScreenshotAsTemporaryFile() throws IOException {
       .hasSize(1);
     assertThat(screenshots.getContextScreenshots().get(0)).isSameAs(screenshot);
   }
+
+  @Test
+  void canTakeFullScreenshot() throws IOException {
```
BorisOsipov, line 57: What this test supposed to test? I changed `WebdriverPhotographer` to old default implementation without full screens support and test won't fail on my pc. I supposed it should fail if we don't take full screen screenshot. What am I missing?
image

asolntsev: Yes, these checks for size a crap. Need to review them.

src/main/java/com/codeborne/selenide/impl/WebdriverPhotographer.java
asolntsev: @BorisOsipov Valid point. Fixed in https://github.com/selenide/selenide/pull/1858.

statics/src/test/java/integration/ScreenshotsTest.java
asolntsev: @BorisOsipov Valid point. Fixed in https://github.com/selenide/selenide/pull/1858 (old tests check the old logic, new tests check the new logic).

asolntsev:
reworked in https://github.com/selenide/selenide/pull/1858
(extracted full-size screenshooting to a separate Selenide plugin).

Issue #1497, opened 2021-07-04
SetValue("") sometimes doesn't work properly with some frameworks and/or validators (React, etc...) 


vinogradoff:
## The problem
 Following code:
 
  $().setValue(""); 

should clear the input field as if the user has cleared the value. But in some cases validator and/or other visual trigger are not working properly, although the field is cleared by the command.

## Details
The reason is, that Selenide is using WebElement.clear() under the hood of Selenium, which doesn't emit the same events, which are typically emitted when user is clearing the field with mouse/keyboard.

These are a numerous reports of this problem for Selenium, for example
https://github.com/SeleniumHQ/selenium/issues/6741
https://stackoverflow.com/questions/50677760/selenium-clear-command-doesnt-clear-the-element
etc.

vinogradoff:
Let's try to introduce a workaround code in Selenide, which will fix this behaviour. PR is following...

PR #1837, opened 2022-06-04
#1497 revert $.clear implementation

asolntsev:
* it didn't work in Safari anyway
* it started failing in Firefox after some of its updates
* it didn't work with TextArea (it cleared only one line of multi-line TextArea)


asolntsev:
Overseded by https://github.com/selenide/selenide/pull/1838

PR #1847, opened 2022-06-08
Feature/#1497 clear with shortcut plugin

asolntsev:
extracts implementation of #1497 to a separate Selenide plugin.


modules/clear-with-shortcut/src/main/java/com/codeborne/selenide/clearwithshortcut/ClearWithShortcut.java
```diff
@@ -0,0 +1,93 @@
...
+  }
+
+  @Nonnull
+  @CheckReturnValue
+  protected Platform getPlatform(Driver driver) {
+    if (!(driver.getWebDriver() instanceof JavascriptExecutor)) {
+      return Platform.UNKNOWN;
+    }
+    try {
+      String platform = driver.executeJavaScript("return navigator.platform");
```
BorisOsipov, line 82: Maybe it would be better if we use https://github.com/selenide/selenide/blob/f6e33870c540d4e91f4cbb29216676d9acba7d81/src/main/java/com/codeborne/selenide/impl/JavaScript.java#L29 here? It seems `JavaScript.java` handles rare cases like driver decorated with `WrapsDriver`

BorisOsipov (approved):
minor comment.

modules/clear-with-shortcut/src/main/java/com/codeborne/selenide/clearwithshortcut/ClearWithShortcut.java
asolntsev: Wow. 
1. Yes
2. Doesn't it mean we need to fix `driver.executeJavaScript` to also support `WrapsDriver` etc.?

asolntsev: @BorisOsipov Fixed it in a separate PR: https://github.com/selenide/selenide/pull/1848

PR #1856, opened 2022-06-14
fix ClearWithShortcut when using EventFiringDriver

petroOv-PDFfiller:
## Proposed changes
In case when we are using WebDriverListeners, getWebDriver return instance of WrapsDriver, which are not instances of Javascript executor. We need to call getWrappedDriver() to get an actual driver instance. 
Similar problem was in com.codeborne.selenide.Driver#getSessionId method. Probably this check should be included in getWebDriver method

## Checklist

### DIFF
--- a/modules/clear-with-shortcut/src/main/java/com/codeborne/selenide/clearwithshortcut/ClearWithShortcut.java
+++ b/modules/clear-with-shortcut/src/main/java/com/codeborne/selenide/clearwithshortcut/ClearWithShortcut.java
@@ -4,9 +4,11 @@ import com.codeborne.selenide.Driver;
 import com.codeborne.selenide.SelenideElement;
 import com.codeborne.selenide.commands.Clear;
 import com.codeborne.selenide.impl.WebElementSource;
-import org.openqa.selenium.JavascriptExecutor;
+import org.openqa.selenium.WebDriver;
 import org.openqa.selenium.WebDriverException;
 import org.openqa.selenium.WebElement;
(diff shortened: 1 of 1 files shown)


modules/clear-with-shortcut/src/main/java/com/codeborne/selenide/clearwithshortcut/ClearWithShortcut.java
```diff
@@ -75,7 +77,12 @@ protected void clearWithShortcut(Driver driver, WebElement input) {
   @Nonnull
   @CheckReturnValue
   protected Platform getPlatform(Driver driver) {
-    if (!(driver.getWebDriver() instanceof JavascriptExecutor)) {
+    WebDriver webDriver = driver.getWebDriver();
+
+    if (webDriver instanceof WrapsDriver)
```
BorisOsipov, line 82: Maybe it will be better move this check to https://github.com/selenide/selenide/blob/master/src/main/java/com/codeborne/selenide/Driver.java#L51 and use it here like `if(!driver.supportsJavascript())`?

petroOv-PDFfiller: I was thinking to change getWebDriver method behavior, but you are right, instanceOf check better to replace with supportsJavascript() call

asolntsev (approved):
merged now, to be continued...

PR #1894, opened 2022-07-14
add cachelookup annotation support

groov1kk:
## Proposed changes
Added support of [@CachedLookup](https://www.selenium.dev/selenium/docs/api/java/org/openqa/selenium/support/CacheLookup.html) selenium annotation for page objects:

```java
public class SelectsPage {

    @CacheLookup
    @FindBy(xpath = "//select[@name='domain']")
    WebElement domainSelect;

    @CacheLookup
    @FindBy(tagName = "h1")
    SelenideElement h1;

    @CacheLookup
    @FindBy(tagName = "h2")
    List<SelenideElement> h2s;

    @CacheLookup
    @FindBy(tagName = "h2")
    ElementsCollection h2sElementsCollection;

    @CacheLookup
    @FindBy(id = "status")
    StatusBlock status;

    @CacheLookup
    @FindBy(css = "#user-table tbody tr")
    List<UserInfo> userInfoList;

  }
```

## Checklist

### DIFF
--- a/src/main/java/com/codeborne/selenide/impl/ElementsContainerCollection.java
+++ b/src/main/java/com/codeborne/selenide/impl/ElementsContainerCollection.java
@@ -4,52 +4,49 @@ import com.codeborne.selenide.Driver;
 import com.codeborne.selenide.ElementsContainer;
 import com.codeborne.selenide.ex.ElementNotFound;
 import com.codeborne.selenide.ex.PageObjectException;
-import org.openqa.selenium.By;
 import org.openqa.selenium.NoSuchElementException;
 
-import javax.annotation.CheckReturnValue;
-import javax.annotation.Nonnull;
-import javax.annotation.Nullable;
-import javax.annotation.ParametersAreNonnullByDefault;
 import java.lang.reflect.Field;
 import java.lang.reflect.Type;
 import java.util.AbstractList;
+import javax.annotation.CheckReturnValue;
+import javax.annotation.Nonnull;
+import javax.annotation.ParametersAreNonnullByDefault;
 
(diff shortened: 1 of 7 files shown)


asolntsev:
Hi @groov1kk !
I like how you implemented this feature. A deep understanding of the existing code base. Nice!

I have only two concerns:
1. Is it really needed? Did you have a chance to measure how many seconds will this feature save? 
When I measured it last time, the gain for so small that I decided element caching is not worth the effort. It gave so few seconds... Most of the times, the slowness comes from browser, backend, services etc., but not from finding the web elements.
2. If we decide it's needed, it would be also great to implement caching of elements in the "standard" methods `$` and `$$`. But what would be the API? Should it be a global setting, or an additional parameter in overloaded `$`/`$$` method?

groov1kk:
Hello
1. From my experience, I received benefit from this feature when I worked with complex static structures such as tables, in cases when I was needed to iterate through many rows and cells to find desired element. And, for example, I had to iterate through this structure several times in one test. But yes, it wasn't significant improvement, something about 5-15 seconds. Anyway, pure selenium has this feature and I just have a habbit to mark all static content with this annotation just not to make additional useless requests :)
2. I thought about this. It shouldn't be a global feature because usually we have to work both with static and dynamic elements. Developer must decide how to work with elements on the spot. So, either we can overload existing methods (`$(cssSelector, shouldCache)`, etc) or create different methods (`cache(seleniumSelector)`, `cacheAll(cssSelector)`, `xCache(xpathSelector)`, etc)

asolntsev (approved):
Seems a very well done PR to me. 
Thumbs up! @groov1kk

Issue #1899, opened 2022-07-26
Window in new tab has different dimension then default


Hetjavier:
When a new tab is opened the default settings from the window size are no longer applied

## Tell us about your environment
* **Selenide Version 6.6.6
* **Chrome\Firefox\IE Version Chrome 103
* **Browser Driver Version**: 
* **Selenium Version** 4.3.0
* **OS Version**: 

## Code To Reproduce Issue [ Good To Have ]

Mainly opening a new tab and switching to it gives a smaller browser size

asolntsev:
@Hetjavier   But how exactly do you open the new tab? What method or code do you use?

BorisOsipov:
It is known issue for Chrome in headless mode - new tab has a smaller browser window size

asolntsev:
@Hetjavier I prepared a pull request which makes `Selenide.switchTo().newWindow(WINDOW)` open a new window of size `Configuration.browserSize` (and even at position `Configuration.browserPosition`). But I am still in doubt is it really so much important feature?

Hetjavier:
It's not that important but it is nowhere documented that it is working. As I saw Boris left a comment that this could have influence in own web drivers that get passed, so it can be discarded if it could cause more problems I assume.

asolntsev:
Decided to abandon this feature, see https://github.com/selenide/selenide/pull/1908

PR #1908, opened 2022-08-03
#1899 make switchTo().newWindow() open the window of the configured size

asolntsev:
fix for #1899


BorisOsipov (changes requested):
@asolntsev We can't do that. If user provides its own webdriver instance(via WebDriverProvider or in another ways) we can't resize the window when we want.

asolntsev:
> @asolntsev We can't do that. If user provides its own webdriver instance(via WebDriverProvider or in another ways) we can't resize the window when we want.

Good point, thanks. 

Actually we could use size of current window for the new window size. 
But yes, probably it's not needed at all. Let's close this PR.

Issue #1917, opened 2022-08-13
Selenide has a memory leak


asolntsev:
## The problem
Every time when Selenide opens a browser, it adds a new "shutdown hook" to JVM. 
These shutdown hooks will be erased only in the end of JVM life. 

The problem is not critical because these objects don't occupy too much much memory.

## Tell us about your environment
* **Selenide Version**: 6.7.1

## Code To Reproduce Issue [ Good To Have ]
```java
final class MemoryUsageTest {
  @Test
  void closeAndOpen() {
    for (int i = 0; i < 1000; i++) {
      open("about:blank");
      closeWebDriver();
      System.gc();
    }
  }
```

Memory consumption:
<img width="564" alt="image" src="https://user-images.githubusercontent.com/279773/184475079-bbf8ee0f-d245-43ec-8e72-a010262e099c.png">

PR #1919, opened 2022-08-13
#1917 fix memory leak

asolntsev:
fix for https://github.com/selenide/selenide/issues/1917


asolntsev:
As a result of this PR, memory consumption does't grow anymore:

image

Issue #1923, opened 2022-08-16
Add partialValue condition


DaniilMoiseevPenza:
## Given
`Configuration.textCheck = FULL_TEXT`

## The problem
In Selenide version 6.6.6 `shouldHave(value("text"))` asserts substring.
F.e. value = 'text1 text2 text3'
in 6.6.6 `element.shouldHave(value("text"))` was ok
but in 6.7.2 it failed

## Solution

Add a new partialValue() condition simular with partialText()

PR #1924, opened 2022-08-16
#1923 add condition `partialValue`

asolntsev:


src/main/java/com/codeborne/selenide/Condition.java
```diff
@@ -213,6 +214,21 @@ public static Condition value(String expectedValue) {
     return new Value(expectedValue);
   }
 
+  /**
+   * Assert that element contains given "value" attribute as substring
+   * NB! Ignores difference in non-visible characters like spaces, non-breakable spaces, tabs, newlines  etc.
+   *
+   * <p>Sample: {@code $("input").shouldHave(value("12345 666 77"));}</p>
```
BorisOsipov, line 221:
```suggestion
   * <p>Sample: {@code $("input").shouldHave(partialValue("12345 666 77"));}</p>
```

BorisOsipov (changes requested):
LGTM. Minor docs issue

src/main/java/com/codeborne/selenide/Condition.java
asolntsev: thanks, fixed.

PR #1934, opened 2022-08-26
Check that element is <select> in methods $.getSelectedText(), getSelectedValue()

asolntsev:
1. renamed `$.getSelectedText()` to `$.getSelectedOptionText()`
2. renamed `getSelectedValue()` to `getSelectedOptionValue()`
3. Now these method verify that the element is `<select>`. Otherwise they throw an `IllegalArgumentException` with message `Expected <select>, but received: <div>`. 
4. Fixed these methods in IE (it seems they never worked correctly in IE).

### DIFF
--- a/src/main/resources/get-selected-option-text.js
+++ b/src/main/resources/get-selected-option-text.js
@@ -1,0 +1,10 @@
+(function(select) {
+  if (select.tagName.toLowerCase() !== 'select') {
+    return ['', `Expected <select>, but received: <${select.tagName.toLowerCase()}>`]
+  }
+  if (!select.options) {
+    return ['', 'Select has no options']
+  }
+  return [select.options[select.selectedIndex]?.text, null]
+})(arguments[0]);
+
--- a/src/main/resources/get-selected-option-text.js
+++ b/src/main/resources/get-selected-option-text.js
@@ -1,10 +1,12 @@
 (function(select) {
   if (select.tagName.toLowerCase() !== 'select') {
-    return ['', `Expected <select>, but received: <${select.tagName.toLowerCase()}>`]
+    return ['', 'Expected <select>, but received: <' + select.tagName.toLowerCase() + '>']
   }
   if (!select.options) {
     return ['', 'Select has no options']
   }
-  return [select.options[select.selectedIndex]?.text, null]
+  let option = select.options[select.selectedIndex];
+  let text = !option ? null : option?.text;
+  return [text, null]
 })(arguments[0]);
 
--- a/src/main/resources/get-selected-option-value.js
+++ b/src/main/resources/get-selected-option-value.js
@@ -1,10 +1,12 @@
 (function(select) {
   if (select.tagName.toLowerCase() !== 'select') {
-    return ['', `Expected <select>, but received: <${select.tagName.toLowerCase()}>`]
+    return ['', 'Expected <select>, but received: <' + select.tagName.toLowerCase() + '>']
   }
   if (!select.options) {
     return ['', 'Select has no options']
   }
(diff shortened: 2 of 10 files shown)


src/main/resources/get-selected-option-text.js
```diff
@@ -0,0 +1,10 @@
+(function(select) {
+  if (select.tagName.toLowerCase() !== 'select') {
+    return ['', `Expected <select>, but received: <${select.tagName.toLowerCase()}>`]
```
BorisOsipov, line 3: IE doesn't suppport template literals.

src/main/resources/get-selected-option-text.js
```diff
@@ -0,0 +1,10 @@
...
+  }
+  if (!select.options) {
+    return ['', 'Select has no options']
+  }
+  return [select.options[select.selectedIndex]?.text, null]
```
BorisOsipov, line 8: IE doesn't suppport optional chaining.

src/main/resources/get-selected-option-value.js
```diff
@@ -0,0 +1,10 @@
+(function(select) {
+  if (select.tagName.toLowerCase() !== 'select') {
+    return ['', `Expected <select>, but received: <${select.tagName.toLowerCase()}>`]
```
BorisOsipov, line 3: IE support please!

src/main/resources/get-selected-option-value.js
```diff
@@ -0,0 +1,10 @@
...
+  }
+  if (!select.options) {
+    return ['', 'Select has no options']
+  }
+  return [select.options[select.selectedIndex]?.value, null]
```
BorisOsipov, line 8: IE support please!

BorisOsipov (approved):
LGTM

@asolntsev you can fix IE compatibility issues via [babel online](https://babeljs.io/repl#?browsers=ie%2011&build=&builtIns=false&corejs=3.21&spec=false&loose=false&code_lz=BQMwrgdgxgLglgewsAzgUwDZtgSgAQDeAUHnnCHqptjAHQwCGA5gHIMC2a9CAMggO5oATgGEG6YPgCEAXhl4A5OiywF-YqVJC0MMEIh4A2goUAaPAAMAogA8ADjTQATPAB5lNAHzmARmBh42lBocABuzgBcbgAkBB6w9MxsnNx8gqLiaJIAvp4WALokeNlF5JRS8XQIdvBIKOpFWjp6BsZmigDK1LB4ABbieBAIeNW1ECgKhaQlTbr6RpW0o4jjhouVzgCSEE5oNvkA_PR7MOYQYBgYhdk4wAxCTGCcEDAohgAM-TgA3EA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=false&timeTravel=false&sourceType=module&lineWrap=true&presets=env&prettier=false&targets=&version=7.18.13&externalPlugins=&assumptions=%7B%7D)

src/main/resources/get-selected-option-text.js
asolntsev: fixed.

src/main/resources/get-selected-option-text.js
asolntsev: fixed.

src/main/resources/get-selected-option-value.js
asolntsev: fixed.

src/main/resources/get-selected-option-value.js
asolntsev: fixed.

PR #1941, opened 2022-09-02
Add chromium support

asolntsev:
Try to run tests on 
* Chromum
* Brave
* ...?


asolntsev:
Probably not needed anymore since I created a separate project demonstrating how to run tests in Chromium (in Docker):

https://github.com/selenide-examples/tests-in-docker

Issue #1903, opened 2022-08-01
Add alias in page initialization by @*FindBy 


wasder:
## The problem
It would be convenient to be able to set element's alias in page object using special annotation. 

## Details
Something like this:
```java
    @Alias("Phone input")
    @AndroidFindBy(id = "phoneInputText")
    @iOSXCUITFindBy(accessibility = "phoneInputField")
    public SelenideElement phoneInput;
```

asolntsev:
Probably we should name this annotation `@As` because selenide already has method `as` for the same purpose.

PR #1956, opened 2022-09-20
1903 add annotation @As for page object fields

asolntsev:
implementation of https://github.com/selenide/selenide/issues/1903

Now you can give human-readable aliases for page object fields:

```java
static class PageObject {
  @As("Large header")
  @FindBy(tagName = "h1")
  SelenideElement header1;
}
```


BorisOsipov (approved):
LGTM

Issue #1850, opened 2022-06-09
NoClassDefFoundError when calling open() after adding a WebDriverListener via WebDriverRunner.addListener()


flange-ipb:
## The problem
After calling `Selenide.open()` I get the following exception:
```
java.lang.NoClassDefFoundError: net/bytebuddy/matcher/ElementMatcher
	at com.codeborne.selenide.drivercommands.CreateDriverCommand.addWebDriverListeners(CreateDriverCommand.java:129)
	at com.codeborne.selenide.drivercommands.CreateDriverCommand.addListeners(CreateDriverCommand.java:103)
	at com.codeborne.selenide.drivercommands.CreateDriverCommand.lambda$createDriver$1(CreateDriverCommand.java:85)
	at com.codeborne.selenide.logevents.SelenideLogger.get(SelenideLogger.java:165)
	at com.codeborne.selenide.drivercommands.CreateDriverCommand.createDriver(CreateDriverCommand.java:53)
	at com.codeborne.selenide.impl.WebDriverThreadLocalContainer.createDriver(WebDriverThreadLocalContainer.java:170)
	at com.codeborne.selenide.impl.WebDriverThreadLocalContainer.getAndCheckWebDriver(WebDriverThreadLocalContainer.java:156)
	at com.codeborne.selenide.WebDriverRunner.getAndCheckWebDriver(WebDriverRunner.java:130)
	at com.codeborne.selenide.impl.StaticDriver.getAndCheckWebDriver(StaticDriver.java:65)
	at com.codeborne.selenide.SelenideDriver.getAndCheckWebDriver(SelenideDriver.java:206)
	at com.codeborne.selenide.drivercommands.Navigator.lambda$navigateTo$0(Navigator.java:72)
	at com.codeborne.selenide.logevents.SelenideLogger.run(SelenideLogger.java:152)
	at com.codeborne.selenide.drivercommands.Navigator.navigateTo(Navigator.java:70)
	at com.codeborne.selenide.drivercommands.Navigator.open(Navigator.java:40)
	at com.codeborne.selenide.SelenideDriver.open(SelenideDriver.java:109)
	at com.codeborne.selenide.SelenideDriver.open(SelenideDriver.java:139)
	at com.codeborne.selenide.SelenideDriver.open(SelenideDriver.java:124)
	at com.codeborne.selenide.Selenide.open(Selenide.java:163)
	at de.ipb_halle.integrationtests.login.LoginTest.before(LoginTest.java:57)
[...]

PR #1963, opened 2022-09-23
optimize dependencies

asolntsev:
* bump LittleProxy from 2.0.11 to 2.0.12
* exclude 'io.opentelemetry' dependency (Selenium brings it transitively, but we really don't use it)
* make `slf4j-api` available for users - apparently all Selenide users need to use it anyway

### DIFF
--- a/gradle/dependencies.gradle
+++ b/gradle/dependencies.gradle
@@ -7,7 +7,7 @@ ext {
   mockitoVersion = '4.8.0'
   slf4jVersion = '2.0.2'
   browserupProxyVersion = '2.2.3'
-  littleProxyVersion = '2.0.11'
+  littleProxyVersion = '2.0.12'
   commonsFileuploadVersion = '1.4'
   jabelVersion = '0.4.2'
   byteBuddyVersion = '1.12.17'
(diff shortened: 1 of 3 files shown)


gradle/dependencies.gradle
```diff
@@ -25,7 +25,11 @@ subprojects {
     compileOnly("com.github.bsideup.jabel:jabel-javac-plugin:${jabelVersion}")
     testCompileOnly("com.github.bsideup.jabel:jabel-javac-plugin:${jabelVersion}")
 
-    api("org.seleniumhq.selenium:selenium-java:$seleniumVersion")
+    api("org.seleniumhq.selenium:selenium-java:$seleniumVersion") {
+      exclude group: 'io.opentelemetry'
+      exclude group: 'org.slf4j'
+      exclude group: 'net.bytebuddy'
```
BorisOsipov, line 31: don't do that because https://github.com/selenide/selenide/issues/1850

BorisOsipov (changes requested):

gradle/dependencies.gradle
asolntsev: wow. 
Oops.
We should add tests to avoid this in future.

BorisOsipov (approved):

PR #1970, opened 2022-09-29
improve resolving proxy host name

asolntsev:
use `new NetworkUtils().getNonLoopbackAddressOfThisMachine()` instead of `ClientUtil.getConnectableAddress()`.


src/main/java/com/codeborne/selenide/proxy/SelenideProxyServer.java
```diff
@@ -147,8 +154,17 @@ static InetSocketAddress getProxyAddress(Proxy proxy) {
...
-      : ClientUtil.createSeleniumProxy(this.proxy, inetAddressResolver.getInetAddressByName(config.proxyHost()));
+      ? ClientUtil.createSeleniumProxy(this.proxy, guessHostName())
+      : ClientUtil.createSeleniumProxy(this.proxy, config.proxyHost());
+  }
+
+  private String guessHostName() {
+    String browserupHostName = getConnectableAddress().getHostName();
+    String seleniumHostName = new NetworkUtils().getNonLoopbackAddressOfThisMachine();
+    if (!Objects.equals(browserupHostName, seleniumHostName)) {
+      log.info("Host name resolved by BrowserUpProxy: '{}', resolved by Selenium: '{}'", browserupHostName, seleniumHostName);
```
BorisOsipov, line 165: Does it helpful log message? It seems just say that there are two different host names but don't say which one will be used.

asolntsev: I hope the actually used IP will be visible in the following logs. 
But it seems it's not always fully visible (it's often truncated).

PR #1972, opened 2022-09-29
jdk-http-client

asolntsev:
see https://www.selenium.dev/blog/2022/using-java11-httpclient/


asolntsev:
Not needed after upgrading to Selenium 4.14.0

PR #2017, opened 2022-11-15
Fix an issue when a new tab size in headless chrome has incorrect size.

BorisOsipov:
## Proposed changes

Fix an issue when a new tab size in headless chrome has incorrect size.

## Checklist


asolntsev (approved):
@BorisOsipov It would be even better if we didn't call method `adjustBrowserSize` for chromium browser.

BorisOsipov:
>It would be even better if we didn't call method adjustBrowserSize for chromium browser.

I have added that.

Issue #1817, opened 2022-05-26
Add methods to copy and paste content


asolntsev:
## The problem
It's not trivial for users to find out how to Copy and Paste text in Selenide/Selenium. 
Selenide could suggest such a functionality. 

Internally, it would send keys "Ctrl+C"/"Cmd+C" and "Ctrl+V"/"Cmd+V" (depending on `navigator.platform`). 

## Tell us about your environment
* **Selenide Version**: 6.5.0

evpl:
@asolntsev 

Hi! Please rate the idea:
1) Add `copy()` method to `SelenideElement` by using `SelenideElement.getText()` and `Clipboard.setText(String)`
2) Add `paste()` method to `SelenideElement` by using `Clipboard.getText()` and `SelenideElement.sendKeys(String)`

Is this implementation suitable?

asolntsev:
@evpl I think `$.paste()` idea is good. It's very common to paste text from clipboard into some field.
But `$.copy()` ... is it a common scenario? Users usually don't copy text of a web element. They usually select some text by click or double-click, and then copy the selected text.

evpl:
@asolntsev 

Ok, I'll do it!

PR #2027, opened 2022-11-19
#1817 add methods to copy and paste content

evpl:
From issue #1817 

## Proposed changes
- added `getSelectedText`, `copy` methods for `SelenideDriver` class
- added `getSelectedText`, `copy` methods for `Selenide` class
- added `paste`, method for `SelenideElement` class
- added `Paste` command

## Checklist

### DIFF
--- a/src/main/java/com/codeborne/selenide/SelenideDriver.java
+++ b/src/main/java/com/codeborne/selenide/SelenideDriver.java
@@ -244,6 +244,27 @@ public class SelenideDriver {
     return executeJavaScript("return document.activeElement");
   }
 
+  /**
+   * Returns selected text or empty string if no text is selected.
+   *
+   * @return selected text
+   */
+  @CheckReturnValue
+  @Nonnull
+  public String getSelectedText() {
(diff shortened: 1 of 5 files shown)


evpl:
@asolntsev Hi! Have a look, please. Is this a good decision? I will start writing tests if it is ok.

asolntsev (commented):
@evpl The code looks good.
But the solution works only in headful (and local) environment.
If browser is remote or graphics not enabled, clipboard is not available.

That's why initially I was thinking about using Ctrl+C / Ctrl+V.

asolntsev (approved):

asolntsev:
I decided to merge this PR. 
This solution works only in headful environment, but it's better than nothing. We can improve it later.

asolntsev:
@evpl I should ask you to write some tests. 
Today I realized that this line was wrong:

```javavascript
return window.getSelection.toString()
```

it always returns the same result:
> 'function getSelection() { [native code] }'

evpl:
@asolntsev Didn't make it a bit. Just got to the computer 🙂

PR #2032, opened 2022-11-26
added Press method to accept Key input

amuthansakthivel:
## Proposed changes
Added command to press any keys from keyboard. It will be really helpful during auto-suggestion dropdown without breaking the code reusability. 

$(element).press(DOWN).press(ENTER);

## Checklist
- [x] I have added tests that prove my fix is effective or that my feature works - not applicable

### DIFF
--- a/src/main/java/com/codeborne/selenide/SelenideElement.java
+++ b/src/main/java/com/codeborne/selenide/SelenideElement.java
@@ -6,6 +6,7 @@ import com.codeborne.selenide.files.FileFilter;
 import com.codeborne.selenide.impl.WebElementSource;
 import com.google.errorprone.annotations.CanIgnoreReturnValue;
 import org.openqa.selenium.By;
+import org.openqa.selenium.Keys;
 import org.openqa.selenium.NoSuchElementException;
 import org.openqa.selenium.TakesScreenshot;
 import org.openqa.selenium.WebElement;
@@ -146,7 +147,7 @@ public interface SelenideElement extends WebElement, WrapsDriver, WrapsElement, 
    */
   @Nonnull
   @CanIgnoreReturnValue
-  SelenideElement pressDownArrow();
+  SelenideElement press(Keys key);
 
   /**
    * Get the visible text of this element, including sub-elements without leading/trailing whitespace.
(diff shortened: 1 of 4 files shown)


asolntsev (commented):
The code is good, but do we really need to add separate methods to all possible keys?

`$.pressDownArrow()` is the same as `$.sendKeys(DOWN)`. The latter is even shorter. 

P.S. Yes, I know we have methods `$.pressEnter()` and `$.pressTab()` - and probably it was a mistake for the same reason...

amuthansakthivel:
> The code is good, but do we really need to add separate methods to all possible keys?
> 
> `$.pressDownArrow()` is the same as `$.sendKeys(DOWN)`. The latter is even shorter.
> 
> P.S. Yes, I know we have methods `$.pressEnter()` and `$.pressTab()` - and probably it was a mistake for the same reason...

@asolntsev - You are correct but the sendkeys method is void. We normally use down arrow before press enter.

currently in our project we do something like 
$.setValue("123").sendKeys(Keys.Down);
$.pressEnter()

From now on, we can use $(element)..setValue("123").pressDownArrow().pressEnter()

asolntsev:
@amuthansakthivel This is a valid argument: current `sendKeys` method is void. Probably we could then introduce method `$.press`? Then it would be possible in one line:
```java
$.press(DOWN).click();
$.press(TAB).should(disappear);
```

amuthansakthivel:
> @amuthansakthivel This is a valid argument: current `sendKeys` method is void. Probably we could then introduce method `$.press`? Then it would be possible in one line:
> 
> ```java
> $.press(DOWN).click();
> $.press(TAB).should(disappear);
> ```

@asolntsev - Makes a whole lot of sense. Let me refactor.

amuthansakthivel:
@asolntsev - Do we need methods like press(Keys[] keysToPress) or press(Keys keyToPress, int numberOfTimesToPress)

asolntsev (approved):

src/main/java/com/codeborne/selenide/SelenideElement.java
```diff
@@ -134,6 +135,20 @@ public interface SelenideElement extends WebElement, WrapsDriver, WrapsElement,
...
 
+  /**
+   * Press a Key in Keyboard. Useful for auto-suggestion dropdown: <pre>
+   *  $(".edit").press(Keys.DOWN).pressEnter();</pre>
+   * <p>
+   * Implementation details:
+   * Check that element is displayed and execute <pre>
+   *  WebElement.sendKeys(Keys.DOWN)</pre>
+   *
+   * @see com.codeborne.selenide.commands.PressEscape
```
BorisOsipov, line 146: `@see com.codeborne.selenide.commands.PressEscape` => 
`@see com.codeborne.selenide.commands.Press`

BorisOsipov:
It will be good to write at least one test to avoid regression in future.

Issue #2034, opened 2022-11-29
Define ScreenShoot folder for TestNG


vlad-velichko:
## The problem
JUnit has a feature to define folder for screenshots. TestNG doesn't have it.
https://ru.selenide.org/documentation/screenshots.html

asolntsev:
@vlad-velichko It's not really needed.
First of all, you generally should not need to change this folder at all.
Second, even if you really need, you can just change `Configuration.reportsFolder = folderWithScreenshots;`. 
JUnit listeners also do it.

PR #2035, opened 2022-11-29
#2034 ScreenShoot folder for TestNG

vlad-velichko:
## Proposed changes
#2034 Define ScreenShoot folder for TestNG

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## 6.10.2 (planned to ??.11.2022)
 * bump slf4j from 2.0.4 to 2.0.5
+* #2034 ScreenShoot folder for TestNG
 
 ## 6.10.1 (released 23.11.2022)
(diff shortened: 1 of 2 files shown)


asolntsev:
@vlad-velichko May I ask why do you want to set the folder? What  are you going to do with that folder?
I am asking because probably we can suggest a better/simpler solution.

vlad-velichko:
@asolntsev By default selenide saves fails screenshots into one folder.
But ScreenShoot creates a separate folder per class.
I need to have all screenshots in one folder.
To save all to the default folder I use path "."

asolntsev:
Yes, but what are you going to do next? These files have random names. What will you do with the bunch of random files in that folder?

vlad-velichko:
getLastScreenshot() -> attach to my 3rd-party custom report

BorisOsipov:
getLastScreenshot() returns `File` instance. You can do whatever you what with it. Why do you need a custom folder? Am I missing smth?

PR #2039, opened 2022-11-29
try to use github secrets

asolntsev:


asolntsev:
At the moment, I don't want to waste time on it. Maybe later.

PR #2051, opened 2022-12-04
Selenide#2050: $.select* methods do not fire events

cocorossello:
## Proposed changes

## Checklist
- [X ] Checkstyle and unit tests are passed locally with my changes by running `gradlew check chrome_headless firefox_headless` command


asolntsev (changes requested):
@cocorossello Thank you for the fix! Well done.

Only one question about order of events. You trigger first `change` and then `click` event.
But isn't it reasonable to change the order: first `click` and then `change`?

cocorossello:
Yes, sure, didn't think about that

asolntsev (approved):

PR #2065, opened 2022-12-09
added method to return last made screenshot

armanayvazyan:
## Proposed changes
After taking screenshots (including page source) I needed functionality to get page source, but all the existing methods were returning only images. 

## Checklist
- [x] I have added one unit test  to check that getter is returning non null value

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,13 +19,15 @@ It defines concise fluent API, natural language assertions and does some magic f
 
 Selenide is based on and is compatible to Selenium WebDriver 4.0+
 
-    @Test
-    public void login() {
-      open("/login");
-      $(By.name("user.name")).setValue("johny");
-      $("#submit").click();
-      $("#username").shouldHave(text("Hello, Johny!"));
-    }
+```java
+@Test
+public void login() {
+  open("/login");
+  $(By.name("user.name")).setValue("johny");
+  $("#submit").click();
+  $("#username").shouldHave(text("Hello, Johny!"));
+}
+```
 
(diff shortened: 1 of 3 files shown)


asolntsev:
@armanayvazyan Thank you for the idea. 
I was thinking about even better solution: what if change existing fields `ThreadLocal<List<File>> currentContextScreenshots` and `protected ThreadLocal<List<File>> threadScreenshots` to keep `Screenshot` instances instead of `File`s? We could leave their getters for `File`s for backward compatibility, and add new getters for `Screenshot`s.

armanayvazyan:
> @armanayvazyan Thank you for the idea. I was thinking about even better solution: what if change existing fields `ThreadLocal<List<File>> currentContextScreenshots` and `protected ThreadLocal<List<File>> threadScreenshots` to keep `Screenshot` instances instead of `File`s? We could leave their getters for `File`s for backward compatibility, and add new getters for `Screenshot`s.

The problems here are the old getters name ` `getScreenshots` , `getThreadScreenshots` , `getContextScreenshots` which actually return `File` , do you have any suggestion for new getter names?

asolntsev:
@armanayvazyan I would suggest to use same names, but without `get` prefix. Nowadays it became popular after introducing records in Java.

```java
public List<File> getThreadScreenshots() {...}
public List<File> getContextScreenshots() {...}

public List<Screenshot> threadScreenshots() {...}
public List<Screenshot> contextScreenshots() {...}
```

asolntsev (approved):

PR #2076, opened 2022-12-16
#2038 Fix flaky test in Selenide when downloading file on Windows

### LINKED ISSUE #2038: Flaky test in Selenide when downloading file on Windows
## The problem
e.g. https://github.com/selenide/selenide/actions/runs/3573568542/jobs/6007775272

```java
FileDownloadToFolderTest > downloadEmptyFile() FAILED
    java.lang.IllegalStateException: Failed to cleanup folder D:\a\selenide\selenide\statics\build\downloads\1669720935384_1000_1
        at com.codeborne.selenide.BrowserDownloadsFolder.cleanupBeforeDownload(BrowserDownloadsFolder.java:31)
        at com.codeborne.selenide.impl.DownloadFileToFolder.clickAndWaitForNewFilesInDownloadsFolder(DownloadFileToFolder.java:77)
        at com.codeborne.selenide.impl.DownloadFileToFolder.lambda$download$0(DownloadFileToFolder.java:58)
        at com.codeborne.selenide.impl.WindowsCloser.runAndCloseArisedWindows(WindowsCloser.java:25)
        at com.codeborne.selenide.impl.DownloadFileToFolder.download(DownloadFileToFolder.java:57)
        at com.codeborne.selenide.commands.DownloadFile.execute(DownloadFile.java:67)
        at com.codeborne.selenide.commands.DownloadFile.execute(DownloadFile.java:29)
        at com.codeborne.selenide.commands.Commands.execute(Commands.java:157)
        at com.codeborne.selenide.impl.SelenideElementProxy.dispatchAndRetry(SelenideElementProxy.java:130)
        at com.codeborne.selenide.impl.SelenideElementProxy.invoke(SelenideElementProxy.java:82)
        at jdk.proxy3/jdk.proxy3.$Proxy37.download(Unknown Source)
        at integration.FileDownloadToFolderTest.downloadEmptyFile(FileDownloadToFolderTest.java:188)

        Caused by:
        org.apache.commons.io.IOExceptionList: D:\a\selenide\selenide\statics\build\downloads\1669720935384_1000_1
            at org.apache.commons.io.FileUtils.cleanDirectory(FileUtils.java:331)
            at com.codeborne.selenide.BrowserDownloadsFolder.cleanupBeforeDownload(BrowserDownloadsFolder.java:27)
            ... 11 more

            Caused by:
            java.io.IOException: Cannot delete file: D:\a\selenide\selenide\statics\build\downloads\1669720935384_1000_1\hello_world.txt.crdownload
                at org.apache.commons.io.FileUtils.forceDelete(FileUtils.java:1344)
                at org.apache.commons.io.FileUtils.cleanDirectory(FileUtils.java:324)
                ... 12 more

                Caused by:
                java.nio.file.FileSystemException: D:\a\selenide\selenide\statics\build\downloads\1669720935384_1000_1\hello_world.txt.crdownload: The process cannot access the file because it is being used by another process
                    at java.base/sun.nio.fs.WindowsException.translateToIOException(WindowsException.java:92)
                    at java.base/sun.nio.fs.WindowsException.rethrowAsIOException(WindowsException.java:103)
                    at java.base/sun.nio.fs.WindowsException.rethrowAsIOException(WindowsException.java:108)
                    at java.base/sun.nio.fs.WindowsFileSystemProvider.implDelete(WindowsFileSystemProvider.java:275)
                    at java.base/sun.nio.fs.AbstractFileSystemProvider.deleteIfExists(AbstractFileSystemProvider.java:110)
                    at java.base/java.nio.file.Files.deleteIfExists(Files.java:1191)
                    at org.apache.commons.io.file.PathUtils.deleteFile(PathUtils.java:487)
                    at org.apache.commons.io.file.PathUtils.delete(PathUtils.java:392)
                    at org.apache.commons.io.FileUtils.forceDelete(FileUtils.java:1341)
                    ... 13 more
```

* **Selenide Version**: 6.10.1

BorisOsipov:
subj.
closes #2038

### DIFF
--- a/statics/src/test/java/integration/FileDownloadToFolderTest.java
+++ b/statics/src/test/java/integration/FileDownloadToFolderTest.java
@@ -1,6 +1,7 @@
 package integration;
 
 import com.codeborne.selenide.Configuration;
+import org.apache.commons.lang3.SystemUtils;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 import org.openqa.selenium.By;
@@ -38,6 +39,9 @@ final class FileDownloadToFolderTest extends IntegrationTest {
 
   @BeforeEach
   void setUp() {
+    if (SystemUtils.IS_OS_WINDOWS) {
+      closeWebDriver();
+    }
     Configuration.fileDownload = FOLDER;
     openFile("page_with_uploads.html");
     timeout = 500;


BorisOsipov:
@asolntsev umm... what kind of root cause we can fix? Windows OS doesn't allow to remove folder if some file is being used by another process. It is just fact.

It is not so flaky behavoiur:

```
final class FileDownloadToFolderTest extends IntegrationTest {
  @BeforeEach
  void setUp() {
    Configuration.fileDownload = FOLDER;
    timeout = 500;
  }

  @Test
  public void test1() {
    open("https://testfiledownload.com/");
    $("#onesignal-slidedown-cancel-button").click();
    SelenideElement downlink = $("[href=\"https://speed.hetzner.de/10GB.bin\"]").scrollIntoView(true);
    var shortIncrementTimeout = using(FOLDER)
      .withTimeout(ofSeconds(5))
      .withIncrementTimeout(ofMillis(100))
      .withFilter(withName("10GB.bin"));
    assertThatThrownBy(() -> downlink
      .download(shortIncrementTimeout))
      .isInstanceOf(FileNotFoundException.class);
  }

  @Test
  void test2() throws IOException {
    openFile("page_with_uploads.html");
    File downloadedFile = $(byText("Download me")).download(withExtension("txt"));
    assertThat(downloadedFile.getName()).matches("hello_world.*\\.txt");
    assertThat(downloadedFile).content().isEqualToIgnoringNewLines("Hello, WinRar!");
  }
}
```

`test1` is always passed but `test2` is always failed if we run them in the same browser on windows. Logs https://gist.github.com/BorisOsipov/4b1d2dd0a58abcc1047d955e3ef4260f

asolntsev:
@BorisOsipov My point is that the same problem can happen with Selenide users too.
I see it happens after warning `Folder XXX still contains files crdownload after 5000 ms.`. 

I see few options how we could improve user experience in this case:
1. close browser only if this warning appeared (but how to detect it in the end of test?).
2. instead of throwing `IllegalStateException: Failed to cleanup folder`, just keep in mind that this file was left from a previous test, and ignore it in future actions.

ilya-corp:
The idea to catch exception and exclude "old" files from probale candidates of downloads looks like good. Usually if browser see that folder is already contains file with same name, it automatically adds digits in the round brackets to the name. So we should not receive mistakes or conflicts. For example, "bukh2020 (1).zip"

BorisOsipov:
@asolntsev Aahh, now it is clear. 

>close browser only if this warning appeared (but how to detect it in the end of test?).

It seems impossible. We can't close the browser whatever we want\need.

>instead of throwing IllegalStateException: Failed to cleanup folder, just keep in mind that this file was left from a previous test, and ignore it in future actions.

It seems possible but need investigations. Also it seems out of scope for the PR of the issue `Flaky test in Selenide when downloading file on Windows ` :) I wanted to exclude flaky tests behavior from our CI jobs but not fix the all root causes.

@ilya-corp 
>Usually if browser see that folder is already contains file with same name, it automatically adds digits in the round brackets to the name. So we should not receive mistakes or conflicts. For example, "bukh2020 (1).zip"

Nope we will receive mistakes. Chrome\FF creates a tmp file like `hello_world.txt.crdownload` until it is finally downloaded. So if we will remember `hello_world.txt.crdownload` after some time it will be renamed to `hello_world.txt ` and it lead to unpredictable behavior in the future, because we remembered only `hello_world.txt.crdownload`

asolntsev:
> Also it seems out of scope for the PR

Agree, this PR fixes the flaky test. We can create another ticket for the found problem.

> Nope we will receive mistakes.

Agree, "just keeping in mind" doesn't help. 

Then I have a new suggestion: let's throw an exception if "Folder XXX still contains files crdownload after 5000 ms.". 
User needs to increase timeout in this case. Or avoid downloading too big files in tests.

asolntsev (approved):

BorisOsipov:
>Then I have a new suggestion: let's throw an exception if "Folder XXX still contains files crdownload after 5000 ms.".
User needs to increase timeout in this case. Or avoid downloading too big files in tests.

That's good. Message can be like: "Can't cleat a browser download folder. Probably folder XXX still contains temporary files after previous test and can be used as reliable download folder for other tests. Please don't use the same browser to avoid flaky tests" Smth like this.

asolntsev:
@BorisOsipov Now I realized it would not help :) 
Even if we throw an exception, current test will fail, but the downloading process will continue. And the next test will still fail.

Issue #2087, opened 2022-12-22
@As annotation command didn't seem to work with SelenideElement


PramodKumarYadav:
## The problem
Taking inspiration from [this article](https://selenide.org/2022/09/24/selenide-6.8.0/#as-annotation ), I tried to use @As annotation alias to get meaningful names in the report but that didn't work. 

Note: When I use it as $.as() that works. 
Example (this works):  `private static final SelenideElement USER_NAME_INPUT = $("input[id='edit-name']").as("USER_NAME_INPUT");` it works. So its only the annotation that didnt work. 

## Details
Below is the report that shows that aliases work when I use the `as` command with element but not when I use @At annotation. 

```
[INFO]	2022-12-22 17:11:40,723	[SimpleReport]: Report for assertThatUserCanNotAddPatientIfMissingMandatoryFields()
+----------------------------------------------------------------------------------+------------------------------------------------------------------------+------------+------------+
| Element                                                                          | Subject                                                                | Status     | ms.        |
+----------------------------------------------------------------------------------+------------------------------------------------------------------------+------------+------------+
| webdriver                                                                        | create()                                                               | PASS       | 4056       |
| open                                                                             | https://dev8.ebloodbanking.com                                         | PASS       | 7294       |
| USER_NAME_INPUT                                                                  | should be(editable)                                                    | PASS       | 70         |
| USER_NAME_INPUT                                                                  | set value(dummyUserName)                                         | PASS       | 60         |
| input[id='edit-pass']                                                            | should be(editable)                                                    | PASS       | 46         |
| input[id='edit-pass']                                                            | set value(*******)                                                   | PASS       | 51         |
| SUBMIT_BUTTON                                                                    | should be(visible)                                                     | PASS       | 23         |
```

## Tell us about your environment
* **Selenide Version**:   <selenide.version>6.10.1</selenide.version>
* **Chrome\Firefox\IE Version**:  Chrome 
* **Browser Driver Version**:  Version 108.0.5359.125 (Official Build) (64-bit)
* **Selenium Version**: default from selenide
* **OS Version**: Windows 10

Running tests with Junit5. 
## Code To Reproduce Issue [ Good To Have ]

```
private static final SelenideElement USER_NAME_INPUT = $("input[id='edit-name']").as("USER_NAME_INPUT");

    @As("PASSWORD_INPUT")
    private static final SelenideElement PASSWORD_INPUT = $("input[id='edit-pass']");
```

asolntsev:
[...]

PR #2088, opened 2022-12-22
#2087 support @As annotation for page object fields

asolntsev:
... not annotated by @FindBy, but initialized with $ or $$.

This pull request allows to declare page object fields with `$` or `$$`, but mark them with annotation `@As`:

```java
static class PageObject {
    @As("Dollar header")
    SelenideElement header4 = $("h1");

    @As("Dollar collection")
    ElementsCollection header5 = $$("h2");
}
```

@PramodKumarYadav

### DIFF
--- a/src/main/java/com/codeborne/selenide/impl/SelenidePageFactory.java
+++ b/src/main/java/com/codeborne/selenide/impl/SelenidePageFactory.java
@@ -36,6 +36,7 @@ import java.util.List;
  *
  * @see <a href="https://github.com/SeleniumHQ/selenium/wiki/PageObjects">Page Objects Wiki</a>
  */
+@SuppressWarnings("ChainOfInstanceofChecks")
 @ParametersAreNonnullByDefault
 public class SelenidePageFactory implements PageObjectFactory {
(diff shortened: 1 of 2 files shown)


PramodKumarYadav (approved):
Looks good to me 🙌!

Only remark: The if-else block in the `initFields` method have grown a bit lengthy but I dont fully understand the implementation to comment if this could be optimised further (or not).

Considering that the selenide elements block of code only use `page` and `proxyIn` parameters to derive everything it needs, maybe there is way to break our implementation into further smaller methods that explain our intention here better? However considering I am not able to provide a better refactor alternative at this moment, this is just a passing thought that I wanted to share with you (in case if you can think of an alternative way - or maybe this is already the best alternative). 

For now, thanks a ton for making this feature available on such a short notice! I appreciate it :) .

asolntsev:
@PramodKumarYadav Good point, I refactored this method: split it to smaller methods. 

There is always a room for improvement. :)

PramodKumarYadav:
The refactored code is super clear 🙌. Thanks @asolntsev 😊.

PR #2091, opened 2022-12-23
Added ElementsCollection & CollectionCondition attributes

AlexLAA:
## Proposed changes

- Added attributes() method to ElementCollections to get List<String> of each element attribute value
- Added CollectionCondition.attributes condition (same as texts() & exactTexts()). Get attribute values List and Assert.

**ElementsCollection - get attribute values list**
```java
elementsCollection.attributes("class");
```

**CollectionCondition - assert by attribute values**
```java
elementsCollection.shouldHave(CollectionCondition.attributes("class", "visa", "mastercard", "amexpress"));
elementsCollection.shouldHave(CollectionCondition.exactAttributes("class", "visa", "mastercard", "amexpress"));
```


AlexLAA:
**Sonar Code Smells:**

- AttributesMismatch & AttributesSizeMismatch Exceptions) **classes have 6 Parents** (Same as TextsSizeMismatch, TextsMismatch)
- ElementsCollection - **Catch Exception instead of Error** 8 Years Old Code Smell

asolntsev:
@AlexLAA Thank you for the PR!
1. It would be great to write some integration tests, like https://github.com/selenide/selenide/blob/master/src/test/java/integration/CollectionMethodsTest.java
2. Is it really needed to check attribute value _partially_? The existing check `$.shouldHave(attribute("class", "mastercard"))` checks the full attribute value. It seems reasonable because attribute value is important (compared to text where we usually want to ignore invisible spaces, tabs, newlines etc.).

AlexLAA:
@asolntsev 

1. Added tests for attributes + exactAttributes conditions
2. We found it very useful condition. It looks & feels better then assert each element separately, but it's very subjective of course. I'm just not sure for 100% in equals \ contains part (with removing all spaces)

`Html.text.equals(element.getAttribute(attribute), expectedValue)`
`Html.text.contains(element.getAttribute(attribute), expectedText)`

asolntsev:
@AlexLAA Yes, I am pretty sure attribute values should be checked fully, without ignoring spaces etc. 
Once again: the existing check `$.shouldHave(attribute("class", "mastercard"))` checks the full attribute value, it doesn't ignore spaces etc.

AlexLAA:
@asolntsev 

**Reduce Spaces was Removed**

`Html.text.equals(element.getAttribute(attribute), expectedValue)` -> `element.getAttribute(attribute).equals(expectedValue)`
`Html.text.contains(element.getAttribute(attribute), expectedText)` -> `element.getAttribute(attribute).contains(expectedValue)`

src/main/java/com/codeborne/selenide/collections/Attributes.java
```diff
@@ -0,0 +1,41 @@
...
+  @Override
+  public boolean test(List<WebElement> elements) {
+    if (elements.size() != expectedValues.size()) {
+      return false;
+    }
+
+    for (int i = 0; i < expectedValues.size(); i++) {
+      WebElement element = elements.get(i);
+      String expectedValue = expectedValues.get(i);
+      if (!element.getAttribute(attribute).contains(expectedValue)) {
```
BorisOsipov, line 30: `element.getAttribute()` is nullable and will cause NPE here

AlexLAA: Fixed
```java
  protected String getAttributeValue(WebElement element) {
    String attr = element.getAttribute(attribute);
    return attr == null ? "" : attr;
  }

```

Added Tests for not existing attribute Behaviour

asolntsev:
> Reduce Spaces was Removed

Nice. But I meant not only "reduce spaces", but also comparing the entire String.

In other words, I suggest to delete class `Attributes` and rename `ExactAttributes` to `Attributes`.

AlexLAA:
@asolntsev
I have been thinking for several days)
Maybe it will be more obvious to have similar logic same as `Texts` and `ExactTexts`
But the last word is yours)

asolntsev:
> Maybe it will be more obvious to have similar logic same as `Texts` and `ExactTexts`

But why it will be more obvious?

Again: I think the logic for attributes and texts should be different.
* In texts, we usually want to ignore invisible spaces, tabs, newlines etc. 
* while in attributes, we usually want to check the exact value.

AlexLAA:
Fixed

PR #2092, opened 2022-12-27
fix broken tests on Windows

asolntsev:
In fact, this PR changes public API a little bit (e.g. method `finishContext()` returns `List<Screenshot>` instead of `List<File>`), but I think it's not a problem because hardly someone uses these methods. They were added a long time ago "just in case". :)


BorisOsipov:
seems it is not broken test it is a buggy code from https://github.com/selenide/selenide/pull/2065

asolntsev:
@BorisOsipov yes. :( 
working on it.

Issue #2054, opened 2022-12-06
Add a delay for auto-closing of alerts


tenor-dev:
## The problem
Selenide auto-closes alerts and similar dialogs if they prevent making a screenshot when a test fails.
The dialogs often contain the reason why the test fails, but Selenide makes it impossible to see it.

## Proposed solution
The solution would be to extract the alert/dialog text it, print it to log, explaining that it is being closed automatically and then add a slight delay before closing it, so that people running the tests in front of them have a chance to see it.

BorisOsipov:
Hi @tenor-dev 

>Selenide auto-closes alerts and similar dialogs if they prevent making a screenshot when a test fails.

No, selenide doesn't autoclose alerts. They are autoclosed by browser driver because we set `unhandledPromptBehavior` to `accept`. Alerts\promts are autoclosed  _before any_ interaction with browser not only before screenshots.

tenor-dev:
@BorisOsipov is it possible to disable this in Selenide?

BorisOsipov:
>Maybe ACCEPT_AND_NOTIFY should be the default? Assuming this will log what was inside

NOTIFY means "throw Exception" - not sure that all users should suffer from unexpected alerts and handle them mannualy all the time.
Also alert text for such exception is optional and e.g. Firefox doesnt provide it in error message.

If you want work with alerts please use existing API:
`Selenide.switchTo().alert().getText()`

tenor-dev:
@BorisOsipov so are you saying that there is no way of at least logging text of auto-closed alerts by default?

This hides important errors, that can be difficult to debug because you don't know that this has happened on your CI server.

asolntsev:
@tenor-dev It's an easy solution for your project: just set `Configuration.browserCapabilities.setCapability(UNHANDLED_PROMPT_BEHAVIOUR, ACCEPT_AND_NOTIFY);`. Let the exception flow.

@BorisOsipov Probably it's still a good idea to enable `ACCEPT_AND_NOTIFY` by default? Yes, it breaks backward compatibility, but it's better to solve technical problems and JS errors (which probably cause alerts) than ignoring them silently.

asolntsev:
> Also alert text for such exception is optional and e.g. Firefox doesnt provide it in error message.

@BorisOsipov 
I just checked: both Chrome and Firefox throws an exception with alert text:
```java
UnhandledAlertException: unexpected alert open: {Alert text : Are you sure, Greg?} // chrome
[...]

PR #2095, opened 2022-12-27
#2054 fail the test if unexpected alert encountered

asolntsev:
By default, Selenide silently closes all unexpected alerts, and users don't have any chance to see it and know the alert message. 

Now you can catch the unexpected alerts and know their message (in order to fix the cause).
You only need to add this line to your setup:
```java
import static org.openqa.selenium.UnexpectedAlertBehaviour.ACCEPT_AND_NOTIFY;
import static org.openqa.selenium.remote.CapabilityType.UNHANDLED_PROMPT_BEHAVIOUR;

Configuration.browserCapabilities.setCapability(UNHANDLED_PROMPT_BEHAVIOUR, ACCEPT_AND_NOTIFY);
```

### DIFF
--- a/src/main/java/com/codeborne/selenide/webdriver/AbstractDriverFactory.java
+++ b/src/main/java/com/codeborne/selenide/webdriver/AbstractDriverFactory.java
@@ -21,7 +21,7 @@ import java.util.regex.Pattern;
 import static com.codeborne.selenide.impl.FileHelper.ensureFolderExists;
 import static java.lang.Integer.parseInt;
 import static org.apache.commons.lang3.StringUtils.isBlank;
-import static org.openqa.selenium.UnexpectedAlertBehaviour.ACCEPT_AND_NOTIFY;
+import static org.openqa.selenium.UnexpectedAlertBehaviour.ACCEPT;
 import static org.openqa.selenium.remote.CapabilityType.ACCEPT_INSECURE_CERTS;
 import static org.openqa.selenium.remote.CapabilityType.BROWSER_VERSION;
 import static org.openqa.selenium.remote.CapabilityType.PAGE_LOAD_STRATEGY;
@@ -75,7 +75,7 @@ public abstract class AbstractDriverFactory implements DriverFactory {
     if (browser.supportsInsecureCerts()) {
       capabilities.setCapability(ACCEPT_INSECURE_CERTS, true);
     }
-    capabilities.setCapability(UNHANDLED_PROMPT_BEHAVIOUR, ACCEPT_AND_NOTIFY);
+    capabilities.setCapability(UNHANDLED_PROMPT_BEHAVIOUR, ACCEPT);
 
     transferCapabilitiesFromSystemProperties(capabilities);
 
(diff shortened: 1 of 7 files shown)


src/main/java/com/codeborne/selenide/webdriver/AbstractDriverFactory.java
```diff
@@ -74,7 +75,7 @@ protected <T extends MutableCapabilities> T createCommonCapabilities(T capabilit
     if (browser.supportsInsecureCerts()) {
       capabilities.setCapability(ACCEPT_INSECURE_CERTS, true);
     }
-    capabilities.setCapability(UNHANDLED_PROMPT_BEHAVIOUR, "accept");
+    capabilities.setCapability(UNHANDLED_PROMPT_BEHAVIOUR, ACCEPT_AND_NOTIFY);
```
BorisOsipov, line 78: I am opposed to changing the UNHANDLED_PROMPT_BEHAVIOUR, as it has been the default for many years and no one has blamed it for any issues. Changing it would break the tests of many users, and they do not want to have to handle the repercussions of such a change.

Instead, we have to make sure that selenide don't hide UnhandledAlertException. It will cover all user cases.

BorisOsipov (commented):
LGTM. One major comment.

src/main/java/com/codeborne/selenide/webdriver/AbstractDriverFactory.java
asolntsev: That's the problem. If we leave "accept", then webdriver doesn't throw `UnhandledAlertException`. 

For users who don't want to investigate, there is still an option to to set `capabilities.setCapability(UNHANDLED_PROMPT_BEHAVIOUR, ACCEPT);` in their tests.

BorisOsipov: >That's the problem. If we leave "accept", then webdriver doesn't throw `UnhandledAlertException`.

Why problem? It is awesome! No one cares about shitty alerts for years and now you break tests with version update, users spend time how to set UNHANDLED_PROMPT_BEHAVIOUR back. For what? If user tests allerts there is api for that.

If someone wants `UnhandledAlertException` there is still an option to to set `capabilities.setCapability(UNHANDLED_PROMPT_BEHAVIOUR, ACCEPT_AND_NOTIFY);`

asolntsev: Agree, we can leave `"accept"` by default, and users will be able to set `ACCEPT_AND_NOTIFY` if they wish.

P.S. Why you call alerts "shitty"? They might show some real errors that should not be ignored. If we ignore these alerts, the SUT will  still behave incorrectly, but we would not even know about it.

BorisOsipov: >They might show some real errors that should not be ignored. If we ignore these alerts, the SUT will still behave incorrectly, but we would not even know about it.

That's true. But alerts can sometimes be perceived as spam or useless if they are not relevant or timely for the user. But developers don't care and use them(e.g. in legacy CRM\CMS). In such a case it is ok to ignore it in the test.

Also incorporating excessive alerts into the user interface is a poor design decision from a usability and user experience standpoint, as it is often difficult to provide helpful information to the user through error messages displayed in alert form. For example, it is not possible to include clickable URLs in an alert, which can make it harder for the user to access additional resources or information that may be necessary to resolve an issue.
That's why alerts "shitty" :)

PR #2124, opened 2023-01-13
adding a unique test run directory to  Configuration.reportsFolder directory

PramodKumarYadav:
## Issue
Currently the concept of a `unique test run` for all tests in a run is missing in the implementation for `Configuration.reportsFolder`. 

And although the user can easily implement this at users end by creating a static final constant to provide a test-run-id as here `Jan-12-2023-23-50-23`, considering a unique test run is a core testing concept, we are better off providing it our users from Selenide itself.

#### Case 1: User does not use ScreenShooterExtension.class
If a user takes examples from [here](as shown here: https://selenide.org/documentation/screenshots.html), they will often end up creating a static reports directory path either in the selenide.properties file or via program something like "test-result/test-runs"

After only a few test runs, this results into a lot of screenshots directly placed under one repository with no direct traceability on which screenshots belong to which test run. Example below.

Screenshot 2023-01-13 at 01 20 29

#### Case 2: Even if user uses ScreenShooterExtension.class, in just a few runs, the failures will keep accommodating under the same test runs, with no direct traceability from the reportsDirectory on which screenshots belong to which run. The only possibility is if user goes into logs and see one by one which screenshot is for which failure - which would be very time consuming. This situation is shown below. 
Screenshot 2023-01-13 at 01 28 23

## Proposed changes
Now we can easily deal with this traceability issue for each run by adding a `unique test run id` to the `Configuration.reportsFolder` where ever we first initialise it. This unique run id will then create a dynamic directory insider the static path of reportsFolder provided by the user to add all failures for a run inside that directory. This would then result in a situation as below where each run is unique and if a user runs tests multiple times, he/she exactly knows which screenshots are for the latest test run failures, without any need to go start looking for those inside the logs. As a side benefit, user now also gets a nice trend line based on date-time, that if user desires, can go and see exactly which run has what failures at what moment (without any confusion or mixing results from different runs)
 
Screenshot 2023-01-13 at 01 33 36

NOTE: At this moment, I have only created a PR for ScreenShooterExtension.class but if this PR and/or the concept is excepted, we would need to do the same for all places where we set  `Configuration.reportsFolder`. If you like this proposal, I can also create more PRs to make the behaviour uniform across app (with or without ScreenShooterExtension class).

## Checklist
     - No they didn't. Considering I saw the same errors with my another PR that involves a typo fix in a PR template, I think these failures already exist from some other PR from past.
- [ x ] I have added necessary documentation (if appropriate)

### DIFF
--- a/statics/src/main/java/com/codeborne/selenide/TestRun.java
+++ b/statics/src/main/java/com/codeborne/selenide/TestRun.java
@@ -18,7 +18,7 @@ public class TestRun {
    */
   public static String getUniqueTestRunName() {
     // Idea is that a user typically would not have more than one month data, so starting with a month name, should keep data well sorted.
-    // Also, any typical test run would last for more than a month, so there is no need to add, milli or micro seconds in the name.
+    // Also, any typical test run would last for more than a second, so there is no need to add, milli or micro seconds in the name.
     DateTimeFormatter dateTimeFormatter = DateTimeFormatter.ofPattern("MMM-dd-yyyy-HH-mm-ss");
     return getUniqueTestRunName(dateTimeFormatter);
   }
--- a/statics/src/main/java/com/codeborne/selenide/TestRun.java
+++ b/statics/src/main/java/com/codeborne/selenide/TestRun.java
@@ -17,7 +17,8 @@ public class TestRun {
    * Since a test run is unique for all tests in a run, we only need to set this value once and as static - final.
    */
   public static String getUniqueTestRunName() {
-    //
+    // Idea is that a user typically would not have more than one month data, so starting with a month name, should keep data well sorted.
+    // Also, any typical test run would last for more than a month, so there is no need to add, milli or micro seconds in the name.
     DateTimeFormatter dateTimeFormatter = DateTimeFormatter.ofPattern("MMM-dd-yyyy-HH-mm-ss");
     return getUniqueTestRunName(dateTimeFormatter);
   }
--- a/statics/src/main/java/com/codeborne/selenide/TestRun.java
+++ b/statics/src/main/java/com/codeborne/selenide/TestRun.java
@@ -1,0 +1,41 @@
+package com.codeborne.selenide;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.time.LocalDateTime;
+import java.time.format.DateTimeFormatter;
+
+public class TestRun {
+  private static final Logger log = LoggerFactory.getLogger(TestRun.class);
+
+  private TestRun() {
+    throw new IllegalStateException("Util class");
+  }
+
+  /**
+   * Since a test run is unique for all tests in a run, we only need to set this value once and as static - final.
+   */
+  public static String getUniqueTestRunName() {
+    //
+    DateTimeFormatter dateTimeFormatter = DateTimeFormatter.ofPattern("MMM-dd-yyyy-HH-mm-ss");
+    return getUniqueTestRunName(dateTimeFormatter);
+  }
+
+  /**
+   * Convenience method to allow for a different test run formatting then the one provided by default.
+   */
+  public static String getUniqueTestRunName(DateTimeFormatter dateTimeFormatter) {
+    String testRun = getFormattedDate(dateTimeFormatter);
+
+    log.info("Test run name: {}", testRun);
+    return testRun;
+  }
+
+  public static String getFormattedDate(DateTimeFormatter dateTimeFormatter) {
+    LocalDateTime localDateTime = LocalDateTime.now();
+    log.debug("Before formatting: {}", localDateTime);
+
+    return localDateTime.format(dateTimeFormatter);
+  }
+}
--- a/statics/src/main/java/com/codeborne/selenide/junit5/ScreenShooterExtension.java
+++ b/statics/src/main/java/com/codeborne/selenide/junit5/ScreenShooterExtension.java
@@ -103,7 +103,7 @@ public class ScreenShooterExtension implements BeforeEachCallback, AfterEachCall
    * @return
    */
   @Nonnull
-  public ScreenShooterExtension to(final String folderWithScreenshots, final DateTimeFormatter testRunDateTimeFormat ) {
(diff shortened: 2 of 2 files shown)


BorisOsipov:
>And although the user can easily implement this at users end by creating a static final constant to provide a test-run-id as here Jan-12-2023-23-50-23, considering a unique test run is a core testing concept, we are better off providing it our users from Selenide itself.

I don't need the concept of 'unique test run.' What should I do after making this and future changes?

PramodKumarYadav:
@BorisOsipov : On the question of "What should I do after making this and future changes?", I am not sure what you mean since you said you don't want/need the concept?

For some background, separating tests based on a test run, gives users a direct access to tests failures due to "this" or any other previous run. 

In absence of this, all failure snapshot files goes under the same directory and over time, unless someone looks into individual logs, they cannot find out which screenshot is for which failure. Screenshot extension, also only solves a part of the problem to atleast map the screenshots with class/methods but not to test run and forces users to look into individual logs to find out which screenshot is for which test run. 

Adding a "unique test run" for each run, eliminates the need for looking into individual logs to see which screenshot is for which test and would give users immediate visibility into 
1. failures due to this run.
2. and if a user user screenshot extension. 
3. than also mapping to class/tests containing screenshots only for "this run" and due to this reason, user does not need to go to logs to see which timestamp.jpg files are for which test failure. 

I would say, do give it a second thought and if you still think we don't need to give this visibility/traceability, we can always discard this PR. I or other users can anyways always build it custom into my/their own tests.

BorisOsipov:
>I am not sure what you mean since you said you don't want/need the concept?

1) It is simple. In docs we write 
```
  /**
   * Folder to store screenshots to.
   * Can be configured either programmatically, via selenide.properties file
   * or by system property "-Dselenide.reportsFolder=test-result/reports".
   * <br>
   * Default value: "build/reports/tests" (this is default for Gradle projects)
   */
  public static String reportsFolder = defaults.reportsFolder();
```
With your changes we make breaking changes. Why should we disturb users who don't need a 'unique' folder by creating one for those who do?

2) 
As a user, I want to be able to store reports in the `foo/bar` folder, not in a unique folder that I don't control.
If we merge this I can't do that. 
If someone really need it they could implement with one-liner
`Configuration.reportsFolder =
                "build/report/" + DateTimeFormatter.ofPattern("MMM-dd-yyyy-HH-mm-ss").format(LocalDate.now());`

even it more customizable e.g. I want to add browser name to folder name

`Configuration.reportsFolder =
                String.format("build/report/%s-%s", Configuration.browser, DateTimeFormatter.ofPattern("MMM-dd-yyyy-HH-mm-ss").format(LocalDate.now()));`

asolntsev:
@PramodKumarYadav Let me ask about this:
> unless someone looks into individual logs, they cannot find out which screenshot is for which failure

It sounds controversial to me. What is your workflow? Do you really track the folder with screenshots and analyze them? It's obviously is ineffective wasting of your time. No?

My workflow is quite opposite: I don't track any files. 
CI server (Jenkins or github actions) run the build. 
If the build fails, CI server sends me a notification with the test report. 
Test report contains error messages with links to screenshots. I just open these links. I don't need to browse any folders. 

P.S. 
> In absence of this, all failure snapshot files goes under the same directory and over time
Yes, but it's not a problem because we start a new CI build with `.gradle clean` or `mvn clean`, right?

PramodKumarYadav:
@BorisOsipov : You make a good point about breaking changes and I also agree that if someone wants they can create a unique folder themselves. 

@asolntsev : Your explanation is also valid when it comes to executing tests in CI. Since in a node, you will always get results from one test run. 

The use case, which I was referring was not for CI but for local development. Where typically a tester would try to "add new tests" or "fix existing tests". In local if a user is running tests from an IDE such as Intellij, "typically" we dont rely on opening html reports. 

In past I mostly relied on logs but looking at screenshots is a nice side helper addition to troubleshoot. And local development is the place where I think it can add the most value if these logs can be separated for each test run and not pull in one common repository. 

P.S: I do agree with Boris that it can bring breaking changes, and that it is easy to implement on the user side. But the use case does has value in the develop mode at users local machine. 

I leave it to you guys to sleep over it and see what to do about this change request (if there can be another solution to implement this without affecting existing users). If you still feel after a second consideration that it's not worth the hassle, we can close the PR.

asolntsev:
> for local development

In local development, it's even more simple.
I just run a test and read the error message. It contains a link to screenshot. I don't need to find screenshot it any folders.
As simple as possible. 

image

PramodKumarYadav:
@asolntsev : hmmm, okay I will close the PR.

I still believe we are missing the usefulness of separating failure results based on each run. Yes, we can see a result of a test case "in this run - in the logs of this run" but if you have a lot of test cases, going through all the logs to see failures and their results is less efficient than quickly glancing though the already open directories that contains the screenshots. At the very least, 
- you make the existing test result storage structure more user friendly 
- with a possibility of even going through previous recent run results - something that you cannot do with existing solution. 

So even though I am closing the PR, do give it a thought. 
Cheers!

swaroop-tata:
@asolntsev ,  Is there a way to get the unique folder name?

swaroop-tata:
@asolntsev ,  Is there a way to get the unique folder name?

asolntsev:
@swaroop-tata Can you please specify WHY you need the folder name? How exactly are you going to use it?

swaroop-tata:
I have some test cases to perform on the downloaded file.  so I need the unique folder name to get to the downloaded file.

(downloadDir/<unique folder name>/<downloadedFile)

asolntsev:
@swaroop-tata That's what I was thinking about. ;)
You don't really need to know this folder. 
What you need is just call method `$.download()`. **It returns the download file**. Then you can perform anything with this file.

```java
  File report = $("a#report").download();
  assertThat(report).content().isEqualTo("Hello, world");
```

See https://youtu.be/GwHG550moGc

swaroop-tata:
Oh..Thank you so much.  Appreciate quick help.

swaroop-tata:
Running into issue "The element does not have href attribute:".  No worries.  Will manage my use case with Java code.

asolntsev:
You just can set `Configuration.fileDownload = FOLDER;`

Issue #2171, opened 2023-02-27
Add method `$.cached()`


asolntsev:
## The problem
Sometimes I want to "fix" (or "cache") the SelenideElement instance, so the following operations with this element would be faster.
This is only actual if the element was received by filtering some long collection. 

## Code To Reproduce Issue
```java
ElementsCollection subscriptions = $$("div.subs-expands").filter(childExactText(
       By.xpath(".//div[contains(@class,'subscription-data')]//div[contains(@class,'sub-heading')]"), orderNumber));
SelenideElement target = subscriptions.filter(childAttributeValue(...)).find(childExactText(....));

return target.cached(); // TODO this is the wanted feature
```

## Current workaround
Currently it's possible to cache the web element this way:
```java
SelenideElement cachedTarget = $(target.getWrappedElement());
```
But it 
1. slightly too long and non-obvious syntax, and
2. probably doesn't look good in reports

## TBD
What should be the proper method name?    @BorisOsipov @vinogradoff 
* `$.snapshot()`  (similar to `$$.snapshot()`)
* `$.cached()`
* `$.frozen()`
* `$.freeze()`

P.S. This feature idea was raised from https://github.com/selenide/selenide/issues/2170 discussion.

PR #2189, opened 2023-03-06
#2171 [feature] add method $.cached()

asolntsev:
Basically it returns the same value as $($.toWebElement()), but with simpler syntax. Plus, it contains information about the original selector, and can report it in the error message.

Implements https://github.com/selenide/selenide/issues/2171

@BorisOsipov @bereg2k I am not sure about the name `$.cached()`. Any suggestions?
In collections, we have similar method `$$.snapshot()`. But is it also a good name for a single element: `$.snapshot()`?


bereg2k:
Both `cached()` and `snapshot()` sound good and are relevant to the purpose of the method.

PR #2191, opened 2023-03-07
Add type command

amuthansakthivel:
## Proposed changes
Setvalue and sendkeys cannot really help in auto suggestion dropdown. We need to mimic user behaviour.
I have added a method called type which can be used while interacting with auto suggestion dropdown.

```java
$(element).shouldBe(visible).type("check").press(DOWN_ARROW).press(ENTER);
$(element).type(text("abc").withDelay(ofMillis(100)).clearFirst(false));
```

## Checklist


asolntsev:
If I understand your idea correctly, this `type` method is the same as old good `sendKeys`, but with pauses after every keypress. Right?
Yes, we probably can have this method, but at least we don't want to hardcode the 200ms. It should be a parameter.
But then it's convenient to pass parameter `200` every time when you call this method.
In the end, I guess such a method is too project-specific. 
Probably it's a good idea to create a help method or custom command in your project?

amuthansakthivel:
@asolntsev - I have added a method to pass delay time to type method.

asolntsev:
> Please let me know if you want me to add it as TypeOptions

@amuthansakthivel Yes, I think it's reasonable to introduce `TypeOptions` instead of `long` parameter.

* Also, I would add the second parameter a'la `is clear needed`. Because users might need to type slowly without clearing the input field.

amuthansakthivel:
@asolntsev - I have pushed all the changes.

amuthansakthivel:
Hi @asolntsev - This PR is pending for your review

asolntsev (changes requested):
Thank you, this feature has been requested for a long time.

One suggestion. Maybe we could replace word "with" by "text"? 
And set other parameters with an explicit name via methods. 

so, instead of unclear "false" and "Duration.ofMillis(200)":
```java
$("#username_input").type(TypeOptions.with("test", Duration.ofMillis(200), false));
```

we will have readable names:
```java
$("#username_input").type(text("test").withDuration(Duration.ofMillis(200)).clearFirst(false));
```

src/main/java/com/codeborne/selenide/SelenideElement.java
```diff
@@ -81,6 +81,25 @@ public interface SelenideElement extends WebElement, WrapsDriver, WrapsElement,
...
+   * Mimic how real user would type in a text field
+   * Useful to work with autosuggestion dropdown
+   * @see com.codeborne.selenide.commands.Type
+   */
+  @Nonnull
+  @CanIgnoreReturnValue
+  SelenideElement type(CharSequence textToType);
+
+  /**
+   *
```
adorne, line 94: You need at least a link to the first method.

```suggestion
   * Similar to {@link #type(java.lang.CharSequence)}
```

src/test/java/integration/TypeOptionsTest.java
```diff
@@ -0,0 +1,32 @@
...
+
+class TypeOptionsTest extends ITest {
+  @BeforeEach
+  void openAndInitTestPage() {
+    openFile("page_with_inputevents_on_clear.html");
+  }
+
+  @Test
+  void typeOptions() {
+    $("#username_input").setValue("test");
```
adorne, line 19: Perhaps you should separate these tests for clarity of the results of the checks.

amuthansakthivel: @adorne  - thanks for the suggestion. Now the test is short and we don't need to separate this. 

@asolntsev - Please review again as I have already pushed all the changes

asolntsev (approved):

nartezion:
Hi @asolntsev!
Faced an issue with new method type() in this MR.

**Element should be editable {By.chained({AppiumBy.accessibilityId: Enter value})}**
Element: '<XCUIElementTypeTextField enabled="true" selected="false" name="Enter value" visible="true">Enter value</XCUIElementTypeTextField>'

My element from page source: 
<XCUIElementTypeTextField type=“XCUIElementTypeTextField” name=“Enter value” label=“Enter value” enabled=“true” visible=“true” accessible=“true” x=“44” y=“201" width=“340” height=“48" index=“1”/>

Caused by: WebDriverException: An unknown server-side error occurred while processing the command. Original error: The attribute ‘readonly’ is unknown. Valid attribute names are: (
Caused by: org.openqa.selenium.WebDriverException: An unknown server-side error occurred while processing the command. Original error: The attribute ‘readonly’ is unknown.

I can enter text with use of sendKeys() method, but new method fails cause of ‘readonly’ attribute

Is there any way to fix it?

amuthansakthivel:
Hi @nartezion - The type method is only currently tested on the web. If you can give me an reproducing example, i can fix the problem in mobile tests

nartezion:
@amuthansakthivel Hi! Could you tell me what information you need? Unfortunately, I cannot share my app with you

nartezion:
@amuthansakthivel Hi! Please, help to solve this)

asolntsev:
@nartezion I think @amuthan-sakthivel was asking for some example of such mobile app. 
But I see that the problem is easily reproducible with any mobile app having text input.

The fix is already building: https://github.com/selenide/selenide/pull/2408 ;)

nartezion:
@asolntsev such a great news! Thank you for your work :)

PR #2229, opened 2023-03-25
Try selenium manager

asolntsev:


rosolko (approved):
Since we are not customize webdrivermanager calls, let's use manager.
But we have to decide future of `driverManagerEnabled` configuration option, it looks like it change nothing

asolntsev:
Tried. Worked. Created another PR: https://github.com/selenide/selenide/pull/2400

PR #2245, opened 2023-04-05
Add method $.dragAndDropTo(WebElement, DragAndDropOptions)

Au6ojlut:
## Proposed changes
Add ability to specify `DragAndDropOptions` when using `dragAndDropTo` for `WebElement`

## Checklist


asolntsev:
@Au6ojlut An open question: probably we should put `WebElement` parameter to `DragAndDropOptions`?

Then we would have only a single method `$.dragAndDropTo(options)` with various parameters:

```java
$.dragAndDropTo(to("#targetDiv"));
$.dragAndDropTo(usingJavaScript().to("#targetDiv"));
$.dragAndDropTo(usingJavaScript().to($("#targetDiv")));
```

Au6ojlut:
@asolntsev 
Should I also remove old API or just deprecate it?

asolntsev:
Deprecate

Au6ojlut:
I think, that usage of `usingJavaScript()` to create `DragAndDropOptions` object could lead to errors and misuse of the API because it allows to create options without target. That's why it marked as deprecated as well. 

So, I suggest such approach.
```
$.dragAndDropTo(to("#targetDiv").usingJS());
```

But, it all debatable. The current changes is more like a `proof of concept` rather than the final version.

asolntsev:
@Au6ojlut Oh yes, this is perfect: `$.dragAndDropTo(to("#targetDiv").usingJS());`!
Let's make everything else deprecated. :)

asolntsev:
@Au6ojlut Nice! :) 

Looking at the changes, I came to a conclusion that 
1. The name of this new method could be without "to" suffix: `$.dragAndDrop(to("div"))`.

2. Probably method `$.dragAndDropTo(String)` is not bad. Do we really need to deprecated it?
What is really bad is method with two arguments: `$.dragAndDropTo(element, options)`

3. Why do you convert `target` to `SelenideElement`? It's enough only to use `WebElement` because anyway the only usage of `target` is `target.getWrappedElement()`.

Au6ojlut:
>  The name of this new method could be without "to" suffix: $.dragAndDrop(to("div")).

Agree, will fix it.

> Probably method $.dragAndDropTo(String) is not bad. Do we really need to deprecated it?
    What is really bad is method with two arguments: $.dragAndDropTo(element, options)

I think that ability to do same things by different ways is bad idea. Only one must remain =)
I believe that the single `options` argument (the builder way) is much better then overloaded methods


> Why do you convert target to SelenideElement? It's enough only to use WebElement because anyway the only usage of target is target.getWrappedElement().

Mostly because I am not too much familiar with the project yet so I just copied it (`monkey see monkey do`).
But, if I am not mistaken, without wrapping the `target` with a proxy we could fail on `shouldBe(visible)` on first try without retry until timeout

Au6ojlut:
Marked this PR as draft, because in the end I want to squash all commits to one with good commit message to avoid commit-history pollution. (maybe I'll close this PR and open another one, maybe it will be more correct)

Renamed method to just `dragAndDrop`.

If there are some other ideas I am ready to discuss

asolntsev:
There is an easier solution: I can choose "Squash" when merging the pull request: :)

image

Issue #2268, opened 2023-04-28
Condition with `LocalDate`


y-lazakovich:
## The problem
It would be nice if Selenide could verify some date field with required pattern format which used in LocalDate class.

Given a field of type `date`:
```java
SelenideElement birthday = $("input#birthday");
LocalDate expected = LocalDate.parse("1981-06-08");
```

We could add few conditions like this:

1. `birthday.shouldHave(date(expected));`
2. `birthday.shouldHave(date(expected, ofPattern("dd.MM.yyyy")));`
3. `birthday.shouldHave(dateBetween(expected.minusDays(2), expected.plusDays(3)));` 

See https://t.me/selenide_ru/49201

asolntsev:
Implemented in https://github.com/selenide/selenide/pull/2281

PR #2281, opened 2023-05-11
#2268 Add conditions `date(...)` for LocalDate and `datetime(...)` for LocalDateTime to check date values

Au6ojlut:
Also:

Add ability to check format with `dateFormat(...)` and `datetimeFormat(...)` Add ability to check that date is inside interval using `dateBetween(...)` and `datetimeBetween(...)`

The pattern is `String`, not `DateTimeFormatter` because there is no ability to get original pattern from `DateTimeFormatter`

## Proposed changes
Ability to be able to check date values easily.

## Checklist


Au6ojlut:
Not sure about naming and default date/datetime pattern

p.s. will fix sonar's QG in near future...

asolntsev:
@Au6ojlut Thank you! Will review the PR this week.

P.S. Few tests are failing: 
```
DateConditionsTest > failure_wrongDateFormat() FAILED
    java.lang.AssertionError: 
    Expecting actual:
      "Element should have date value: "2022-11-11" (with date value format: "yyyy-MM-dd") {#birthdate}
    Element: '<input id="birthdate" type="text" value="2022/10/11"></input>'
    Actual value: 2022/10/11
 ```

File: src/test/java/integration/DateConditionsTest.java [hunk 1]
```diff
@@ -0,0 +1,74 @@
...
+      .hasMessageContaining("Actual value: 2022/10/11");
+  }
+
+  @Test
+  public void failure_wrongDateFormat() {
+    assertThatThrownBy(() ->
+      $("#birthdate").shouldHave(date(LocalDate.of(2022, 11, 11), "yyyy-MM-dd"))
+    )
+      .isInstanceOf(ElementShould.class)
+      .hasMessageStartingWith("Element should have date value: \"2022-11-11\" (with date value format: \"yyyy-MM-dd\") {#birthdate}")
```
asolntsev, line 54: I recommend Java multi-line strings in such cases. Then you don't need to escape quotes to much:

```java
  .hasMessageStartingWith("""
     Element should have date value: "2022-11-11" (with date value format: "yyyy-MM-dd") {#birthdate}
     """)
```

src/main/java/com/codeborne/selenide/Condition.java
```diff
@@ -771,4 +779,166 @@ public String getName() {
...
+   * that is satisfied to the pattern and is equal to expected datetime
+   *
+   * <p>Sample: {@code $("input").shouldHave(datetime(expectedLocalDateTime, "dd/MM/yyyy HH:mm:ss"));}</p>
+   *
+   * @param expectedDateTime expected value of "value" attribute
+   * @param pattern datetime format of value of "value" attribute
+   */
+  @CheckReturnValue
+  @Nonnull
+  public static Condition datetime(LocalDateTime expectedDateTime, String pattern) {
```
asolntsev, line 874: Do you have integration tests that try to set/check time (or date+time)?

Try to add such inputs to `page_with_datetime_input.html`:

```html
      <input id="birthday" type="date" value="2022-10-11"/>
      <input id="birthtime" type="time" value="12:13:14"/>
      <input id="birthdate" type="datetime-local" value="2022-10-11 12:13:14"/>
```

src/main/java/com/codeborne/selenide/Condition.java
```diff
@@ -771,4 +779,166 @@ public String getName() {
...
+   * Assert that element contains "value" attribute with date value that is satisfied to the pattern and is equal to expected date
+   *
+   * <p>Sample: {@code $("input").shouldHave(date(expectedLocalDate, "dd/MM/yyyy"));}</p>
+   *
+   * @param expectedDate expected value of "value" attribute
+   * @param pattern date format of value of "value" attribute
+   */
+  @CheckReturnValue
+  @Nonnull
+  public static Condition date(LocalDate expectedDate, String pattern) {
```
asolntsev, line 793: I am thinking about adding so many overloaded methods to `Condition`. 
Probably we could add lesser number of those overloaded method by using parameter object like `ClickOptions`?

asolntsev (changes requested):
@Au6ojlut Thank you for the very high-quality pull request. Good job! :) 

Still, I have number of thoughs in comments. :)

Au6ojlut:
@asolntsev 
Thanks for the review

Sorry for the long wait, minor health issues...

This PR is not abandoned, I hope to fix everything next week

src/main/java/com/codeborne/selenide/conditions/LocalDateBetweenCondition.java
```diff
@@ -0,0 +1,60 @@
...
+  @CheckReturnValue
+  @Override
+  public CheckResult check(Driver driver, WebElement element) {
+    CheckResult formatted = formatCondition.check(driver, element);
+
+    if (formatted.verdict() == REJECT) return formatted;
+
+    LocalDate localDateValue = (LocalDate) formatted.actualValue();
+
+    if (localDateValue == null) throw new NullPointerException();
```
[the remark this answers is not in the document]
Au6ojlut: Done

src/test/java/integration/DateConditionsTest.java
Au6ojlut: Fixed

src/main/java/com/codeborne/selenide/Condition.java
Au6ojlut: There are already  `page_with_date_inputs.html` with date and `page_with_datetime_inputs.html` with datetime

There is no ability to check time condition for now...)

Or did I misunderstand your comment?

src/main/java/com/codeborne/selenide/Condition.java
Au6ojlut: I also thought about it but didn't find any good-looking solution.

The `Condition` class is like an entry point, where you can find all `pre-build` conditions. You just type `Condition.` and IDE shows all the possible variants. I think, that it's a good idea to have all condition in one place.


---

The best idea, that I have was intoducing some complex Condition, like

```Java
 class DateCondition {
   ...
   DateCondition withFormat(...)
   DateCondition withDate(...)
 }
 ```
 and then `public static DateCondition date( ?what is here, only date? )`.
 
 But, in my opinion, it's a little bit contra-intuitive, that there is a condition, that needs to be set up before use.
 And still there must be methods for `date`, `datetime`, `dateBetween`, `datetimeBetween`, `dateFormat` and `datetimeFormat`. It's just 4 methods lesser then with overloads...

Comment (Au6ojlut) [reply src/main/java/com/codeborne/selenide/Condition.java:793]: @asolntsev 
Oops.

Did you mean something like `public static Condition date(DateConditionOptions)`, where 
```Java
class DateConditionOptions {
   public static DateConditionOptions of(...) // common method, maybe with overloads 
   
   public static DateConditionOptions format(...) // for format check only
   
   public static DateConditionOptions beteween(DateConditionOptions, DateConditionOptions)
   
   ....
   
   public DateConditionOptions withFormat(...) //to be able to change format
}
```

asolntsev (approved):
Now I have the opposite feeling, like we have over-engineered a bit. :) 
But anyway, I will merge the PR. Later we'll see IF we get ideas how to improve the api.

@Au6ojlut  Thank you for the contribution!

Issue #2283, opened 2023-05-14
Escape special characters in SimpleReport **Subject section only**


Au6ojlut:
### Discussed in https://github.com/selenide/selenide/discussions/2279

<div type='discussions-op-text'>

<sup>Originally posted by **Au6ojlut** May 10, 2023</sup>
In case of  multi-line text condition the report looks a bit ugly
  ```
  +----------------------------------------------------+-------------------------------------------------+------------+------------+
  | Element                                            | Subject                                         | Status     | ms.        |
  +----------------------------------------------------+-------------------------------------------------+------------+------------+
  | open                                               | about:blank                                     | PASS       | 100        |
  | #any-long-ugly-selector-should-be-entirely-visible | should have(text "Not big
  but
  multiline")       | PASS       | 100        |
  | close                                              |                                                 | PASS       | 100        |
  +----------------------------------------------------+-------------------------------------------------+------------+------------+
  
  ```
  
  so, the idea is to escape all special characters in report, so it become smth like 
  
  ```
    +----------------------------------------------------+---------------------------------------------------+------------+------------+
  | Element                                            | Subject                                           | Status     | ms.        |
  +----------------------------------------------------+---------------------------------------------------+------------+------------+
  | open                                               | about:blank                                       | PASS       | 100        |
  | #any-long-ugly-selector-should-be-entirely-visible | should have(text "Not big\nbut\nmultiline")       | PASS       | 100        |
  | close                                              |                                                   | PASS       | 100        |
  +----------------------------------------------------+---------------------------------------------------+------------+------------+
  ```
  
  The replace rule is very simple
  ```
  \n -> \\n
  \t -> \\t
  \r -> \\r
  \f -> \\f
  and so on
  ```</div>

asolntsev:
Fixed in PR https://github.com/selenide/selenide/pull/2284

PR #2284, opened 2023-05-14
#2283 Escape special characters in subject

Au6ojlut:
For example, in case of new line in subject the report schema is broken and all the report starts look ugly.

We need to escape characters before the line length computing.

## Proposed changes

## Checklist


src/main/java/com/codeborne/selenide/logevents/SimpleReport.java
```diff
@@ -208,6 +218,36 @@ private static void checkThatSlf4jIsConfigured() {
...
+    for (int i = 0; i < text.length(); i++) {
+      var symbol = text.charAt(i);
+
+      switch (symbol) {
+        case '\t' -> builder.append("\\t");
+        case '\r' -> builder.append("\\r");
+        case '\n' -> builder.append("\\n");
+        case '\f' -> builder.append("\\f");
+        case '\b' -> builder.append("\\b");
+        default -> builder.append(symbol);
```
asolntsev, line 247: @Au6ojlut Maybe we should also escape all invisible characters below 32 (space)?

src/main/java/com/codeborne/selenide/logevents/SimpleReport.java
```diff
@@ -72,16 +72,26 @@ String generateReport(String title, List<LogEvent> events) {
...
-  private List<LogEventWithNestingLevel> computeNestingLevel(List<LogEvent> events) {
-    var stack = new ArrayDeque<LogEvent>();
+  private List<LogEventWithNestingLevel> escapeSubjectAndComputeNestingLevel(List<LogEvent> events) {
+    var stack = new ArrayDeque<LogEventWithNestingLevel.SimpleLogEvent>();
     var result = new ArrayList<LogEventWithNestingLevel>(events.size());
     for (LogEvent event : events) {
-      while (!stack.isEmpty() && stack.peekFirst().getEndTime() <= event.getStartTime())
+      var eventWithEscapedSubject = new LogEventWithNestingLevel.SimpleLogEvent(
+        event.getElement(),
+        escape(event.getSubject()),
```
asolntsev, line 81: At a first glance, the good moment to escape characters is not the construction of `SimpleLogEvent`, but when we render `SimpleLogEvent`. In other words, when we generate the resulting report. Then we could escape all fields, not only subject.

asolntsev (changes requested):

src/main/java/com/codeborne/selenide/logevents/SimpleReport.java
Au6ojlut: But we need to know the max size of a cell of the report table. If the calculation was made before the `escape` was applied then the final max size may be incorrect.

Au6ojlut: Since `element` and `subject` are just strings, I believe, that they are  the only place where special characters can appear. So, it's enough to escape them, not all the report line/row.

But, I also didn't find a case when the `element` could contain some special characters (only some synthetic example), that's why I left it as is, but it can be easily fixed.

src/main/java/com/codeborne/selenide/logevents/SimpleReport.java
Au6ojlut: Agree, I'll fix it

Au6ojlut: Done

asolntsev (approved):

PR #2367, opened 2023-07-10
Add basic webdriver cookie conditions

adorne:
Added three conditions for checking cookies: the presence of a cookie with a given name, the value of a cookie with a given name, and the presence of exactly the same Selenium Cookie object. Written checks are covered by unit tests.

### DIFF
--- a/src/main/java/com/codeborne/selenide/WebDriverConditions.java
+++ b/src/main/java/com/codeborne/selenide/WebDriverConditions.java
@@ -1,5 +1,6 @@
 package com.codeborne.selenide;
 
+import com.codeborne.selenide.conditions.webdriver.CookieObject;
 import com.codeborne.selenide.conditions.webdriver.CookieWithName;
 import com.codeborne.selenide.conditions.webdriver.CookieWithNameAndValue;
 import com.codeborne.selenide.conditions.webdriver.CurrentFrameUrl;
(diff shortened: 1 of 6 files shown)


src/main/java/com/codeborne/selenide/WebDriverConditions.java
```diff
@@ -72,4 +74,16 @@ public static ObjectCondition<WebDriver> numberOfWindows(int numberOfWindows) {
   public static ObjectCondition<WebDriver> title(String expectedTitle) {
     return new Title(expectedTitle);
   }
+
+  @CheckReturnValue
+  @Nonnull
+  public static ObjectCondition<WebDriver> cookieWithName(String name) {
```
rosolko, line 80: Maybe just `cookie` and `cookieValue`?
All of because `name` and `value` parameters obviously describe method

asolntsev: @adorne I agree that having one name `cookie` would be simpler:

* `shouldHave(cookie("session_id"));`
* `shouldHave(cookie("session_id", "9182827334645"));`

and maybe in future something like this:
* `shouldHave(cookie(new Cookie("session_id", "9182827334645")));`

Then it stays consistent e.g. with current attribute checks:
* `$.shouldHave(attribute("name"));`
* `$.shouldHave(attribute("name", "value"));`

asolntsev (changes requested):

src/main/java/com/codeborne/selenide/WebDriverConditions.java
adorne: Okay, I agree with the consistency argument, I'll fix it.

adorne: @asolntsev fixed.

asolntsev (approved):

PR #2369, opened 2023-07-14
feat: added new command to swipe right and left

amuthansakthivel:
## Proposed changes
We introduced the scrollTo element in the previous versions and thought the left and right directions were unnecessary. 

But there have been consistent requests to have the option to search for elements by swiping right and left.

I could have added to the existing scroll command but the readability would be awkward. Hence created a new swipe command which is what it needs to be called. 

Sample usage is updated in the readme.md

$(element).swipeTo().click();

## Checklist


modules/appium/src/main/java/com/codeborne/selenide/appium/commands/SelenideAppiumCommands.java
```diff
@@ -16,5 +16,6 @@ public SelenideAppiumCommands() {
     add("hideKeyboard", new HideKeyboard());
     add("scrollTo", new AppiumScrollTo());
     add("scroll", new AppiumScrollTo());
+    add("swipe", new AppiumSwipeTo());
```
OleksandrOwl, line 19: Чи потрібно сюди додати ще `add("swipeTo", new AppiumSwipeTo());` ?

amuthansakthivel: Fixed it. Good catch!

asolntsev: @amuthan-sakthivel You added only "swipe" command, but don't you also need to add "swipeTo" command?

asolntsev (changes requested):

modules/appium/src/main/java/com/codeborne/selenide/appium/commands/SelenideAppiumCommands.java
amuthansakthivel: @asolntsev - SwipeTo is already there

image

asolntsev (approved):

asolntsev:
@amuthansakthivel The build is failing. Can you fix it?

> Task :modules:appium:checkstyleMain
> Error: eckstyle] [ERROR] > AppiumSwipeOptions.java:3:65: Using the '.*' form of import should be avoided - com.codeborne.selenide.appium.AppiumSwipeDirection.*. [AvoidStarImport]

amuthansakthivel:
Pushed the fix @asolntsev

asolntsev (approved):

PR #2399, opened 2023-08-01
downgrade WebDriverManager from 5.4.x to 5.3.x to support Java8

asolntsev:
Starting from 5.4.0, WDM requires Java11+. But Selenide still supports Java8.

Let's upgrade to Java11 in major release Selenide 7.0.0

### DIFF
--- a/modules/core/build.gradle
+++ b/modules/core/build.gradle
@@ -16,7 +16,7 @@ sourceSets {
 dependencies {
   api('org.opentest4j:opentest4j:1.3.0')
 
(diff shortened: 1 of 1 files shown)


BorisOsipov:
@asolntsev 5.3 wdm doesn't support chrome more than 114.

asolntsev:
> @asolntsev 5.3 wdm doesn't support chrome more than 114.

Hm... In some sense, it's even good news.
It means that we don't have other options than just require Java 11. :)

Issue #2385, opened 2023-07-25
UnsupportedClassVersion exception after upgrade to Selenide 6.16.0


ccea-dev-team:
## The problem
Hi, I run my tests using Java 8 and had no issues with Selenide version 6.15.0.  After upgrading to 6.16.0 I'm getting an UnsupportedClassVersion exception when trying to run tests - requires class version 55 (Java 11) but environment only supports up to 52 (Java 8).  Is there something in the 6.16.0 release that requires Java 11 to run?

## Tell us about your environment
* **Selenide Version**:  6.16.0
* **Chrome\Firefox\IE Version**:  115
* **Browser Driver Version**: 114
* **Selenium Version**:  4.10
* **OS Version**:  win10

asolntsev:
@ccea-dev-team Yes, it was unintentional change in 6.16.0.
We upgraded to WebDriverManager 5.4.0 which started requiring Java11+. 
One simple workaround is to downgrade to WebDriverManager 5.3.3. 

BUT

Is it really a good idea? I believe it's much better to upgrade to Java 11+. Selenium will stop supporting Java8 very soon.

ccea-dev-team:
Hi,

Thanks for the reply.  I think this will have to be a decision on your side if you want to keep supporting Java 8.  A lot of people are still using Java 8 (especiallyin organistions that are not using cutting edge tech) and it is offically supported until 2027 or so.

We do plan to upgrade to Java11 or 17 sometime soon so we can work around this for now, however in the future If you do not offer Java 8 support there should be somewhere mentioning Selenide requires Java11+ in the docs/FAQ.

Thank you, 

Brian

asolntsev:
@ccea-dev-team  I agree, Selenide 6.x should still support Java8.
I downgraded WDM, see https://github.com/selenide/selenide/pull/2399

Will be released as Selenide 6.17.0

asolntsev:
@ccea-dev-team Oops, it seems we just cannot fix this issue. 
1. WDM 5.4.x requires Java 11+
2. WDM 5.3 doesn't support chrome 115+.

It means that we don't have other options than just require Java 11. 

As a workaround, anyone who needs to run tests with Java 8, can just exclude WebDriverManager from dependencies (or use older version of WDM), and set webdriver version explicitly.

PR #2400, opened 2023-08-01
#2385 replace WebDriverManager by Selenium built-in SeleniumManager

asolntsev:
Starting from 5.4.0, WDM requires Java11+. But Selenide still needs to support Java8.


modules/selenoid/build.gradle
```diff
@@ -7,6 +7,7 @@ dependencies {
   testImplementation "org.seleniumhq.selenium:selenium-http-jdk-client:$seleniumVersion"
   implementation("org.apache.httpcomponents.client5:httpclient5:$httpClientVersion")
   implementation("commons-io:commons-io:2.13.0")
+  implementation 'com.fasterxml.jackson.core:jackson-databind:2.15.2'
```
rosolko, line 10:
```suggestion
  implementation("com.fasterxml.jackson.core:jackson-databind:2.15.2")
```

src/main/java/com/codeborne/selenide/SelenideConfig.java
```diff
@@ -17,7 +17,7 @@ public class SelenideConfig implements Config {
   private String browserSize = getProperty("selenide.browserSize", "1366x768");
   private String browserVersion = getProperty("selenide.browserVersion", null);
   private String browserPosition = getProperty("selenide.browserPosition", null);
-  private boolean driverManagerEnabled = Boolean.parseBoolean(getProperty("selenide.driverManagerEnabled", "true"));
+  private boolean driverManagerEnabled = Boolean.parseBoolean(getProperty("selenide.driverManagerEnabled", "false"));
```
rosolko, line 20: `false` by default, this line can be simply deleted.

src/main/java/com/codeborne/selenide/webdriver/DriverFactory.java
```diff
@@ -12,7 +12,8 @@
 import java.io.File;
 
 public interface DriverFactory {
-  void setupWebdriverBinary();
+  default void setupWebdriverBinary() {
```
rosolko, line 15: Do we really need this method in interface?

rosolko (approved):

src/main/java/com/codeborne/selenide/webdriver/DriverFactory.java
asolntsev: @rosolko  I was thinking that it's needed because some people might implement their own `DriverFactory`.
But now I see that we really can just remove this code. It would not be called anyway.

src/main/java/com/codeborne/selenide/SelenideConfig.java
asolntsev: Yes, I've deleted it.

Issue #2459, opened 2023-09-04
Method `$.type()` enters only the last characters in Android


asolntsev:
## The problem
Method `$("foo").type(TypeOptions.text("1234").clearFirst(false))` doesn't just enter text "1234". Instead, it:
1. enters "1"
2. removes "1",
3. enters "2"
4. removes "2"
5. enters "3"
6. etc.

In the end, the field contains only text "4" (but should contain "1234").

## Tell us about your environment
* **Selenide Version**: 6.17.2
* **OS Version**: Android

## Code To Reproduce Issue

```java
options.setAppPackage("com.google.android.dialer");
options.setAppActivity("com.google.android.dialer.extensions.GoogleDialtactsActivity");

SelenideAppium.$(By.id("com.google.android.dialer:id/search_box_start_search")).click();
SelenideAppium.$(By.id("com.google.android.dialer:id/search_view")).type(TypeOptions.text("1234").clearFirst(false));
```

asolntsev:
As far as I see, it's rather a bug in Appium. 
When an input field already contains value "1", and I call method `$.sendKeys("2")`, it REPLACES "1" by "2". 
(But webdriver spec says `sendKeys` should APPEND character).

In Appium log I see that Appium translates `sendKeys` method to `AppiumDriver.setValue()` call:

```
[HTTP] --> POST /wd/hub/session/3d6f10a8-af06-4435-b5c3-8aec9ec0f233/element/00000000-0000-05e4-ffff-ffff00000033/value
[HTTP] {"id":"00000000-0000-05e4-ffff-ffff00000033","text":"2","value":["2"]}
[debug] [AndroidUiautomator2Driver@3a42 (3d6f10a8)] Calling AppiumDriver.setValue() with args: [["2"],"00000000-0000-05e4-ffff-ffff00000033","3d6f10a8-af06-4435-b5c3-8aec9ec0f233"]
[debug] [AndroidUiautomator2Driver@3a42 (3d6f10a8)] Matched '/element/00000000-0000-05e4-ffff-ffff00000033/value' to command name 'setValue'
[debug] [AndroidUiautomator2Driver@3a42 (3d6f10a8)] Added 'value' property ["2"] to 'setValue' request body
[...]

PR #2544, opened 2023-11-04
fix: type command issue

amuthansakthivel:
## Proposed changes
Fixes the issue - https://github.com/selenide/selenide/issues/2459

Sendkeys work different on android and ios.
For Android we have to use actions.sendKeys and for ios we can use default sendKeys (action.sendkeys is not working in ios)
Added tests for safety

## Checklist

### DIFF
--- a/modules/appium/src/main/java/com/codeborne/selenide/appium/commands/AppiumType.java
+++ b/modules/appium/src/main/java/com/codeborne/selenide/appium/commands/AppiumType.java
@@ -1,19 +1,50 @@
 package com.codeborne.selenide.appium.commands;
 
+import com.codeborne.selenide.SelenideElement;
+import com.codeborne.selenide.TypeOptions;
+import com.codeborne.selenide.appium.AppiumDriverRunner;
 import com.codeborne.selenide.commands.Type;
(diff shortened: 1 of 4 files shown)


modules/appium/src/main/java/com/codeborne/selenide/appium/commands/AppiumType.java
```diff
@@ -1,21 +1,53 @@
...
+    typeIntoField(locator, element, typeOptions);
+    return proxy;
+  }
+
+  private void typeIntoField(WebElementSource locator, WebElement element, TypeOptions typeOptions) {
+    Actions actions = new Actions(locator.driver().getWebDriver());
+    for (int i = 0; i < typeOptions.textToType().length(); i++) {
+      CharSequence character = typeOptions.textToType().subSequence(i, i + 1);
+      if (AppiumDriverRunner.isAndroidDriver()) {
+        element.click();
```
amuthansakthivel, line 42: I tried to use actions.click().sendKeys(character).perform() but the SauceLabLoginTest.canTypeTextSlowly was failing. Hence I used this way.

asolntsev (approved):

Issue #2130, opened 2023-01-19
Add conditions oneOfTexts, oneOfExactTexts, oneOfTextsCaseSensitive, oneOfExactTextsCaseSensitive


ilya-corp:
## The problem
Currently it's not very easy to check if text of an element matches to TEXT1 or to TEXT2. For example, if I need to ensure that error message with text "Critical error" or "Fatal situation" appeared, but it is not "Trivial error", I need to made custom conditions and apply them.
Looks like it's goog idea to have same conditions from out of the box.
Briefly describe the issue you are experiencing (or the feature you want to see added to Selenide).
Tell us what you were trying to do and what happened instead. 

## Details
Usage examples:
$(By.cssSelector(".error")).shouldHave(Condition.oneOfTexts("Gipar is died", "Elephant is to big", "Wolf is not so pretty")).shouldBe(Condition.visible);

$(By.cssSelector(".problem")).shouldHave(Condition.oneOfTexts("Requires lowercase symbols", "Requires uppercase symbols")).shouldBe(Condition.visible);

So, the number of arguments may be 1+.

OR we need to add described functionality to existed conditions without creating new conditions. If only 1 argument is provided - work as expected, if there are 2 or more arguments - work like described there.

The alternative way is using Condition.or or creating custom condition with the Condition.or under the hood.
The cons: requires more long strings; requires creating custom conditions for every your project.

## Tell us about your environment
* **Selenide Version**: 6.11
* **Chrome\Firefox\IE Version**: Any
* **Browser Driver Version**: Have no matter
* **Selenium Version**:  Have no matter
* **OS Version**:  Have no matter

PR #2557, opened 2023-11-18
#2130 oneOfTexts() - like conditions

Achitheus:
Hi everybody!

## Proposed changes
#2130

1. Four conditions added:
    - oneOfTexts(Collection\<String>)
    - oneOfTextsCaseSensitive(Collection\<String>)
    - oneOfExactTexts(Collection\<String>)
    - oneOfExactTextsCaseSensitive(Collection\<String>)
1. Unit tests added.

1. CHANGELOG.md is updated (but I'm not sure I did that correctly).

Feel free to ask questions and suggest corrections!

## Checklist
(my comment: `gradle check` fails with checkstyle errors because of LF/CRLF, but git must be fixed it while commiting/pushing)
- [x] I have added necessary documentation

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Changelog
 
+## 7.0.3 (released ...)
+* add `oneOfTexts(Collection)`- like Conditions (#2130) -- thanks to Yury Yurchenko
+
 ## 7.0.2 (released 01.11.2023)
 * bump Selenium from 4.14.1 to 4.15.0 (#2540)
 
--- a/src/main/java/com/codeborne/selenide/Condition.java
+++ b/src/main/java/com/codeborne/selenide/Condition.java
@@ -25,6 +25,10 @@ import com.codeborne.selenide.conditions.IsImageLoaded;
 import com.codeborne.selenide.conditions.MatchAttributeWithValue;
 import com.codeborne.selenide.conditions.MatchText;
 import com.codeborne.selenide.conditions.NamedCondition;
+import com.codeborne.selenide.conditions.OneOfExactTexts;
+import com.codeborne.selenide.conditions.OneOfExactTextsCaseSensitive;
+import com.codeborne.selenide.conditions.OneOfTexts;
+import com.codeborne.selenide.conditions.OneOfTextsCaseSensitive;
 import com.codeborne.selenide.conditions.Or;
 import com.codeborne.selenide.conditions.OwnText;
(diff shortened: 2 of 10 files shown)


asolntsev (approved):

asolntsev:
@Achitheus I already merged the pull request, but then started thinking :)

I have a question: why do you pass expected text as a collection?
 It makes my tests longer because I have to write `List.of`:
```java
$("#upper-case").shouldHave(oneOfTexts(List.of("foo", "this is sparta!", "bar")));
```

If we passed the expected texts as varargs, the test would be shorter (without `List.of`):
```java
$("#upper-case").shouldHave(oneOfTexts("foo", "this is sparta!", "bar"));
```

Achitheus:
> @Achitheus I already merged the pull request, but then started thinking :)
> 
> I have a question: why do you pass expected text as a collection? It makes my tests longer because I have to write `List.of`:
> 
> ```java
> $("#upper-case").shouldHave(oneOfTexts(List.of("foo", "this is sparta!", "bar")));
> ```
> 
> If we passed the expected texts as varargs, the test would be shorter (without `List.of`):
> 
> ```java
> $("#upper-case").shouldHave(oneOfTexts("foo", "this is sparta!", "bar"));
> ```

Do you mean it makes *writing* tests longer?

To be honest I dont see an objectively correct answer to the question, that could satisfy all users. Looks like we need another overloaded method :D
I was thinking about cases with data providers via MethodSource. 
If we take Collection approach and for some reason we have an array we should do the following:
`$.shouldHave(oneOfTexts(Arrays.toList(targetArray)));`
If we want to do it in your way and there is a List for some reason then:
`$.shouldHave(oneOfTexts(targetList.toArray(new String[0])));`
I dont like both conversion processes :)

I see two main questions here
1. *What* users prefer to pass from data providers?
2. How often they write asserts just “by hand”, as in your example, when “hard-coded” strings are used?

I don't have enough experience to answer these questions, so I prefer to leave the solution to you or ask the community.  
Sir :)

asolntsev:
@Achitheus Almost always, I personally write asserts "just by hand", and I believe this is the right way to write asserts. All kinds of data providers, factories, excels etc. bring complexity to the tests without a real need. Complexity should be avoided. 

If somebody pass data from MethodSource, they can return Array as easily as List. That's why I prefer array to list.

asolntsev:
I created a follow-up PR: https://github.com/selenide/selenide/pull/2568

PR #2559, opened 2023-11-20
Chain element from page object item

rosolko:
## Proposed changes
Add test for bug reproduction related to element chaining from page object item.  

## Checklist


rosolko:
Same behavior on top of pure selenium

PR #2593, opened 2023-12-18
Add getSearchLocator command

rosolko:
## Proposed changes
`$x("//div[id='sample']").getSearchLocator()` execution will return `//div[id='sample']` string.

## Checklist

### DIFF
--- a/src/main/java/com/codeborne/selenide/SelenideElement.java
+++ b/src/main/java/com/codeborne/selenide/SelenideElement.java
@@ -1257,6 +1257,16 @@ public interface SelenideElement extends WebElement, WrapsDriver, WrapsElement, 
   String getSearchCriteria();
 
   /**
+   * Return locator of criteria by which this element is located
+   *
+   * @return e.g. "//div[data-test-id='sample']"
+   * @see com.codeborne.selenide.commands.GetSearchLocator
+   */
+  @CheckReturnValue
+  @Nonnull
+  String getSearchLocator();
+
+  /**
    * @return the original Selenium {@link WebElement} wrapped by this object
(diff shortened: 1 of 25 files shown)


BorisOsipov:
Why do we need it? It can only work for a very limited number of cases, as I can imagine.

Obviously it cannot return a valid selector in all cases, but if we implement it, I would like to see as much functionality as possible.

Now even the simple case `$x("//div[id='sample']").$x("//div[id='baz']").getSearchLocator()` does not seem to return the correct selector, but the last one `"//div[id='baz']"` that is wrong.

Also 
```
val foo = $("ss").toWebElement();
$(foo).getSearchLocator();
```
will result in a stackoverflow exception :D

rosolko:
Yes, only for corner cases. One of it - containers for mobile platforms where you have no ability to wrap items by contanainer using annotation-style locator definition. This case only root-element contstuctor with direct locator access approach works.

For chaining cases it nearly impossible build full locator because you cae mix strategies thats why I decide to limit this ability  for latest element only.

rosolko:
I forgot to mention, it was just simply replaced by custom command:

```java
public static final Command<String> getLocator = new Command<>() {
    @Nullable
    @Override
    public String execute(@NonNull SelenideElement proxy, @NonNull WebElementSource locator, @Nullable Object[] args) {
        Object criteria;
        try {
            criteria = FieldUtils.readField(locator, "criteria", true);
        } catch (IllegalAccessException e) {
            throw new RuntimeException("Unable to get locator criteria", e);
        }

        final By.Remotable remotable = (By.Remotable) criteria;
        final By.Remotable.Parameters remoteParameters = remotable.getRemoteParameters();
        return String.valueOf(remoteParameters.value());
    }
};
```

Issue #2617, opened 2024-01-24
User can safely call `addRequestFilter`/`addResponseFilter` with the same filter


asolntsev:
## The problem
Every time when I call `open(url)` in my tests, I want to add a proxy filter. 
But Selenide doesn't allow adding multiple filters with the same name.
So I need to add an IF like this:

```java
open();
if (getSelenideProxy().requestFilter("request-logger") == null) {
  getSelenideProxy().addRequestFilter("request-logger", MY_LOGGER);
}
```

## Solution
Method `addRequestFilter` could silently ignore without throwing an error IF the filter is exactly the same instance as previously.
Then I could safely add `getSelenideProxy().addRequestFilter("request-logger", MY_CONSTANT_LOGGER);` all over my tests.

PR #2630, opened 2024-01-28
#2617 User can safely add the same proxy filter many times

asolntsev:
it's convenient to store a filter into a constant, and add it before every `open(url)` - without checking if it was already added earlier.

```
var MY_CONSTANT_LOGGER = new RequestFilter() {...}
getSelenideProxy().addRequestFilter("request-logger", MY_CONSTANT_LOGGER);
```

### DIFF
--- a/src/main/java/com/codeborne/selenide/proxy/SelenideProxyServer.java
+++ b/src/main/java/com/codeborne/selenide/proxy/SelenideProxyServer.java
@@ -116,15 +116,14 @@ public class SelenideProxyServer {
    * @param requestFilter the filter
    */
   public void addRequestFilter(String name, RequestFilter requestFilter) {
-    if (isRequestFilterAdded(name)) {
+    RequestFilter currentFilter = requestFilter(name);
+    if (currentFilter != null && !currentFilter.equals(requestFilter)) {
       throw new IllegalArgumentException("Duplicate request filter: " + name);
     }
-    proxy.addRequestFilter(requestFilter);
-    requestFilters.put(name, requestFilter);
-  }
-
-  private boolean isRequestFilterAdded(String name) {
-    return requestFilters.containsKey(name);
+    if (currentFilter == null) {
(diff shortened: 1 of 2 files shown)


rosolko (approved):

src/main/java/com/codeborne/selenide/proxy/SelenideProxyServer.java
```diff
@@ -116,15 +116,14 @@ public boolean isStarted() {
    * @param requestFilter the filter
    */
   public void addRequestFilter(String name, RequestFilter requestFilter) {
-    if (isRequestFilterAdded(name)) {
+    RequestFilter currentFilter = requestFilter(name);
+    if (currentFilter != null && !currentFilter.equals(requestFilter)) {
```
rosolko, line 120: Maybe better to use such verification and skip next `currentFilter` null check as well?
```suggestion
    if (Objects.equals(currentFilter, requestFilter)) {
```

src/main/java/com/codeborne/selenide/proxy/SelenideProxyServer.java
```diff
@@ -134,11 +133,14 @@ private boolean isRequestFilterAdded(String name) {
    * @param responseFilter the filter
    */
   public void addResponseFilter(String name, ResponseFilter responseFilter) {
-    if (responseFilters.containsKey(name)) {
+    ResponseFilter currentFilter = responseFilter(name);
+    if (currentFilter != null && !currentFilter.equals(responseFilter)) {
```
rosolko, line 137: Same here

src/main/java/com/codeborne/selenide/proxy/SelenideProxyServer.java
asolntsev: @rosolko Nice try, but no :)
Then this method will throw exception if `currentFilter == null`. But should not.

Issue #2638, opened 2024-02-04
Add method to unfocus the element


asolntsev:
## The problem
In tests, we often need to fill an input field that has some post-processing (e.g. card number, address auto-completion or validation etc.). 
For such input fields, I need to enter the value and then _loose the focus_ from the element. 

As a workaround, I can call `$.pressTab()`, but it might have an unexpected side-effect: some other element will get focused. 

## Solution
I would like to have method like `$.unfocus()` which is equivalent to clicking an empty spaces between elements.
Then the element looses focus, and none of other elements gets focused.

## Tell us about your environment
* **Selenide Version**: 7.0.6

## Code To Reproduce Issue
```java
$("#card-number").sendKeys("1111-222-33-4444");
$("#card-number").unfocus();
```

PR #2639, opened 2024-02-04
#2638 Add the ability to unfocus an element

asolntsev:
Add method `$.unfocus()`


rosolko (approved):

BorisOsipov (commented):
It will be nice to override this command for appium with a meaningful message that is only for web and is not implemented for mobile.

Issue #2590, opened 2023-12-15
Check if element is visible within a timeout without catching an error


andreisuslov:
## The problem
I propose the implementation of a method in Selenide that allows checking if an element is visible within a specified timeout, without throwing an error if the element does not become visible.
Currently, methods like `shouldBe` combined with visibility conditions throw an exception if the condition is not met within the specified timeout. While this behavior is suitable for many test scenarios, there are cases where a non-exception-based check would be more appropriate.

## Details
A method that returns a boolean value indicating whether an element is visible/enabled/etc within a specified timeout. For instance:
```java
boolean isVisible = $("selector").is(Condition.visible, 10000);
```

At the moment, something like this could be created in the helper methods locally in my framework:
```java
boolean isVisible;
try {
    $(selector).shouldBe(Condition.visible, Duration.ofSeconds(timeout));
    isVisible = true;
} catch (ElementNotFound e) {
    isVisible = false;
}
```
, but it appears to me that other developers would also benefit from having this functionality implemented natively in Selenide.

asolntsev:
@andreisuslov The question is: what are you going to do with this `isVisible` variable?

P.S. For background information: https://selenide.org/2019/12/02/advent-calendar-how-to-abuse-selenide/

andreisuslov:
Thank you for your work, by the way, it's greatly appreciated. I still would like to address the concerns you raised regarding the proposal for a non-exception-based visibility check in Selenide. The goal here isn't to toss out the whole idea of solid, dependable testing. It's more about flexibility, making things more adaptable and easier to read and use, especially in some specific test cases.

My app is slow. It has dynamic content, and the presence of certain elements may not be guaranteed. The proposed method allows for a more nuanced approach to handle these cases. The tests need to be adjusted for that as the app will likely not get faster. We sometimes need to click onto an element if it's visible, so we wait for it to load. As you pointed out, it can be slow and inefficient but it is what it is - the app under test won't get faster. In my view, this is not about bypassing critical failures but about managing conditional flows within tests.

Imagine a web application that has feature toggles, which might or might not be present depending on the user's settings or A/B testing scenarios.

With the proposed method, you could easily handle this in Groovy as follows:
```groovy
def checkFeatureToggle() {
    def isFeatureVisible = $("div.toggle").is(Condition.visible, 10000)
    if (isFeatureVisible) {
        // actions related to the feature toggle
        println("feature is available")
        // more code...
    } else {
        // handle the absence of the feature
[...]

PR #2640, opened 2024-02-04
#2590 Add method `$.is(visible, timeout)`

asolntsev:
This is the historical moment: finally, we allow writing IFs in tests ¯¯\_(ツ)_/¯¯


rosolko (approved):
I'm not really a fan of this change

asolntsev:
> I'm not really a fan of this change

#metoo  :)

BorisOsipov (approved):

Issue #1372, opened 2021-01-15
Add posibility to set different capabilities for each browser session in Configuration class


ivandros:
## The problem
For running tests in Selenoid containers we set browser capabilities via Configuration.browserCapabilities = capabilities;
Like in this sample: https://github.com/selenide-examples/selenoid-sample/blob/main/src/test/java/org/selenide/examples/selenoid/FileDownloadTest.java

The problem is that in this case all simultaneous browser session in Selenoid have one the same set of capabilities.
Сouse: Configuration.browserCapabilities is static!

## Details
1. There is necessity to set in some browsers different settings. E.g. some tests need to run in browser with enabled mobile mode.
But now all browsers work only in desktope mode.

2. Selenoid has useful feature which allows to set throw capabilities unique name for each browser session and name for recorded video-file by each running test. (https://aerokube.com/selenoid/latest/#_special_capabilities). And we can do it unique and set via the code.
But now all browser session and videofile in Selenoid has the same name that complicates its identification.

3. Similar problems form other users:
 https://t.me/selenide_ru/702
 https://t.me/selenide_ru/2450
 https://t.me/selenide_ru/4601
 https://t.me/selenide_ru/5663
 https://t.me/selenide_ru/6503
 https://t.me/selenide_ru/9409
 https://t.me/selenide_ru/9509
 https://t.me/selenide_ru/16572
 https://t.me/selenide_ru/18035
 

## Tell us about your environment
* **Selenide Version**: 5.16.0
* **Chrome\Firefox\IE Version**: selenoid/vnc:chrome_87.0

## Code To Reproduce Issue

```java
 public class TestBase(){
   boolean isMobileDevice = false;   

   @BeforeMethod
   initBrowserCapabilities(this.getClass().getSimpleName(), isMobileDevice);
 
   @AfterMethod
    closeWebDriver();
}

    public class BrowserCapabilities {
    public void initBrowserCapabilities(String currentClassName, boolean isMobileDevice) {
       private String videoIdentification;
       videoIdentification = currentClassName + "_" + new SimpleDateFormat("yyyy-MM-dd_HH:mm:ss").format(new Date()).mp4";

            ChromeOptions chromeOptions = new ChromeOptions();
            DesiredCapabilities capabilities = new DesiredCapabilities();
            capabilities.setBrowserName("chrome");
            capabilities.setCapability("enableLog", true);

                capabilities.setCapability("enableVNC", true);
[...]

PR #2666, opened 2024-03-01
add setDriver method to WebDriverRunner

PetroOv:
## Proposed changes
Selenide already has methods to create a new browser on demand, but it doesn't allow us to change their configuration or we could create our own WebDriver, but in this case, additional selenide functionality wouldn't be applied. So, I introduced a method that creates a new browser but is fully controlled by selenide (listeners, deadwatch etc).

## Checklist

### DIFF
--- a/statics/src/main/java/com/codeborne/selenide/Selenide.java
+++ b/statics/src/main/java/com/codeborne/selenide/Selenide.java
@@ -114,19 +114,22 @@ public class Selenide {
   }
 
   /**
-   * Replace browser with a new one, using given Config
-   * Please note that config would affect only browser startup props.
+   * If browser window was already opened before, it will be reused. Config will be ignored.
+   * Otherwise, a new browser window will be opened.
+   * Please note:
+   * config will be ignored if browser window was already opened before.
+   * that config would affect only browser startup props.
    * Such as: browser, browserCapabilities, remote, browserSize, browserVersion, startMaximized, etc.
    * Global configs will be the same (timeout, pollingInterval, baseUrl, etc.).
    * Also default WebDriverRunner setup would be applied to your browser: proxy, listeners
    */
   public static void open(String url, Config config) {
-    WebDriverRunner.replaceBrowser(config);
+    WebDriverRunner.getOrCreateNewBrowser(config);
     open(url);
   }
 
   public static void open(Config config) {
-    WebDriverRunner.replaceBrowser(config);
+    WebDriverRunner.getOrCreateNewBrowser(config);
   }
 
   /**
--- a/statics/src/main/java/com/codeborne/selenide/Selenide.java
+++ b/statics/src/main/java/com/codeborne/selenide/Selenide.java
@@ -1,9 +1,24 @@
 package com.codeborne.selenide;
 
+import static com.codeborne.selenide.ModalOptions.withExpectedText;
+import static com.codeborne.selenide.WebDriverRunner.getSelenideDriver;
+import static com.codeborne.selenide.logevents.SelenideLogger.getReadableSubject;
+
 import com.codeborne.selenide.ex.DialogTextMismatch;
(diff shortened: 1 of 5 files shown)


BorisOsipov (changes requested):
We have already discussed this [here](https://github.com/selenide/selenide/issues/1372), and it is a good api decision to allow users to create new drivers with config via the `Selenide.open(config)` api.

`WebDriverRunner.setDriver(config);` api looks confused. It looks like a setter, but it does different work.

PetroOv:
> We have already discussed this [here](https://github.com/selenide/selenide/issues/1372), and it is a good api decision to allow users to create new drivers with config via the `Selenide.open(config)` api.
> 
> `WebDriverRunner.setDriver(config);` api looks confused. It looks like a setter, but it does different work.

Ok, could it be like 2 methods in Selenide class.
Selenide#open(java.lang.String, com.codeborne.selenide.Config) 
Selenide#open(com.codeborne.selenide.Config) 

And WebDriverRunner method rename from setDriver() to replaceBrowser()
What do u think, does it sound good?

BorisOsipov:
>Ok, could it be like 2 methods in Selenide class.
>Selenide#open(java.lang.String, com.codeborne.selenide.Config)
>Selenide#open(com.codeborne.selenide.Config)

It seems good for first iteration.

>And WebDriverRunner method rename from setDriver() to replaceBrowser()
>What do u think, does it sound good?

There are at least two cases:
1) You want to create a driver with a specific configuration, but you don't want to mutate the global `Configuration' object.
2) You want to close the current driver and start a new one with a specific configuration, but you don't want to mutate the global `Configuration` object.

Question 1:

Should `Selenide#open(com.codeborne.selenide.Config)` close the current driver if it already exists?

For case 2 it is obvious yes. For case 1 it is not clear. 

From my point of view, it must not touch the previous driver.
1) As a user I don't expect `Selenide#open(com.codeborne.selenide.Config)` to close the previous one. I may have references to driver outside selenide for some reason and may want to work with them.
2) `com.codeborne.selenide.WebDriverRunner#setWebDriver(org.openqa.selenium.WebDriver)` does not close the previous one and silently change the current driver to provided.
3) It is not a problem at all to call `closeWebdriver()` if  you need and then call `open(config)`. 

Are there any insights you may have that I could miss for closing previous drivers?

PetroOv:
> > Ok, could it be like 2 methods in Selenide class.
> > Selenide#open(java.lang.String, com.codeborne.selenide.Config)
> > Selenide#open(com.codeborne.selenide.Config)
> 
> It seems good for first iteration.
> 
> > And WebDriverRunner method rename from setDriver() to replaceBrowser()
> > What do u think, does it sound good?
> 
> There are at least two cases:
> 
> 1. You want to create a driver with a specific configuration, but you don't want to mutate the global `Configuration' object.
> 2. You want to close the current driver and start a new one with a specific configuration, but you don't want to mutate the global `Configuration` object.
> 
> Question 1:
> 
> Should `Selenide#open(com.codeborne.selenide.Config)` close the current driver if it already exists?
> 
> For case 2 it is obvious yes. For case 1 it is not clear.
> 
> From my point of view, it must not touch the previous driver.
> 
> 1. As a user I don't expect `Selenide#open(com.codeborne.selenide.Config)` to close the previous one. I may have references to driver outside selenide for some reason and may want to work with them.
> 2. `com.codeborne.selenide.WebDriverRunner#setWebDriver(org.openqa.selenium.WebDriver)` does not close the previous one and silently change the current driver to provided.
> 3. It is not a problem at all to call `closeWebdriver()` if  you need and then call `open(config)`.
> 
> Are there any insights you may have that I could miss for closing previous drivers?

You are right, new open(Config) method should be aligned with the existing ones
So, we need to use the old browser if possible. As a solution, I decided to add newBrowser method which is similar to InNewBrowser. The only thing that bothers me, now we have cases where config would be ignored.

PR #2672, opened 2024-03-04
fix before/after events order

PetroOv:
## Proposed changes
This pr fix before/after events order if the browser is closed. Now before and after events in the same order
before changes:
Screenshot 2024-03-04 at 10 57 38
after changes:
Screenshot 2024-03-04 at 11 08 46

## Checklist

### DIFF
--- a/src/main/java/com/codeborne/selenide/drivercommands/Navigator.java
+++ b/src/main/java/com/codeborne/selenide/drivercommands/Navigator.java
@@ -1,5 +1,11 @@
 package com.codeborne.selenide.drivercommands;
 
+import static com.codeborne.selenide.AuthenticationType.BASIC;
+import static com.codeborne.selenide.FileDownloadMode.PROXY;
+import static com.codeborne.selenide.drivercommands.BasicAuthUtils.appendBasicAuthToURL;
+import static com.codeborne.selenide.drivercommands.BasicAuthUtils.registerBasicAuth;
+import static java.util.regex.Pattern.DOTALL;
+
 import com.codeborne.selenide.AuthenticationType;
 import com.codeborne.selenide.BasicAuthCredentials;
 import com.codeborne.selenide.Config;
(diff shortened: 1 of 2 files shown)


asolntsev:
@PetroOv Well, it's not a bug... The initial idea was that "create" is a sub-step of "open".
Look, in Selenide report it looks like an internal steps:

image

PetroOv:
> @PetroOv Well, it's not a bug... The initial idea was that "create" is a sub-step of "open". Look, in Selenide report it looks like an internal steps:
> 
> image

Hmm, I missed this report view. It makes sense to have such tracing. 
But, In my use case, it seems like a bug. I used LogEventListener to add an extra header/cookie before opening any URL. I noticed that the listener works fine only if a browser is already created. 
 I am not sure if LogEventListener was designed for such a purpose. I also tried Selenium's WebListener, but it operates in conjunction with EventFiringDriver, and this combination significantly slows down the tests, approximately doubling the execution time.

asolntsev:
> I am not sure if LogEventListener was designed for such a purpose

Pretty sure, LogEventListener was **not** designed to add cookies. It's intended _only for logging_. 
To add a cookie, you can run code like this:
```
open();
addCookie();
open(your url);
```

P.S. But we can still change the order as you suggested, why not.

PetroOv:
> > I am not sure if LogEventListener was designed for such a purpose
> 
> Pretty sure, LogEventListener was **not** designed to add cookies. It's intended _only for logging_. To add a cookie, you can run code like this:
> 
> ```
> open();
> addCookie();
> open(your url);
> ```
> 
> P.S. But we can still change the order as you suggested, why not.

> open();
> addCookie();
> open(your URL)

we are using cookies to route users/test (blue-green deployment), if we add addCookie() explicitly it would generate a lot of boilerplate code.

This change could be very useful for our project because it allows to replace all Selenium listeners.

asolntsev:
> because it allows to replace all Selenium listeners.

But WHY? Why do you want to replace Selenium listeners (that were invented exactly for this purpos) by Selenide log listener (that was invented for a different purpose - logging)? 
Why do you think it's a good idea?

asolntsev:
> Our basic open-page test runtime has increased from 1 to 1.5 seconds.

Can you show the code: how exactly are you adding "Selenium listener"? It seems this problem deserves an investigation.

PetroOv:
> > Our basic open-page test runtime has increased from 1 to 1.5 seconds.
> 
> Can you show the code: how exactly are you adding "Selenium listener"? It seems this problem deserves an investigation.

Our stack Selenide, testNg, Selenoid.
Selenium Listener is adding onStart(suite) event using WebDriverRunner.addListener method. Tests are running on Ci in parallel

`public class BrowserConfigListener implements ISuiteListener {
  private static UICustomHeaderListener uiCustomHeaderListener = UICustomHeaderListener.createInstance();

  @Override
  public void onStart(ISuite suite) {
    WebDriverRunner.addListener(uiCustomHeaderListener);
  }
}`

asolntsev:
@PetroOv And what's inside `UICustomHeaderListener`?

Could you please add few options to your tests and investigate why it's slower with listener?
See https://www.youtube.com/watch?v=4Kh5yppP8zY&ab_channel=DEVCLUB.EU&t=56m35s

PetroOv:
> https://www.youtube.com/watch?v=4Kh5yppP8zY&ab_channel=DEVCLUB.EU&t=56m35s



> @PetroOv And what's inside `UICustomHeaderListener`?
> 
> Could you please add few options to your tests and investigate why it's slower with listener? See https://www.youtube.com/watch?v=4Kh5yppP8zY&ab_channel=DEVCLUB.EU&t=56m35s

Here the code, problem reproduces even if https://pastebin.com/uDzBJWBK

asolntsev:
@PetroOv I don't see any evidence for your words. I tried to run few tests with and without your listener - they take the same time. 

I think there is something more in your code that makes the tests slower.
Or could you provide some reproducible example?
So that I could run `git clone && gradle test` and observe the problem.

PetroOv:
@asolntsev  hmmm, I couldn't reproduce the problem in the fresh repository,  maybe problem with another part of the code or the old browser version that was used previously.

A1exKH (approved):
LGTM.

PR #2687, opened 2024-03-18
New "or" condition

asolntsev:
Born from discussion https://github.com/selenide/selenide/discussions/2681

NB! I still don't recommend using `or` methods. This is a flag of bad test design. Consider splitting this "or" check into two different methods or tests.


rosolko (approved):
Ok, let's sync element and collection behavior!

PR #2728, opened 2024-04-25
Fix CDP download for custom browsers

PetroOv:
## Proposed changes
This PR fixes IllegalArgumentException for custom DriverProvider/DriverFactory when they support cdp

## Checklist

### DIFF
--- a/modules/grid/src/test/java/integration/DownloadFileFromGridWithCdpTest.java
+++ b/modules/grid/src/test/java/integration/DownloadFileFromGridWithCdpTest.java
@@ -1,14 +1,22 @@
 package integration;
 
+import com.codeborne.selenide.Browser;
+import com.codeborne.selenide.Config;
 import com.codeborne.selenide.Configuration;
 import com.codeborne.selenide.ex.FileNotDownloadedError;
 import com.codeborne.selenide.impl.FileContent;
+import com.codeborne.selenide.webdriver.FirefoxDriverFactory;
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.params.ParameterizedTest;
 import org.junit.jupiter.params.provider.ValueSource;
+import org.openqa.selenium.Proxy;
+import org.openqa.selenium.WebDriver;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import com.codeborne.selenide.webdriver.ChromeDriverFactory;
 
 import java.io.File;
 import java.io.IOException;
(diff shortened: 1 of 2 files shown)


modules/grid/src/test/java/integration/DownloadFileFromGridWithCdpTest.java
```diff
@@ -30,6 +17,26 @@
...
+import javax.annotation.Nullable;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.ValueSource;
+import org.openqa.selenium.Proxy;
+import org.openqa.selenium.WebDriver;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
```
asolntsev, line 39: @PetroOv Please avoid massive changing order of imports. Let them rest in peace.

modules/grid/src/test/java/integration/DownloadFileFromGridWithCdpTest.java
```diff
@@ -42,6 +49,7 @@ void openFileUploadForm() {
     Configuration.remote = gridUrl.toString();
     Configuration.browserCapabilities.setCapability("se:downloadsEnabled", true);
     Configuration.fileDownload = CDP;
+    closeWebDriver();
```
asolntsev, line 52: it will slow down all previous tests.
Let's close the browser only in one test.

src/main/java/com/codeborne/selenide/impl/DownloadFileWithCdp.java
```diff
@@ -126,17 +126,16 @@ private File waitUntilDownloadsCompleted(Driver driver, FileFilter fileFilter,
...
-      Optional<HasDevTools> cdpBrowser = cast(driver.getWebDriver(), HasDevTools.class);
-      if (cdpBrowser.isPresent()) {
-        DevTools devTools = cdpBrowser.get().getDevTools();
-        devTools.createSessionIfThereIsNotOne();
-        devTools.send(Page.enable());
-        return devTools;
-      }
+    Optional<HasDevTools> cdpBrowser = cast(driver.getWebDriver(), HasDevTools.class);
+    if (cdpBrowser.isPresent()) {
+      DevTools devTools = cdpBrowser.get().getDevTools();
```
asolntsev, line 131: The problem is that Firefox doesn't support downloading files via CDP.
We need to exclude Firefox in this IF. @PetroOv

asolntsev (changes requested):

src/main/java/com/codeborne/selenide/impl/DownloadFileWithCdp.java
PetroOv: I didn't know it( I returned isChromium check, but now we get browser name from capabilities of instance

Issue #807, opened 2018-09-20
Remove deprecated feature `Configuration.holdBrowserOpen`


asolntsev:
I suggest to remove this setting and feature. 
It's only usable for debugging on your work machine. But in this case you can put breakpoint or add logging anyway.

Please share your thoughts if you still need it. 

@eichisanden @SergeyPirogov Why the standard error message with screenshot and html is not enough?

simple-elf:
holdBrowserOpen is useful, not very often, but still useful

SergeyPirogov:
You are able to see network requests, also it's not always ok to set breakpoint at the last line of my test methos. In Such case I should add some dummy line to debug smt that happens after last line

asolntsev:
@SergeyPirogov Adding line `Configuration.holdBrowserOpen = true;` is also a "dummy line". How is it better?
But the bigger problem is: this approach doesn't help anyway when the test fails on Jenkins. And in most cases, tests fail on Jenkins.

SergeyPirogov:
The case that it's a config. And as end user I want ability to set this config, those who think this is useless junt not use it. Deletition of this line will not make selenide more comfortable to work with. 

Also to run tests in Jenkins, firstly we should write them locally, so ability to debug using holdBrowserOpen is much usefull in this case

asolntsev:
@SergeyPirogov Yes, this is config, and... so what?
It's hard to use product that has too many settings. We want to make people's life easier by removing as many settings as possible. 

I understand that you use this feature. But when developing/debugging a test, you can just add `sleep(666666)` in the end of your test. It's also simple.

SergeyPirogov:
The problem is that I dont' want to add anything to my tests. I want to have configuration via framework/library etc. TO delete config that is useful for people is a bad decision

eichisanden:
Rather than changing the test code, I think that it is more smart to tweak the setting.
However, I think that it may be deleted if you want to reduce the setting items.
Because holdBrowserOpen is useful, but not necessary.

asolntsev:
@SergeyPirogov @simple-elf @eichisanden Well, it seems that many people use this feature. One possible use case is to enable this setting in CI job and login to the CI machine remotely to investigate the browser state. 

So, we decide to keep `Configuration.holdBrowserOpen` feature.

PR #2730, opened 2024-04-28
deprecate setting "holdBrowserOpen" (#2730)

asolntsev:


ASTRO-CZ:
Hello will there be a replacement for this setting? I use it heavily for debugging to get better idea where the test ended. I also saw this old topic where the motion to remove this setting was also debated and eventually declined. https://github.com/selenide/selenide/issues/807

Issue #2755, opened 2024-05-30
Soft Assert mode with a FAIL step is not marking the whole test as FAILED


jensenbrian:
## The problem
In Soft Assert mode with a FAIL step is not marking the whole test as FAILED

## Details
Running the `TestNgSoftTest.textArea()` in SOFT mode with an added line that will fail. The build is marked as "BUILD SUCCESSFUL". (see branch testng-softassert-not-failing [HERE](https://github.com/jensenbrian/selenide/blob/dea03bece6e15dab419ee784d122fdbc8711ef34/modules/testng/src/test/java/integration/TestNgSoftTest.java#L64))
* **I think the test should continue as it does but mark the build as failed. Am I missing a basic concept or is this a bug?**

Running the same change made to `TestNgSampleTest.textArea()` in STRICT mode. The test does stop processing and marks the build as "BUILD FAILED". (This is not an issue. Working as expected.)

## Tell us about your environment
* **Selenide Version**: latest = 7.3.2
* **Chrome\Firefox\IE Version**: Chrome Version=125.0.6422.113 Platform=mac
* **Browser Driver Version**: 
* **Selenide-Appium Version**:
* **IOS\Android\WEB version**:
* **Selenium Version**: 4.21.0 build revision: 79ed462ef4
* **OS Version**: macOS Sonoma 14.5

## Code To Reproduce Issue
[modules/testng/src/test/java/integration/TestNgSoftTest.java](https://github.com/jensenbrian/selenide/blob/1dd8d005c217210b5aa080a80aa7766f45b7de6b/modules/testng/src/test/java/integration/TestNgSoftTest.java)

```java
  @Override
  @BeforeSuite
  final void setupAsserts() {
    Configuration.assertionMode = SOFT;
    SoftAsserts.fullStacktraces = false;
    closeWebDriver();
    WebDriverRunner.addListener(listener1);
  }

  @Test
  public void textAreaShouldFail() {
    $("#empty-text-area").val("text for textarea");
    // Will fail (on purpose) test will continue but should be marked as a FAILED test is in SOFT mode
    $("#empty-text-area").shouldNotHave(value("text for textarea"));
    $("#empty-text-area").shouldHave(value("text for textarea"));
  }
```

jensenbrian:
I found in the changelog:
6.6.0 (released 08.06.2022)
#1834 fix SoftAsserts with TestNG (downgraded TestNG from 7.5 to 7.4.0) -- see PR #1843

[...]

PR #2756, opened 2024-05-31
#2755 TestNG SoftAssert not failing (downgrade TestNG from 7.9.0 to 7.4.0)

jensenbrian:
## Proposed changes
In Soft Assert mode with a FAIL step is not marking the whole test as FAILED

See #2755 

## Checklist
-- Sort of... I added a test that fails now with TestNg 7.4.0. I tried to write a test that checks the status but that is the root of the problem. That does not work in a higher version of TestNg. The ITestResult is FAILED but the test shows as passed. I can remove the test since it is bad to have a test fail all the time. Any other ideas?

This is my first contribution. Please let me know if anything needs changing.
Thank you

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,5 @@
+.DS_Store
+
 build
 hs_err_pid*
 !.idea
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Changelog
 
+## 7.3.3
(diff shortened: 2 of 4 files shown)


jensenbrian:
Needs more work.
Looks like the "right" fix mentioned in #3046 is to use TestNG 7.10.0 and "Ordering listeners in TestNG" documented in [Specifying listeners with ServiceLoader](https://testng.org/#_specifying_listeners_with_serviceloader)

PR #2783, opened 2024-07-16
Add dom attribute and property conditions

rosolko:
## Proposed changes
Add dom attribute and property conditions.

Or just create a custom command for such cases and keep core lib as clean as possible. 

Reference https://github.com/selenide/selenide/discussions/2782

## Checklist


asolntsev (changes requested):
@rosolko Currently I don't fully understand in which cases these methods would be useful. 

At least some test showing difference in behaviour of `getAttribute`/`getDomAttribute`/`getDomProperty` would be useful.

rosolko:
That why I'm also propose custom commands for such cases

Issue #2800, opened 2024-08-05
Collection check for attributes doesn't see property values


asolntsev:
## 
Before [Selenide 6.8.0](https://ru.selenide.org/2022/09/24/selenide-6.8.0/), Selenide used Selenium method `WebElement.getAttribute(String)` that checks both _attributes_ and _properties_. 
Starting from 6.8.0, Selenide uses JS code to quickly get all attributes at once, but now it checks only _attributes_, not _properties_.

## Tell us about your environment
* **Selenide Version**: 7.4.1

## Code To Reproduce Issue
```java
    $$(".element").shouldHave(attributes("value", "#1", "#2", "#3"));
```

```html
<div id="one" class="element">One</div>
<div id="two" class="element">Two</div>
<div id="three" class="element">Three</div>

<script>
  document.getElementById('one').setAttribute('value', '11111')
  document.getElementById('two').value = '22222'
  document.getElementById('three').value = '33333'
</script>
```

Test failure:
```java
Attribute "value" values mismatch (#1 expected: "22222", actual: "null")
Actual (3): [11111, null, null]
Expected (3): [11111, 22222, 33333]
```

asolntsev:
fixed by https://github.com/selenide/selenide/pull/2801

PR #2801, opened 2024-08-05
#2800 Check dynamic properties in addition to attributes

asolntsev:


BorisOsipov (changes requested):
@asolntsev I've added the test that shows implementations problems. The original `getAttribute()` works well for this test.

In general, it will be nice to fully align the implementation with the original `getAttribute()` docs https://www.selenium.dev/selenium/docs/api/java/org/openqa/selenium/WebElement.html#getAttribute(java.lang.String)

At least for case  `The following are deemed to be "boolean" attributes, and will return either "true" or null:...`

BorisOsipov (approved):

Issue #2769, opened 2024-06-23
Need a method to scroll an element by a specified amount


donnieHub:
## The problem
Quite often there is a need to scroll down a particular element on a page (for example left menu on youtube page). This can be useful when content inside the element appears dynamically, i.e. only when you scroll down to it. 
Less often there is a need to scroll an element to the right.

## Details
This behavior can be easily achieved by using the javascript method `scrollBy()`. For example, if you want to scroll an element with id “box” down by 50 pixels, you can do:
```
const element = document.getElementById("box");
element.scrollBy(0, 50);
```
I have already implemented this method locally for selenide. I just need to get confirmation that this method is really necessary. It will look like this:
```
SelenideElement element = $("#box");
element.scrollOnElement(1000); // scroll element down by 1000 pixels
element.scrollOnElement(-1000); // scroll element up by 1000 pixels
element.scrollOnElement(0, 1000); // scroll element to the right by 1000 pixels
```

## Tell us about your environment
* **Selenide Version 7.3.1**: 
* **Chrome Version 126.0.6478.114**:

asolntsev:
Implemented in https://github.com/selenide/selenide/pull/2809

PR #2809, opened 2024-08-10
#2769 add method `$.scroll()` with direction and length parameters

donnieHub:
## Proposed changes
Quite often there is a need to scroll down a particular element on a page (for example left menu on youtube page). This can be useful when content inside the element appears dynamically, i.e. only when you scroll down to it.
Less often there is a need to scroll an element to the right.

## Checklist


otter1990:
Add somebody method `scrollToCenter` please :)

My entire project is riddled with `.scrollIntoView("{block: 'center'}")` methods

asolntsev:
> Someone add method `scrollToCenter()` please :)

@otter1990 Why don't you add this method? This is open source, your contribution is welcome! ;)

asolntsev (changes requested):
@donnieHub Thank you for the contribution!
Overall, it's a useful feature.

Few nuances:
1. Are you sure the name "Scroll on element" is clear for all people? Maybe it's better to call it like JavaScript function: `$.scrollTo()`? It's also not very clear, but at least such a function already exists in JavaScript, so it will be consistent. 
2. Instead of having two overloaded methods `scrollOnElement`, I suggest adding ONE method with parameter object `ScrollOptions`. Similar to `ClickOptions`, `DownloadOptions`, `TypeOptions`. It will let us easily add more parameters in future.

asolntsev:
> Someone add method `scrollToCenter()` please :)
> My entire project is riddled with `.scrollIntoView("{block: 'center'}")` methods

@otter1990 Why don't you use the old good method `$.scrollTo()`?

otter1990:
> @otter1990 Why don't you use the old good method `$.scrollTo()`?

I write tests for web mobile. scrollTo() method scrolls in such a way that the element ends up at the very top of the screen and is covered by the top menu of the site. If I use scrollIntoView(false), the element is hidden by the footer of the site. Therefore, I have to use scrollIntoView("{block: 'center'}")

asolntsev:
@otter1990 Thank you! 
I was thinking that probably we just need to change behaviour of `scrollTo()` method? Let it scroll the element to the center of screen instead of top.

otter1990:
> @otter1990 Thank you! I was thinking that probably we just need to change behaviour of `scrollTo()` method? Let it scroll the element to the center of screen instead of top.

May be. Personally, I wouldn't mind :)

asolntsev:
> Someone add method `scrollToCenter()` please :)

@otter1990 Method `scrollToCenter()` has been added in Selenide 7.6.0.
See https://github.com/selenide/selenide/pull/2842

asolntsev (changes requested):

asolntsev (approved):

PR #2835, opened 2024-09-05
Refactor selects

asolntsev:


src/main/resources/selects.js
```diff
@@ -0,0 +1,27 @@
...
+  return Array.from(select.options).find(option => option.value === requestedValue)
+}
+
+function applyChanges(select, changes) {
+  let previousSelectedOptions = getSelectedOptionsString(select)
+
+  select.focus()
+  changes(select)
+
+  select.dispatchEvent(new Event('click'));
```
BorisOsipov, line 21: It will be nice to restore event bubbling as it was [before](https://github.com/selenide/selenide/blob/v7.4.0/src/main/resources/select-options-by-value.js#L34)

asolntsev: @BorisOsipov Good point. Fixed.

PR #2879, opened 2024-10-23
Fix browser not created exception with AppiumPageFactory

PetroOv:
## Proposed changes
This is a potential solution for cases where a page initializes at the field level, but the driver hasn't been created yet. This issue occurs specifically when using the Appium plugin.
It's probably ok for Appium-only tests, but it is a problem if WEB and Mobile tests in the same repo

example code:
code https://pastebin.com/G4Y9gbmv
stacktrace: https://pastebin.com/JU3vLM2n

## Checklist

### DIFF
--- a/modules/appium/src/main/java/com/codeborne/selenide/appium/SelenideAppiumPageFactory.java
+++ b/modules/appium/src/main/java/com/codeborne/selenide/appium/SelenideAppiumPageFactory.java
@@ -16,12 +16,18 @@ import com.codeborne.selenide.impl.WebElementSource;
 import io.appium.java_client.HasBrowserCheck;
 import io.appium.java_client.pagefactory.AndroidFindAll;
 import io.appium.java_client.pagefactory.AndroidFindBy;
+import io.appium.java_client.pagefactory.AndroidFindByAllSet;
+import io.appium.java_client.pagefactory.AndroidFindByChainSet;
+import io.appium.java_client.pagefactory.AndroidFindBySet;
 import io.appium.java_client.pagefactory.AndroidFindBys;
 import io.appium.java_client.pagefactory.AppiumFieldDecorator;
 import io.appium.java_client.pagefactory.DefaultElementByBuilder;
 import io.appium.java_client.pagefactory.bys.builder.AppiumByBuilder;
 import io.appium.java_client.pagefactory.iOSXCUITFindAll;
 import io.appium.java_client.pagefactory.iOSXCUITFindBy;
+import io.appium.java_client.pagefactory.iOSXCUITFindByAllSet;
+import io.appium.java_client.pagefactory.iOSXCUITFindByChainSet;
+import io.appium.java_client.pagefactory.iOSXCUITFindBySet;
 import io.appium.java_client.pagefactory.iOSXCUITFindBys;
 import java.lang.annotation.Annotation;
 import java.util.List;
@@ -29,6 +35,7 @@ import org.openqa.selenium.By;
 import org.openqa.selenium.Capabilities;
 import org.openqa.selenium.HasCapabilities;
 import org.openqa.selenium.SearchContext;
+import org.openqa.selenium.WebDriverException;
 import org.openqa.selenium.WebElement;
 import org.openqa.selenium.support.ByIdOrName;
 import org.openqa.selenium.support.pagefactory.FieldDecorator;
@@ -47,6 +54,10 @@ import java.util.Optional;
 @ParametersAreNonnullByDefault
 public class SelenideAppiumPageFactory extends SelenidePageFactory {
   private static final Logger logger = LoggerFactory.getLogger(SelenideAppiumPageFactory.class);
+  private final List<Class<? extends Annotation>> platformAnnotations =
+    List.of(AndroidFindBy.class, AndroidFindBys.class, AndroidFindAll.class, AndroidFindByAllSet.class, AndroidFindByChainSet.class,
+      AndroidFindBySet.class, iOSXCUITFindBy.class, iOSXCUITFindBys.class, iOSXCUITFindAll.class, iOSXCUITFindByAllSet.class,
+      iOSXCUITFindByChainSet.class, iOSXCUITFindBySet.class);
 
   @Override
   @Nonnull
@@ -58,7 +69,6 @@ public class SelenideAppiumPageFactory extends SelenidePageFactory {
     return selector != null ? selector : super.findSelector(driver, field);
   }
 
-
   @Nonnull
   @CheckReturnValue
   private DefaultElementByBuilder byBuilder(Driver driver, Field field) {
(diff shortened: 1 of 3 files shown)


BorisOsipov:
>This is a potential solution

The code change is a Pyrrhic victory - it fixes one issue but create new. `DefaultElementByBuilder(platform, automationName)` needs to know the platform because it handles Android and iOS appium annotations @AndoridBy\@IosBlablaBy()\etc .If you pass nulls, it will always handle only general Selenium annotations like @FindBy, @FindAll, etc. isn't it?

P.S. Please avoid using static WebDriverRunner methods. You can obtain the WebDriver from the method parameter.

PetroOv:
> > This is a potential solution
> 
> The code change is a Pyrrhic victory - it fixes one issue but create new. `DefaultElementByBuilder(platform, automationName)` needs to know the platform because it handles Android and iOS appium annotations @AndoridBy@IosBlablaBy()\etc .If you pass nulls, it will always handle only general Selenium annotations like @findby, @FINDALL, etc. isn't it?
> 
> P.S. Please avoid using static WebDriverRunner methods. You can obtain the WebDriver from the method parameter.

You're right, this "fix" could introduce a new problem that is currently handled by throwing an exception. The deeper issue is that we can't use new selectors without creating a driver, as the AppiumFieldDecorator requires casting to the driver.

Regarding the static WebDriverRunner, I attempted to replace it, but I couldn't find a method to check whether the driver has started the browser. Are you suggesting that I should use a try-catch  on getWebdriver block to handle this exception?

BorisOsipov:
>The deeper issue is that we can't use new selectors without creating a driver, as the AppiumFieldDecorator requires casting to the driver.

Yes, I see the problem. But for now, it seems impossible to fix - we need to know the platform, and we can only get it from the capabilities. We don't have access to the user capabilities\options instance, and the only way to get them is from the driver.

Even if we create Android/iOSDriverFactories and allow users to pass capabilities via `config.browserCapabilities` and we will be able to get platform from it, we still can't cover cases like the current implementation through the WebDriverProvider.
It's mostly a design trade-off with Appium decorators, allowing the use of one field with different platform decorators.

PetroOv:
> > The deeper issue is that we can't use new selectors without creating a driver, as the AppiumFieldDecorator requires casting to the driver.
> 
> Yes, I see the problem. But for now, it seems impossible to fix - we need to know the platform, and we can only get it from the capabilities. We don't have access to the user capabilities\options instance, and the only way to get them is from the driver.
> 
> Even if we create Android/iOSDriverFactories and allow users to pass capabilities via `config.browserCapabilities` enabling us to get the platform from it, we still can't cover cases like the current implementation through the WebDriverProvider. It's mostly a design trade-off with Appium decorators, allowing the use of one field with different platform decorators.

Yeah, without a real driver, it's hard to determine its type clearly. But what if we check not only the device but also the locator? What I mean is, if the selector is a special one we try to define platform, otherwise use the default factory.
example
https://pastebin.com/RHdS2nJA

BorisOsipov:
>What I mean is, if the selector is a special one we try to define platform, otherwise use the default factor

seems OK

PetroOv:
> > What I mean is, if the selector is a special one we try to define platform, otherwise use the default factor
> 
> seems OK

I have updated pr with discussed changes

modules/appium/src/main/java/com/codeborne/selenide/appium/SelenideAppiumPageFactory.java
```diff
@@ -136,4 +154,11 @@ private static class SelenideAppiumList extends SelenideAppiumCollection impleme
       super(collection);
     }
   }
+
+  private boolean isPlatformAnnotationAdded(Field field) {
+    List<Class<? extends Annotation>> classes =
+      List.of(AndroidFindBy.class, AndroidFindBys.class, AndroidFindAll.class, iOSXCUITFindBy.class, iOSXCUITFindBys.class,
```
BorisOsipov, line 160: Add all annotaitons please. + extract list to class field.
```
AndroidFindAll
AndroidFindBy
AndroidFindByAllSet
AndroidFindByChainSet
AndroidFindBys
AndroidFindBySet
iOSXCUITFindAll
iOSXCUITFindBy
iOSXCUITFindByAllSet
iOSXCUITFindByChainSet
iOSXCUITFindBys
iOSXCUITFindBySet
```

modules/appium/src/main/java/com/codeborne/selenide/appium/SelenideAppiumPageFactory.java
```diff
@@ -44,15 +52,25 @@ public class SelenideAppiumPageFactory extends SelenidePageFactory {
...
   @Nonnull
   @CheckReturnValue
-  private DefaultElementByBuilder byBuilder(Driver driver) {
+  private DefaultElementByBuilder byBuilder(Driver driver, Field field) {
+    if (!isPlatformAnnotationAdded(field)) {
+      return new DefaultElementByBuilder(null, null);
+    }
+
+    if (!driver.hasWebDriverStarted()) {
+      throw new RuntimeException("The Appium Page factory requires a browser instance to be created before calling" +
```
BorisOsipov, line 70: I don't think we really need it. Either way, users will get  `No webdriver is bound to current thread: 1. You need to call open(url) first.`
And wording `The Appium Page factory requires a browser instance` may confuse. Which browser do we ask to launch when we test a mobile application?

BorisOsipov:
It would be nice to have tests to avoid regressions in the future.

modules/appium/src/main/java/com/codeborne/selenide/appium/SelenideAppiumPageFactory.java
PetroOv: I thought it would be helpful to handle such cases better by adding additional context to aid in debugging. Initially, it wasn’t clear why I received the exception, and I wasn’t sure if it was expected.

I've updated the exception message. Please review it, and if my changes don't align with your perspective, let me know, and I’ll remove this part.

BorisOsipov (approved):

asolntsev:
@PetroOv @BorisOsipov The new test `SelenideAppiumPageFactoryTest` is broken.
This test cannot work because `page.element` is null:

```java
  @Test
  void web_element_page_factory_doesnt_require_webdriver_instance() {
    var page = new PageWithWebSelectors();
    assertThat(page.element).isNotNull()
  }
```

The build is green because `SelenideAppiumPageFactoryTest` is not executed. GitHub Actions executes only tests in packages `it.mobile.ios` and `it.mobile.android`. :)

PetroOv:
> @PetroOv @BorisOsipov The new test `SelenideAppiumPageFactoryTest` is broken. This test cannot work because `page.element` is null.
> 
> ```java
>   @Test
>   void web_element_page_factory_doesnt_require_webdriver_instance() {
>     var page = new PageWithWebSelectors();
>     assertThat(page.element).isNotNull()
>   }
> ```
> 
> Page object is created with just `new`, without any factories. That's why all its fields are `null`.
> 
> The build is green because `SelenideAppiumPageFactoryTest` is not executed. GitHub Actions executes only tests in packages `it.mobile.ios` and `it.mobile.android`. :)

Sorry, my bad. I haven't checked tests after linter fix. I have fixed tests and also moved test class to `com/codeborne/selenide/appium` package is that ok?

PetroOv:
@asolntsev Could I ask you to review my latest changes? The tests should be working now.

asolntsev (approved):

asolntsev:
@PetroOv Thank you for the contribution. It has been released in Selenide 7.6.0.

PR #2858, opened 2024-09-26
fix regExp for spaces separators

DaniilMoiseevPenza:
## Proposed changes
Some space separator don`t deletes from Strings in texts assertions

## Checklist
- [x] Checkstyle and unit tests are passed locally with my changes by running `gradlew check chrome_headless firefox_headless` command
- [x] I have added tests that prove my fix is effective or that my feature works
- [ ] I have added necessary documentation (if appropriate)

PR #2884, opened 2024-11-01
Support more spaces

asolntsev:
Continuation of #2858

1. support special space characters when finding element by text
2. add few more space characters


src/main/java/com/codeborne/selenide/impl/Html.java
```diff
@@ -8,7 +8,9 @@
 
 @ParametersAreNonnullByDefault
 public class Html {
-  private static final Pattern REGEX_SPACES = Pattern.compile("[\\s\\p{Zs}]+");
+  private static final Pattern REGEX_SPACES = Pattern.compile(
+    "[\\s\\p{Zs}\u1680\u2000\u2001\u2002\u2003\u2004\u2005\u2006\u2007\u2008\u2009\u200A\u200B\u200C\u200D\u202F\u205F\u2060\u3000]+");
```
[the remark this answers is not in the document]
asolntsev: Good point.
* I removed most of these characters from `Html.java` (except `‌‍⁠‌‍⁠‌‍⁠‌u200C`, `200D` and `u2060`).
* But I still cannot remove them from `texts.js` and `Xpath.java` :(

src/main/resources/find-elements-by-text-case-insensitive.js
```diff
@@ -1,18 +1,9 @@
 (function () {
+  import 'texts.js'
```
BorisOsipov, line 2: I'd really prefer not to use the word `import` here to avoid any ambiguity with `ESM modules in JS`. Any other word that conveys we're adding text from another file would be better, to make it clear that this is not an ESM import at all.

embed\appead sounds good.

asolntsev: What I like about `import` is that IDEA shows it without highlighting syntax errors, and allows to navigate into "texts.js" by clicking it.

image

If I replace `import` by `embed`, IDEA highlights it as an error, and doesn't allow navigation by click:

image

BorisOsipov (approved):

PR #2889, opened 2024-11-09
JSR305 -> JSpecify

asolntsev:
With the help of "package-info.java" files, now
* all Selenide parameters and return values are not-nullable by default;
* all Selenide methods are annotated by @CheckReturnValue by default.

Also, I introduced `FluentCommand` class for all commands returning the same `SelenideElement`. This was an easy way to make their `execute` method non-nullable.


modules/appium/src/test/java/it/mobile/android/AndroidScreenshotTest.java
```diff
@@ -54,6 +54,7 @@ void elementScreenshot() {
   @Test
   void elementScreenshotAsImage() throws IOException {
     BufferedImage screenshot = $(By.xpath(".//*[@text='Views']")).screenshotAsImage();
+    assertThat(screenshot).isNotNull();
```
BorisOsipov, line 57: There is the same assertion below

BorisOsipov:
It will be nice to wrap [this](https://github.com/selenide/selenide/blob/c9398c0f234bb845e9d511e5e28ec8855e5fc920/src/main/java/com/codeborne/selenide/impl/Describe.java#L53) with requireNonNull() too. `new TreeMap<>(null);` throws NPE.

modules/appium/src/test/java/it/mobile/android/AndroidScreenshotTest.java
asolntsev: fixed

asolntsev:
> It will also be nice to wrap [this](https://github.com/selenide/selenide/blob/c9398c0f234bb845e9d511e5e28ec8855e5fc920/src/main/java/com/codeborne/selenide/impl/Describe.java#L53) with requireNonNull() too. `new TreeMap<>(null);` throws NPE.

fixed

rosolko (approved):

BorisOsipov (commented):
It's hard to review this because there's a lot of change in a single commit. There are two main types of changes: annotation updates and some improvements to nullability. I took a quick look and shared some advice, but a full review would take more time than I have right now.
From what I can see, I’m happy to remove the unnecessary annotations.

PR #2945, opened 2025-01-28
Added ability to use custom types for elements in appium page factory

qwez:
## Proposed changes
Ability to use custom element types (extended from `SelenideAppiumElement`) with page factory. Usage example in the readme

## Checklist

### DIFF
--- a/modules/appium/src/main/java/com/codeborne/selenide/appium/SelenideAppiumPageFactory.java
+++ b/modules/appium/src/main/java/com/codeborne/selenide/appium/SelenideAppiumPageFactory.java
@@ -135,7 +135,7 @@ public class SelenideAppiumPageFactory extends SelenidePageFactory {
                                                Field field, @Nullable String alias) {
     return shouldCache(field) ?
       LazyWebElementSnapshot.wrap(new ElementFinder(driver, searchContext, selector, 0, alias)) :
-      ElementFinder.wrap(driver, SelenideAppiumElement.class, searchContext, selector, 0, alias);
+      ElementFinder.wrap(driver, getTargetType(field), searchContext, selector, 0, alias);
   }
 
   @Override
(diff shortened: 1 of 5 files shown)


modules/appium/src/main/java/com/codeborne/selenide/appium/SelenideAppiumPageFactory.java
```diff
@@ -156,4 +156,10 @@ private static class SelenideAppiumList extends SelenideAppiumCollection impleme
   private boolean isPlatformAnnotationAdded(Field field) {
     return platformAnnotations.stream().anyMatch(field::isAnnotationPresent);
   }
+
+  @SuppressWarnings("unchecked")
+  private <T extends SelenideAppiumElement> Class<T> getTargetType(Field field) {
+    return SelenideAppiumElement.class.isAssignableFrom(field.getType()) ?
+      (Class<T>) field.getType() : (Class<T>) SelenideAppiumElement.class;
+  }
```
asolntsev, line 164: Maybe we should throw an exception if field is not a `SelenideAppiumElement` subclass?

asolntsev (changes requested):
@qwez Thank you, a nice feature!
Why only Appium? This feature would be useful in web tests as well.
(I can implement it by myself if you are busy).

modules/appium/src/main/java/com/codeborne/selenide/appium/SelenideAppiumPageFactory.java
qwez: There wasn't such check before, we just got `IllegalArgumentException` when try to `Field.set()`. But, sure, I think it will be useful.

qwez:
@asolntsev `run-selenoid-tests` task failed because of selenoid failed to start. Is it blocker?

asolntsev:
> @asolntsev `run-selenoid-tests` task failed because of selenoid failed to start. Is it blocker?

Don't worry, Selenoid became unstable. And it's [not maintained anymore](https://selenide.org/2025/01/07/selenide-7.7.0/#moon).

asolntsev (approved):
@qwez  Great job!

PR #2964, opened 2025-02-28
Extend appium interacts with apps abilities

rosolko:
## Proposed changes

Added abilities:
- activate app
- terminate app with timeout
- relaunch app (terminate + activate)
- relaunch app with timeout

## Checklist

### DIFF
--- a/modules/appium/src/main/java/com/codeborne/selenide/appium/AppiumNavigator.java
+++ b/modules/appium/src/main/java/com/codeborne/selenide/appium/AppiumNavigator.java
@@ -4,7 +4,12 @@ import com.codeborne.selenide.Configuration;
 import com.codeborne.selenide.logevents.SelenideLogger;
 import io.appium.java_client.AppiumDriver;
 import io.appium.java_client.InteractsWithApps;
+import io.appium.java_client.appmanagement.BaseTerminateApplicationOptions;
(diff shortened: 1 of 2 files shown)


asolntsev (approved):
@rosolko LGTM
1. Is it good to mark timeout as `@Nullable`? Maybe it's better to always set some timeout?
2. Don't we need to add timeout parameter for method `activateApp` as well?

rosolko:
1. By default timeout paramenter for command has 500ms value, so I suppose we only need to pass it when changing it.
2. activateApp command has no timeout parameter, only appId/bundleId available. To be specific, only android support timeout for terminateApp command, but in our care implementation is more generic one, and extra parameter for ios command does nothing.

android references:

1. [activateapp](https://github.com/appium/appium-uiautomator2-driver?tab=readme-ov-file#mobile-activateapp)
2. [terminateapp](https://github.com/appium/appium-uiautomator2-driver?tab=readme-ov-file#mobile-terminateapp)

ios references:

1. [activateapp](https://github.com/appium/appium-xcuitest-driver/blob/master/docs/reference/execute-methods.md#mobile-activateapp)
2. [terminateapp](https://github.com/appium/appium-xcuitest-driver/blob/master/docs/reference/execute-methods.md#mobile-terminateapp)

PR #2968, opened 2025-03-04
Add possibility to work with Shadow DOM using selenide page factory

groov1kk:
## Proposed changes

Added new annotation `@ShadowHost` to be able working with elements inside the shadow roots using page factory.
 For example:

```java
public class PageObject {

    @ShadowHost(@FindBy(id = "shadow-host"))
    @FindBy(css = "#shadowId")
    private SelenideElement shadowElement;

    @ShadowHost({@FindBy(id = "shadow-host"), @FindBy(css = "#inner-shadow-host")})
    @FindBy(css = "#innerShadowId")
    private SelenideElement innerShadowElement;

}
```
Where the annotation`@ShadowHost({shadow-host-locator, inner-shadow-host-locator, ... other-hosts-locators})` provides to specify a shadow path to a target element, and this element would be found using the shadow root of the last specified locator.

There are a couple of specific things:
1. This should work both for elements and containers.
2. Unfortunately, only the first locator of this path could use any of the given Selenium strategies to find a shadow host. Both all the rest inner shadow host locators and the target element must use only _CSS locators_. This is webdriver restriction (for example, see [here](https://github.com/w3c/webdriver/issues/1610)) and I hope it will be fixed in a future.
3. To find a root element I used JS instead of [native approach](https://w3c.github.io/webdriver/#get-element-shadow-root) just to wide the range of compatible browsers.

One another interesting case with Shadow DOM is when we'd like to specify a container of elements where _self_ element of this container is divided from its descendants by shadow root. For example, when we would like to put this _div_ element into container:

Screenshot 2025-03-04 at 4 28 53 pm

We should specify that the _self_ element itself is a shadow host. It is possible to do it using the `@ShadowRoot` annotation. For example:

```java
public class PageObject {

      @ShadowRoot
      @FindBy(id = "shadow-container")
      private MyAwsomeContainer container;

}
```
or

```java

@ShadowRoot
public class MyAwsomeContainer implements Container {

      @FindBy(css = ".shadow-container-child")
      private List<SelenideElement> children;
}

```
And all descendants would be found using shadow root of the _self_ element.

## Checklist


asolntsev (approved):
LGTM!

PR #3018, opened 2025-04-25
fix(plugins): Fix SpotBugs config after upgrade to 6.0.2

ylazakovich:
## Proposed changes

Here the [comment](https://github.com/spotbugs/spotbugs-gradle-plugin/issues/972#issuecomment-1769247750) which describes solution for fixing this problem with property `effort`

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,13 @@
 # Changelog
 
+## 7.9.0 (19.04.2025)
+* #2763 Can click with JS holding keys ALT, CTRL etc. (#3008)
+* #2992 Write screenshots in a format that Jenkins and GitLab treat as attachments (#2998)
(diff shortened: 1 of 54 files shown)


asolntsev (approved):
I have no idea what is happening here, but if it works - let it be.

@BorisOsipov @rosolko What do you think?

BorisOsipov (commented):
Would be nice to leave a comment in the code or commit message to "explain" what’s going on (and maybe drop a link to the GitHub issue), because in a year we’ll forget why we did it this way

ylazakovich:
> Would be nice to leave a comment in the code or commit message to "explain" what’s going on (and maybe drop a link to the GitHub issue), because in a year we’ll forget why we did it this way

In description you can find the comment which explain this solution

I also was curious that enum value or valueOf("MAX") doesn't work in gradle configuration
Only this solution works for defining effort = MAX

I have added comment to code

PR #3053, opened 2025-06-19
Add static methods for Selenide Containers initialization

groov1kk:
## Proposed changes

Added the following methods to be able to initialize custom Selenide Container:

```
static ContainerClass $(String, Class<ContainerClass>)
static ContainerClass $(By, Class<ContainerClass>)
static ContainerClass $(WebElement, Class<ContainerClass>)
static ContainerClass $x(String, Class<ContainerClass>)

// Their aliases
static ContainerClass container(String, Class<ContainerClass>)
static ContainerClass container(By, Class<ContainerClass>)
static ContainerClass container(WebElement, Class<ContainerClass>)
```
And methods to initialize a list of Containers:
```
static List<ContainerClass> $$(String, Class<ContainerClass>)
static List<ContainerClass> $$(By, Class<ContainerClass>)
static List<ContainerClass> $$(Collection<WebElement>, Class<ContainerClass>)
static List<ContainerClass> $$x(String, Class<ContainerClass>)

// Their aliases
static List<ContainerClass> containers(String, Class<ContainerClass>)
static List<ContainerClass> containers(By, Class<ContainerClass>)
static List<ContainerClass> containers(Collection<WebElement>, Class<ContainerClass>)
```

## Checklist


asolntsev:
@groov1kk Thank you for the PR! Generally, it looks good. 
But can you show some realistic example where it might be useful?

Like, if I have `<input type="button" id="login" value="Login">`, then usually I write `$("#login").click();`.
What will change now?

I will need to write a container class like
```java
  class Button {
    @Self
    SelenideElement button;
  }

  @Test void myTest() {
     Button login = $("#login", Button.class);
     login.button.click();
  }
```

Why it's better? What benefits it gives? 
I personally don't like overuse of `@Self` and `self`.

groov1kk:
> @groov1kk Thank you for the PR! Generally, it looks good. But can you show some realistic example where it might be useful?
> 
> Like, if I have `<input type="button" id="login" value="Login">`, then usually I write `$("#login").click();`. What will change now?
> 
> I will need to write a container class like
> 
> ```java
>   class Button {
>     @Self
>     SelenideElement button;
>   }
> 
>   @Test void myTest() {
>      Button login = $("#login", Button.class);
>      login.button.click();
>   }
> ```
> 
> Why it's better? What benefits it gives? I personally don't like overuse of `@Self` and `self`.

Hello!

There is a fresh example of a rather silly situation when I had, for example, two popups with "Next" and "Back" buttons where pressing on these buttons I should open-another/return-to-the-previous popup. Using the given Selenide API I could write something like...:

```java
class PopupA implements Container {

  private PopupB popupB;

  public PopupB next() {
    // click next
    return popupB;
  }
}

class PopupB implements Container {

 private PopupA popupA;

 public PopupA back() {
   // click back
   return popupA;
 }
}
```
...and receive ```StackOverflowException``` on page object initialization because these containers contain links on each other.

Writing something like:

```java
class MyPage {
  
 private PopupA popupA;
 private PopupB popupB;

 public PopupB clickNextInPopupA() {
   popupA.next();
   return popupB;
 }
}
```
also was not an option because we had a single page application and this page contained quite a lot of different popups and our page object was overloaded with such methods, part of which can be invoked only when correct popup is open.

Having possibility of initializing containers on demand I could re-write this code to avoid this problem:

```java
class PopupA implements Container {

  public PopupB next() {
    // click next
    return container("#popupb", PopupB.class);
  }
}

class PopupB implements Container {

 public PopupA back() {
    // click back
    return container("#popupa", PopupA.class);
  }
}
```
I know that this is an exceptional case and of course this feature is not for simple elements like buttons. But couple of times I was pity that Selenide does not support this feature.

asolntsev:
And how is it different from this?

```java
class PopupA {
  private final SelenideElement a = $("#popupa");

  public PopupB next() {
    return page(PopupB.class);
  }
}

class PopupB {
  private final SelenideElement b = $("#popupb");

 public PopupA back() {
    return page(PopupA.class);
  }
}
```

groov1kk:
From HTTP perspective, popup is not a page, it is exactly a part of the page.

Making popups as pages we also need to extend all its locators by the value of popup's locator itself:

```java
class PopupA {

  private final SelenideElement a = $("#popupa a");
  private final SelenideElement b = $("#popupa b");
  private final SelenideElement c = $("#popupa c");

}
```

asolntsev:
> From HTTP perspective, popup is not a page, it is exactly a part of the page.


Yes, it's totally fine. From a well-known post by Martin Fowler:
> A page object wraps an HTML page, or fragment


> we also need to extend all its locators by the value of popup's locator itself

It's easily solvable:

 ```java
class PopupA {

  @FindBy(css="a")
  private SelenideElement a;

  @FindBy(css="b")
  private SelenideElement b;

}
```

**Comment by groov1kk** [comment]
> > we also need to extend all its locators by the value of popup's locator itself
> 
> It's easily solvable:
> 
> ```java
> class PopupA {
> 
>  @FindBy(css="a")
>  private SelenideElement a;
> 
>  @FindBy(css="b")
>  private SelenideElement b;
> 
> }
> ```

No, I mean the situtation when a page, for example, contains several popups, for example with locators #popupa and #popupb. And let's say, I need to find an element with locator ```@FindBy(css="b")``` in the popup with locator #popupb, which is the second from the beggining of the page.
 
The problem is that both popups may contain elements with the same locators and, as popups are the parts of the same page, this locator may return something from the another, a currently invisible popup. To avoid this I have to include the popup's locator itself there ```@FindBy(css="#popupb b")``` and for the every element in all popups. In other words, I have to perform contextual search for every element in every popup.

As I understend, Containers are responsible for blocks of elements with contextual search in Selenide and the problem is that we cannot initialize these containers except that making them as part of the pages.

**Review by asolntsev — APPROVED** [review]

**Comment by asolntsev** [comment]
Thanks, now I understand better your need.

I am not against it, just trying to understand it deeper and discuss alternatives...

I see that your problem can be solved without any components at all, with just old good Java classes.
For example, like this:

```java
class PopupA {
  private final String root;

  public PopupB next() {
     return new PopupB("#popupb");
  }
}

class PopupB {
  private final String root;

  public PopupA back() {
     return new PopupA("#popupa");
  }
  public void login() {
    $(root + " #login").click();
  }
  public SelenideElement title() {
    return $(root + " #title");
  }
}
```

Issue #3058, opened 2025-06-28
Video recorder Bugs with colors, long time, gif


starygin:
selenide-video-recorder 7.9.3
1) When running tests, the screens are in normal colors, after gluing the colors into the video, there is a defect with the colors.
2) Video splicing takes a very long time
3) Maybe it's worth gluing it into a GIF?

![Image](https://github.com/user-attachments/assets/1f05ed40-6273-4154-b947-74bb63d8ca11)

![Image](https://github.com/user-attachments/assets/66451a69-7a7d-4e7a-9fec-0b36f48335ac)

asolntsev:
1. How exactly did you create the screenshots? By Selenium? In which browser?
Probably it's the same issue as #3042 and it was already fixed in #3043 (will be released as Selenide 7.9.4).
2. I tried to replace "Bytedeco" library by just CMD call `ffmpeg -i ...` - and it works much faster. Will rework it.
3. Maybe even better to generate MP4 files. Seems to work fast and supported by all browsers.

starygin:
1) Screenshots were taken by a video recorder. Selenide 7.9.3 / Selenium 4.33.0 / Sberbrowser (chromium under the hood)
2,3  Very goods ideas. wait release 7.9.4

PR #3077, opened 2025-07-27
#3058 take screenshots for video recorder using BiDi or DevTools method

asolntsev:
This is good for 2 reasons:
1. these methods are faster than the "standard" method `TakesScreenshot.getScreenshotAs`
2. browser doesn't take over the focus (which may break the test, especially in multithreaded test run).

1. in Chromium browsers, DevTools is the fastest method (and BiDi is slow)
2. in Firefox, BiDi is the fastest method.

See my experiments with different screenshooting methods: https://github.com/asolntsev/selenium-labs/blob/main/src/test/java/labs/screenshots/ScreenshotPerformanceTest.java

### DIFF
--- a/modules/video-recorder/src/main/java/org/selenide/videorecorder/core/ScreenShooter.java
+++ b/modules/video-recorder/src/main/java/org/selenide/videorecorder/core/ScreenShooter.java
@@ -1,8 +1,14 @@
 package org.selenide.videorecorder.core;
 
 import com.codeborne.selenide.drivercommands.WebdriversRegistry;
+import com.codeborne.selenide.impl.WebDriverInstance;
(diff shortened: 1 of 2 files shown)


vlad8x8:
@asolntsev why don't you use built-in https://chromedevtools.github.io/devtools-protocol/tot/Page/#method-startScreencast ?

asolntsev:
@vlad8x8 I don't know. I've never tried it. Maybe it's a good idea.
But anyway, it words only in Chrome, right? So we have to implement video recorder for other browsers anyway.

rosolko:
As well, this feature is marked as 'Experimental' as of now.

PR #3078, opened 2025-07-27
add method $.scrollIntoView(options)

asolntsev:
`ScrollIntoViewOptions` has benefits over boolean or String parameter:
1. A convenient factory method for constructing `ScrollIntoViewOptions` is  `instant()` which performs scrolling instantly, thus avoiding flaky tests.
2. compile-safe (compared to String parameter)

Also,
* deprecated method `$.scrollIntoView(boolean)`
* changed behaviour of these methods - now they perform the scrolling **instantly** (ignoring CSS property "scroll-behavior: smooth").
  1. `$.scrollTo()`
  2. `$.scrollIntoCenter()`
  3. `$.scroll(ScrollOptions)`

### DIFF
--- a/src/main/java/com/codeborne/selenide/ScrollIntoViewOptions.java
+++ b/src/main/java/com/codeborne/selenide/ScrollIntoViewOptions.java
@@ -1,0 +1,55 @@
+package com.codeborne.selenide;
+
+import static com.codeborne.selenide.ScrollIntoViewOptions.Behavior.instant;
+import static com.codeborne.selenide.ScrollIntoViewOptions.Block.start;
+import static com.codeborne.selenide.ScrollIntoViewOptions.Inline.nearest;
+
+/**
+ * See <a href="https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoView">scrollIntoView documentation</a>
+ *
+ * @param behavior whether scrolling is instant or animates smoothly
+ * @param block the vertical alignment of the element within the scrollable ancestor container
+ * @param inline the horizontal alignment of the element within the scrollable ancestor container
+ */
+public record ScrollIntoViewOptions(
+  Behavior behavior,
+  Block block,
+  Inline inline
+) {
+
+  public static ScrollIntoViewOptions instant() {
+    return new ScrollIntoViewOptions(instant, start, nearest);
+  }
+
+  public ScrollIntoViewOptions block(Block block) {
+    return new ScrollIntoViewOptions(behavior, block, inline);
+  }
+
+  public ScrollIntoViewOptions inline(Inline inline) {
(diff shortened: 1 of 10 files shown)


src/main/java/com/codeborne/selenide/ScrollIntoViewOptions.java
```diff
@@ -0,0 +1,55 @@
...
+ * @param block the vertical alignment of the element within the scrollable ancestor container
+ * @param inline the horizontal alignment of the element within the scrollable ancestor container
+ */
+public record ScrollIntoViewOptions(
+  Behavior behavior,
+  Block block,
+  Inline inline
+) {
+
+  public static ScrollIntoViewOptions instant() {
```
rosolko, line 20: Maybe it's better to call it `default` and fullfill with default values based on documentation: 

- behavior: null (or auto)
- block: start
- inline: nearest

Give an ability to set `behavior` also. And then override values for `scrollTo` and `scrollIntoView`methods.

asolntsev: Usually, it would be reasonable.
But I wanted to force using "instant" because it avoids risk of flaky tests.

The default value "auto" can scroll smoothly (slowly) which often causes random test failures.

src/main/java/com/codeborne/selenide/SelenideElement.java
```diff
@@ -1109,9 +1115,23 @@ public interface SelenideElement extends WebElement, WrapsDriver, WrapsElement,
   @CanIgnoreReturnValue
   SelenideElement scrollIntoView(String scrollIntoViewOptions);
```
rosolko, line 1116: If we introduce typed method let's deprecate or even remove this one (which accept `String`)

src/main/java/com/codeborne/selenide/ScrollIntoViewOptions.java
asolntsev: Maybe.
But it's still possible to set behaviour in constructor: `new ScrollIntoViewOptions(auto, block, inline)`.

src/main/java/com/codeborne/selenide/SelenideElement.java
asolntsev: done.

Soumendu-Khilar:
@asolntsev If still there is a need to use smooth scrolling behavior with alingnToTop, is there any way to achieve it?

asolntsev:
> @asolntsev If still there is a need to use smooth scrolling behavior with alingnToTop, is there any way to achieve it?

Sure, you just need to use constructor: `new ScrollIntoViewOptions(Behavior.smooth, Block.center, Inline.center)`

PR #3095, opened 2025-08-25
Fix gradle scripts

asolntsev:


rosolko:
Related to this bug https://github.com/gradle/gradle/issues/19815 and related pull request https://github.com/gradle/gradle/pull/33548 it can be cause of such behavior.

asolntsev:
overseded by https://github.com/selenide/selenide/pull/3313

PR #3126, opened 2025-10-01
Potential fix for code scanning alert no. 17: Uncontrolled data used in path expression

rosolko:
https://github.com/selenide/selenide/security/code-scanning/17


asolntsev (approved):
It was not a problem because it's used only in Selenide own tests.

But now it's better, thanks. :)

PR #3130, opened 2025-10-02
Avoid `static` if not really necessary

mkurz:
I think it's not necessary to make this members static. Better for encapsulation to make them non-static.


mkurz:
Ready to merge, thanks!

rosolko (approved):

PR #3134, opened 2025-10-02
Allow to pass custom `ScreenShotLaboratory` to `SelenideDriver`

mkurz:
I want to be able to pass my own `ScreenShotLaboratory` when creating a `SelenideDriver` instance.
Right now it's hardcoded to `inject()`..

### DIFF
--- a/src/main/java/com/codeborne/selenide/SelenideDriver.java
+++ b/src/main/java/com/codeborne/selenide/SelenideDriver.java
@@ -42,10 +42,10 @@ import static java.util.Objects.requireNonNull;
 public class SelenideDriver {
   private final JavaScript zoomJs = new JavaScript("zoom.js");
   private final Navigator navigator = new Navigator();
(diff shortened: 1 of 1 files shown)


asolntsev:
@mkurz That's the point of `inject`. You can provide your own implementation by adding file

`src/main/resources/META-INF/services/com.codeborne.selenide.impl.ScreenShotLaboratory` with full class name of your implementation. Not needed to pass it to constructor.

mkurz:
@asolntsev I know that, but I prefer to do it by code.

mkurz:
@asolntsev is there a gradle task to format everything correctly? Thanks!

asolntsev:
>  is there a gradle task to format everything correctly? 

In fact, I don't know :)
Usually, I just write code in IDEA (which understands file "editorconfig" in the project) and run `./gradlew check`.

PR #3135, opened 2025-10-02
Allow subclassing of `ScreenShotLaboratory`

mkurz:
Together with 
- #3134

I can finally create my own custom `SelenideDriver`:

```java
    final ScreenShotLaboratory screenShotLaboratory = new NoThreadLocalScreenShotLaboratory(
            new WebdriverPhotographer(),
            new WebPageSourceExtractor(),
            new AttachmentPrinter(),
            new Clock()
    );
    final SelenideConfig config = new SelenideConfig();
    final SelenideDriver driver = new SelenideDriver(config, someWebDriver, null, screenShotLaboratory);
```

Main reason is I do not want to use the static `Selenide` class, where everything is done in a static way.
I want to have `SelenideDriver` instances which act independent of each other.

Further I want to avoid all the `ThreadLocal`s in your default `ScreenShotLaboratory` by subclassing it and override methods that access and set a `ThreadLocal` like `generateScreenshotFileName`.

Basically I want to have total encapsulated instances.

```java
import com.codeborne.selenide.impl.AttachmentHandler;
import com.codeborne.selenide.impl.Clock;
import com.codeborne.selenide.impl.PageSourceExtractor;
import com.codeborne.selenide.impl.Photographer;
import com.codeborne.selenide.impl.ScreenShotLaboratory;
import com.codeborne.selenide.impl.Screenshot;
import org.jspecify.annotations.Nullable;

import java.io.File;
import java.util.List;
import java.util.Optional;

public class NoThreadLocalScreenShotLaboratory extends ScreenShotLaboratory {

    private NoThreadLocalScreenShotLaboratory() {
    }

    protected NoThreadLocalScreenShotLaboratory(Photographer photographer, PageSourceExtractor extractor, AttachmentHandler attachmentHandler, Clock clock) {
      super(photographer, extractor, attachmentHandler, clock);
    }


    @Override
    protected String generateScreenshotFileName() {
        return clock.timestamp() + "." + screenshotCounter.getAndIncrement();
    }

    @Override
    protected void addToHistory(Screenshot screenshot) {
        synchronized (allScreenshots) {
            allScreenshots.add(screenshot);
        }
    }

    @Override
    public void startContext(String className, String methodName) {
        // no-op
    }

    @Override
    public void startContext(String context) {
        // no-op
    }

    @Override
    public List<Screenshot> finishContext() {
        // no-op
        return screenshots();
    }

    @Override
    public List<@Nullable File> getThreadScreenshots() {
        return getScreenshots();
    }

    @Override
    public List<Screenshot> threadScreenshots() {
        return screenshots();
    }

    @Override
    public Optional<File> getLastThreadScreenshot() {
        return Optional.of(getLastScreenshot());
    }

    @Override
    public Optional<Screenshot> lastThreadScreenshot() {
        return Optional.of(lastScreenshot());
    }

    @Override
    public List<@Nullable File> getContextScreenshots() {
        return getScreenshots();
    }

    @Override
    public List<Screenshot> contextScreenshots() {
        return screenshots();
    }

    @Override
    public Optional<File> getLastContextScreenshot() {
        return Optional.of(getLastScreenshot());
    }

    @Override
    public Optional<Screenshot> lastContextScreenshot() {
        return Optional.of(lastScreenshot());
    }
}
```


mkurz:
The whole reason I am doing this is because I want to replace [FluentLenium](https://github.com/FluentLenium/FluentLenium) (which is archived and not maintained anymore since a while) with Selenide in the [Play Framework](https://github.com/playframework/playframework/) (but not version 1 which you have been working on).

asolntsev:
@mkurz The idea is generally good, for sure. But you are trying to make too many things public... 
Maybe a better option would be adding `NoThreadLocalScreenShotLaboratory` itself to Selenide?

asolntsev:
> Maybe a better option would be adding `NoThreadLocalScreenShotLaboratory` itself to Selenide? 

@mkurz hm... I started thinking that we should just replace current ScreenshotLabratory implementation by yours. It doesn't need to use ThreadLocals since version 5.0.0 :)

mkurz:
Sure, please use it. I just want to get rid of all the Thread locals (and static code) in it.

mkurz:
> But you are trying to make too many things public...

Can you explain why I am trying to make too many things public?


For example `AttachmentPrinter` and `Clock`. Why I am not allowed to instantiate them?

```java
new ScreenShotLaboratory(
            new WebdriverPhotographer(),
            new WebPageSourceExtractor(),
            new AttachmentPrinter(), // can not access
            new Clock() // can not access
    );
```

rosolko:
@asolntsev let's first open and then consider about changing implementation

asolntsev:
> Why I am not allowed to instantiate them?

It's just the default strategy: in a public project, try to keep as less public APIs as possible. 
Otherwise people tend to misuse them, and it's getting harder for us to refactor existing code.

But yes, in this case `Clock` and `AttachmentPrinter` and simple enough. Not a problem.

mkurz:
Thanks! That's all we need :+1: 
Any eta for the 7.10.2 release? Thank you!

asolntsev:
@mkurz Sure, we can release 7.10.2 at any moment. Unless you want to make few more changes, then we can wait for it. :)

mkurz:
No please release 7.10.2 now. If I run into other problems, in worst case, we can do 7.10.3 but I do not think this will be necessary.

Thank you very much!

asolntsev:
Yep, will release 7.11.0 (instead of 7.10.2) today.

asolntsev:
@mkurz Selenide 7.11.0 has been released. Thank you for the contribution!

mkurz:
Thank you!

PR #3167, opened 2025-11-06
Add an ability to get all request/response proxy filters

### LINKED ISSUE #3166: Add getters for requestFilters and responseFilters in SelenideProxyServer
### The problem

We reuse the browser across tests without closing, so we need to remove all proxy filters after each test. This is easily solved with hard-coded filter names, but it becomes problematic when a test requires multiple filters (since duplicate names are not allowed).

We could use hard-coded prefixed names with random suffixes, but the current implementation doesn't provide a way to retrieve all filters to identify and remove those matching our prefix pattern.

### Solution
Add two public getter methods to the SelenideProxyServer class for:
```
private final Map<String, RequestFilter> requestFilters = new HashMap<>();
private final Map<String, ResponseFilter> responseFilters = new HashMap<>();
```

rosolko:
## Proposed changes

Add methods to Selenide proxy:
```java
Map<String, RequestFilter> filters = proxy.requestFilters();
List<String> names = proxy.requestFilterNames();

Map<String, ResponseFilter> filters = proxy.responseFilters();
List<String> names = proxy.responseFilterNames();
```

This fixes #3166

## Checklist

### DIFF
--- a/src/test/java/com/codeborne/selenide/proxy/SelenideProxyServerTest.java
+++ b/src/test/java/com/codeborne/selenide/proxy/SelenideProxyServerTest.java
@@ -13,7 +13,11 @@ import java.util.List;
 import java.util.Map;
 
 import static org.assertj.core.api.Assertions.assertThat;
-import static org.mockito.Mockito.*;
+import static org.mockito.Mockito.any;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.never;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
 
 final class SelenideProxyServerTest {
   private final BrowserUpProxyServer bmp = mock();


asolntsev (changes requested):
@rosolko It would be great to add the real case:
1. add "dummy-1"
2. add "dummy-2"
3. add "other-3"
4. remove all filters starting from "dummy-"
5. verify that only "dummy-" have been removed.

rosolko:
@asolntsev This case we have to decide what kind of api should we provide for users.
As for now there are only several possible candidates and I'm not sure that we need any of them:
1. Remove all request/response filters
2. Remove request/response filters by predicate
3. Get a list with names of request/response filters (this method can be the first point for any user specific case, but will collect only stream.tolist logic)
4. Get a list with names of request/response filters by predicate

asolntsev:
Option 1 is not ok because Selenide has its own filters. Apparently, we shouldn't allow users do remove them. 
Options 2..4 are good. 
I guess we should pick an option that is convenient, but generic (not tightly related to one specific case). Probably #3 is the best option in that sense.

PR #3197, opened 2025-12-03
Fix frame rate for video

asolntsev:
fixes #3189

### DIFF
--- a/modules/video-recorder/src/main/java/org/selenide/videorecorder/core/VideoMerger.java
+++ b/modules/video-recorder/src/main/java/org/selenide/videorecorder/core/VideoMerger.java
@@ -49,7 +49,7 @@ class VideoMerger {
 
   private void generateVideo() throws IOException {
     FFmpegBuilder builder = new FFmpegBuilder()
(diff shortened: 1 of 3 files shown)


modules/video-recorder/src/main/java/org/selenide/videorecorder/core/VideoMerger.java
```diff
@@ -49,11 +49,10 @@ public String videoUrl() {
 
   private void generateVideo() throws IOException {
     FFmpegBuilder builder = new FFmpegBuilder()
-      .addExtraArgs("-framerate", String.valueOf(config.fps()))
```
rosolko: IMHO we still should give an ability to override fps by config

asolntsev: Yes, users still can set any FPS.
When consider it when taking screenshots (we try to take FPS screenshots per second).

But when generating the video, we look how many screenshots we actually could take. In case of slow browser or network, we probably took less than FPS screenshots per second. Then the video will just reflect the reality.

rosolko (approved):

PR #3232, opened 2026-01-13
Add an ability to pass proxy filters via config, and add them as global, even before open() method call

### LINKED ISSUE #2616: User can add a global proxy filter
## The problem
If I want to add a proxy filter (either RequestFilter or ResponseFilter), I can do it only per webdriver.
And only _after_ I've started webdriver.

So I need to write code like this:

```java
    open(); // starts webdriver
    getSelenideProxy().addRequestFilter("requestLogger", new RequestFilter() {
      @Override
      public HttpResponse filterRequest(HttpRequest request, HttpMessageContents contents, HttpMessageInfo messageInfo) {
        log.info(request.headers());
        return null;
      }
    });
    open("https://my.site.com");
```

## The problem
1. I need to write extra `open()` call (minor)
2. I need to find all places in my tests where I call `open(url)`, and add the beforementioned code to all of them. (this might be problematic)

## Solution
I need a method for adding "global" proxy filters that 
1. Can be called before I open a webdriver,
2. Will be automatically added to all future webdrivers (proxies).

rosolko:
## Proposed changes
Possible solution to resolve #2616

## Checklist

### DIFF
--- a/modules/proxy/src/test/java/integration/ProxyFiltersFromConfigurationTest.java
+++ b/modules/proxy/src/test/java/integration/ProxyFiltersFromConfigurationTest.java
@@ -1,0 +1,32 @@
+package integration;
+
+import com.codeborne.selenide.Configuration;
+import com.codeborne.selenide.RequestFilters;
+import com.codeborne.selenide.ResponseFilters;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+import java.util.List;
+
+import static com.codeborne.selenide.WebDriverRunner.getSelenideProxy;
+import static org.assertj.core.api.Assertions.assertThat;
+
+final class ProxyFiltersFromConfigurationTest extends ProxyIntegrationTest {
+  @BeforeEach
+  void setUp() {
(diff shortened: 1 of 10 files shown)


asolntsev:
I am in doubt. :) 

I feel that `Config` was previously holding only primitive values (constants): timeouts, folder names (long, int, String etc.).
Now it's holding the whole objects with some behaviour.

Also, these objects (request filters) usually hold some state, so they need to be per-thread (or per-browser). How would e handle it?

Third, I feel like "proxy" should be a separate module. Theoretically. 
In this case, proxy-specific settings (ideally) shouldn't be in core Config class.

But I don't have good suggestions yet... :(

rosolko:
@asolntsev Yes, I've also not sure about such changes. 
While implementing this changes, I was choosing between two implementation: 
1. Put ability inside configuration, like it was for browser capabilities, so it looks like a similar change.
2. Add an ability for SelenideProxy to set up a list of filers using static methods, but latest I've decided to skip such implementation because of unpredictable mutation of class inside threads.

rosolko:
Additionally, @asolntsev, we can designate this configuration as experimental, which may be eliminated at any time in a future release.

PR #3362, opened 2026-07-05
#2761 Make SelenideAppiumElement.$$ return SelenideAppiumCollection

### LINKED ISSUE #2761: Inconsistency with SelenideAppiumElement and $$ method returning SelenideAppiumCollection?
## The problem
I've encountered an issue with the SelenideAppiumElement class not behaving as expected when using the $$ method. It seems that SelenideAppiumElement inherits the $$ method from SelenideElement, but instead of returning a SelenideAppiumCollection, it returns an ElementsCollection. This creates a problem when trying to chain methods specifically designed for SelenideAppiumCollection.
Is there a recommended way to handle this situation? Do I need to extend or override something within the framework to get the desired functionality? Or is this something that needs to be fixed in the framework itself?
It's also possible that I might be misunderstanding something or using the framework incorrectly. Any guidance would be appreciated.

## Details
This does not compile because SelenideAppiumElement inherits the $$ method from SelenideElement, which returns an ElementsCollection instead of a SelenideAppiumCollection. Here is the relevant code snippet:
`public static SelenideAppiumCollection findElementsByLabel(String text, SelenideAppiumElement parentElement) {
    return parentElement.$$(By.xpath(String.format("..." + FIND_BY_LABEL, text)));
}
`

* **Selenide Version*7.3.2*:

svasenkov:
## Summary
- Override `findAll`, `$$`, and `$$x` on `SelenideAppiumElement` to return `SelenideAppiumCollection` instead of `ElementsCollection`
- Register Appium-specific `FindAll` commands so nested searches inside a parent element produce Appium collections and elements
- Make `SelenideAppiumCollection` extend `ElementsCollection` (required for covariant return types in Java) while preserving Appium element typing in `get()`, `first()`, `last()`, and chained collection operations

Fixes #2761

## Test plan
- [x] `./gradlew :modules:appium:check`
- [ ] Run Android integration test `AppiumCollectionsTest.appiumCollectionFromParentElement` on a device/emulator

Made with [Cursor](https://cursor.com)

## Summary by CodeRabbit

* **New Features**
  * Added Appium collection lookup methods, including CSS, XPath, and shorthand selectors.
  * Enabled collecting child elements from a parent element in Appium flows.
* **Bug Fixes**
  * Improved collection handling so returned Appium element groups support more fluent actions and searches.
* **Tests**
  * Added coverage for collecting elements from a parent Appium element.

### DIFF
--- a/modules/appium/src/main/java/com/codeborne/selenide/appium/SelenideAppiumCollection.java
+++ b/modules/appium/src/main/java/com/codeborne/selenide/appium/SelenideAppiumCollection.java
@@ -1,28 +1,147 @@
 package com.codeborne.selenide.appium;
 
-import com.codeborne.selenide.BaseElementsCollection;
 import com.codeborne.selenide.Driver;
+import com.codeborne.selenide.ElementsCollection;
+import com.codeborne.selenide.WebElementCondition;
+import com.codeborne.selenide.WebElementsCondition;
+import com.google.errorprone.annotations.CanIgnoreReturnValue;
+import com.codeborne.selenide.impl.BySelectorCollection;
+import com.codeborne.selenide.impl.CollectionElement;
+import com.codeborne.selenide.impl.CollectionElementByCondition;
 import com.codeborne.selenide.impl.CollectionSource;
+import com.codeborne.selenide.impl.LastCollectionElement;
+import com.codeborne.selenide.impl.WebElementSource;
+import com.codeborne.selenide.impl.WebElementsCollectionWrapper;
 import org.openqa.selenium.By;
 import org.openqa.selenium.WebElement;
 
+import java.time.Duration;
 import java.util.Collection;
 
-public class SelenideAppiumCollection extends BaseElementsCollection<SelenideAppiumElement, SelenideAppiumCollection> {
+public class SelenideAppiumCollection extends ElementsCollection {
+  private final CollectionSource source;
+
   SelenideAppiumCollection(CollectionSource collection) {
     super(collection);
+    this.source = collection;
   }
 
   SelenideAppiumCollection(Driver driver, Collection<? extends WebElement> elements) {
-    super(driver, elements);
+    this(new WebElementsCollectionWrapper(driver, elements));
   }
 
   SelenideAppiumCollection(Driver driver, By seleniumSelector) {
-    super(driver, seleniumSelector);
+    this(new BySelectorCollection(driver, seleniumSelector));
+  }
+
+  public SelenideAppiumCollection(WebElementSource parent, By selector) {
+    this(new BySelectorCollection(parent.driver(), parent, selector));
   }
 
   @Override
   protected SelenideAppiumCollection create(CollectionSource source) {
     return new SelenideAppiumCollection(source);
+  }
+
+  @Override
+  @CanIgnoreReturnValue
+  public SelenideAppiumCollection should(WebElementsCondition... conditions) {
+    return (SelenideAppiumCollection) super.should(conditions);
+  }
+
+  @Override
+  @CanIgnoreReturnValue
+  public SelenideAppiumCollection should(WebElementsCondition condition, Duration timeout) {
+    return (SelenideAppiumCollection) super.should(condition, timeout);
+  }
+
+  @Override
+  @CanIgnoreReturnValue
+  public SelenideAppiumCollection shouldBe(WebElementsCondition... conditions) {
+    return (SelenideAppiumCollection) super.shouldBe(conditions);
+  }
+
+  @Override
+  @CanIgnoreReturnValue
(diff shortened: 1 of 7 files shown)


asolntsev:
I feel that it's wrong that `SelenideAppiumCollection extends ElementsCollection`. Because `ElementsCollection` contains `SelenideElement` instances, not `SelenideAppiumElement`.

In that sense, the previous `SelenideAppiumCollection extends BaseElementsCollection<SelenideAppiumElement, SelenideAppiumCollection>` was logically correct.

**I asked Claude to analyze it:**

Your suspicion is correct, and it's not just aesthetic — it produces a real bug.

SelenideElement.findAll()/$$()/$$x() in core (unchanged by this PR) are declared to return the concrete class ElementsCollection, not a generic bound. Java's covariant-return-type rule then forces any override to return a subtype of ElementsCollection — so extends BaseElementsCollection<SelenideAppiumElement, SelenideAppiumCollection> (your proposed, and the pre-PR, design) simply won't compile against the current core interface. That's a real language constraint, not an arbitrary choice by the PR author.

But ElementsCollection hardcodes its element type to SelenideElement (ElementsCollection extends BaseElementsCollection<SelenideElement, ElementsCollection>), and that fixed type leaks through a private clazz field used internally for iteration. The PR overrides get()/first()/last()/find()/findBy() to fix this for direct element access, but misses iterator(), stream(), asFixedIterable(), and asDynamicIterable() — verified below.

Findings (most severe first)

1. modules/appium/.../SelenideAppiumCollection.java (whole file) — iterating a collection yields SelenideElement-only proxies, not SelenideAppiumElement. Confirmed: iterator()/stream()/asFixedIterable()/asDynamicIterable() are inherited from BaseElementsCollection, which builds each element via CollectionElement.wrap(clazz, ...) using a clazz field fixed to SelenideElement.class (set when ElementsCollection's constructor runs super(collection) with T=SelenideElement). CollectionElement.wrap creates a JDK dynamic proxy implementing only the interfaces passed to it — so SelenideAppiumCollection items = el.$$(...); items.stream().forEach(e -> ((SelenideAppiumElement) e).tap()); throws ClassCastException, even though items.first() or items.get(0) correctly return real SelenideAppiumElement proxies. This inconsistency is the direct cost of the extends ElementsCollection workaround, and it's not covered by the PR's new test (which only exercises .first()).
2. src/main/java/com/codeborne/selenide/SelenideElement.java (out of this PR's scope) — the actual fix belongs in core. The right long-term design is what your instinct points at: findAll/$$/$$x should return a generic bound (e.g. BaseElementsCollection<?, ?>) rather than the concrete ElementsCollection class, which would let SelenideAppiumCollection extend BaseElementsCollection<SelenideAppiumElement, SelenideAppiumCollection> directly with no overrides at all — and incidentally, your own uncommitted local change to that file is already this exact fix. It's a legitimate breaking API change though (anything doing ElementsCollection ec = element.$$(...) would need to change), so it's reasonable it wasn't bundled into this bugfix PR, but the tradeoff should probably be called out in the PR description rather than silently worked around.
3. modules/appium/.../SelenideAppiumCollection.java:29-33 — new private final CollectionSource source field duplicates the inherited-but-private collection field in BaseElementsCollection. should*/filter/exclude/snapshot operate through the inherited field via super calls, while get/first/last/find/findBy use the new source field — same object today (both set together in every constructor), but two names for one thing is a maintenance trap, and would be unnecessary if BaseElementsCollection exposed a protected accessor instead.
4. modules/appium/.../SelenideAppiumElement.java —  untested. The new integration test only exercisesloginForm.$$(By...) → .first(). findAll(String), findAll(By), $$(String), $$x(String), and the overridden                          filter/filterBy/exclude/excludeWith/snapshot on Selro coverage — per CLAUDE.md, "Write tests for newcode."                                                                                                                             5. modules/appium/.../SelenideAppiumPageFactory.jav longer extends SelenideAppiumCollection, so it nowindependently re-implements the same "collection of SelenideAppiumElement" wiring instead of reusing it. I checked for a concrete  break (e.g. instanceof SelenideAppiumCollection on d none — createCollection() already returns thegeneric wildcard BaseElementsCollection<?,?> — so this is a maintainability cost, not a live bug, but it's a second sign the class hierarchy doesn't fit cleanly anymore.

Minor/lower-priority (not detailed further): filterr-call-plus-cast pattern while the sibling findBy()self-delegates to find() — inconsistent style within the same file; and the new public SelenideAppiumCollection(WebElementSource, By) constructor diverges slightly from core FindAll.jav.

@svasenkov

asolntsev:
Superseded by https://github.com/selenide/selenide/pull/3373

(PR #3373 implements the same approach, but fixes issues described in https://github.com/selenide/selenide/pull/3362#issuecomment-4926654638)

PR #2138, opened 2023-01-23
#1741 Experimental visible text condition

aakachurin:
## Proposed changes
For #1741
I don't know how it works with react or another framework
## Checklist
- [x] Checkstyle and unit tests are passed locally with my changes by running `gradlew check chrome_headless firefox_headless` command
- [x] I have added tests that prove my fix is effective or that my feature works
- [ ] I have added necessary documentation (if appropriate)

PR #3366, opened 2026-07-06
#1741 Add visibleText() and exactVisibleText() conditions

### LINKED ISSUE #1741: Method `$.shouldHave(text(...))` should take into account only a visible text
## The problem
When a div contains a long text, but only a part of this text is visible (due to specific CSS), user de-facto can see only  the part of the text - and this is a bug. We want to verify that user can see the whole text. 

## Details
Given the following HTML:
```html
<div id="partial" style="border: 1px solid blue; display: block; height: 19px; max-width: 152.5px; overflow: hidden; text-overflow: ellipsis;">
  <span id="field_value">987 654 321 100 100.876543321 AUTOOQODS</span>
</div>
```

The test should fail, but it works:
```java
    $("#partial").shouldHave(exactText("987 654 321 100 100.876543321 AUTOOQODS"));
```

* **Selenide Version**: 6.3.3
* **Chrome\Firefox\IE Version**: Chrome 98, Firefox 97

svasenkov:
## Summary
- Add `Condition.visibleText()` — partial match on text that fits within the element's visible layout
- Add `Condition.exactVisibleText()` — exact match on visibly rendered text
- Both respect CSS that hides overflowing text (`overflow: hidden`, `text-overflow: ellipsis`, etc.)
- Does **not** change behavior of existing `text()` / `exactText()` conditions

## Issue
Closes #1741

## Approach
`innerText` and Range API do not reflect CSS ellipsis clipping in Chromium. The JS helper clones the element without width constraints, compares full vs. available width, then binary-searches the longest text prefix that still fits.

Shared reading logic lives in `GetVisibleText` (mirrors `GetOwnText`).

## Test plan
- [x] `./gradlew :modules:core:check`
- [x] `./gradlew :modules:core:chrome_headless --tests integration.VisibleTextTest`
- [x] `./gradlew :modules:core:test --tests com.codeborne.selenide.conditions.VisibleTextTest`
- [x] `./gradlew :modules:core:test --tests com.codeborne.selenide.conditions.ExactVisibleTextTest`

## Notes for reviewer
- Previous experimental PR #2138 used Range API on `firstChild` only; this version handles nested text nodes and ellipsis via layout measurement
- Changing default `text()` behavior is intentionally out of scope

## Summary by CodeRabbit

* **New Features**
  * Added `visibleText()` and `exactVisibleText()` conditions for checking text that is actually visible when content is truncated or overflow-hidden.
  * Improved text handling for partially visible elements while still supporting fully visible content and select elements.
* **Tests**
  * Added coverage for partial and full visibility scenarios, including case-sensitive and case-insensitive matching behavior.
* **Documentation**
  * Updated the changelog for the new release with the new text conditions.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Changelog
 
 ## 7.17.0 (TBD)
+* #1741 added `visibleText()` condition for overflow-truncated elements
 * update Selenium from 4.44.0 to 4.45.0 (incl. CDP from 148 -> 149)
 
 ## 7.16.2 (27.05.2026)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 # Changelog
 
 ## 7.17.0 (TBD)
-* #1741 added `visibleText()` condition for overflow-truncated elements
+* #1741 added `visibleText()` and `exactVisibleText()` conditions for overflow-truncated elements
 * update Selenium from 4.44.0 to 4.45.0 (incl. CDP from 148 -> 149)
 
 ## 7.16.2 (27.05.2026)
--- a/src/main/java/com/codeborne/selenide/Condition.java
+++ b/src/main/java/com/codeborne/selenide/Condition.java
@@ -26,6 +26,7 @@ import com.codeborne.selenide.conditions.Hidden;
 import com.codeborne.selenide.conditions.Href;
 import com.codeborne.selenide.conditions.InnerText;
 import com.codeborne.selenide.conditions.VisibleText;
+import com.codeborne.selenide.conditions.ExactVisibleText;
 import com.codeborne.selenide.conditions.Interactable;
 import com.codeborne.selenide.conditions.IsImageLoaded;
 import com.codeborne.selenide.conditions.MatchAttributeWithValue;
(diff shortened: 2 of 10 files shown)


asolntsev:
## Code review

Reviewed the diff (`GetVisibleText`, `VisibleText`/`ExactVisibleText` conditions, `visible-text.js`, tests). Two confirmed correctness bugs should be fixed before merging:

1. **Ellipsis glyph width isn't accounted for** (`visible-text.js`, in the binary search around the substring-fits check). The candidate substring is measured against the clipped box width, but no width is reserved for the actual `…` character a real `text-overflow: ellipsis` element renders. This means `exactVisibleText(...)` can report more characters as visible than are actually shown on screen.

2. **Measurement clone loses ancestor-scoped CSS** (`visible-text.js`, where the clone is created and appended). The clone is appended directly to `document.body` rather than in place, so any styling applied via ancestor/descendant selectors (e.g. `.card .title { font-size: 20px }`) is lost, and the clone falls back to `<body>`-level defaults — producing an incorrect width measurement. This is masked in the current tests because they style elements inline.

Additional lower-priority notes:

- **RTL / start-anchored ellipsis isn't handled.** The algorithm assumes the visible text is always a *prefix* of the full DOM text (`fullText.substring(0, ...)`). For `dir="rtl"` content where truncation happens from the left, this returns exactly the hidden portion instead of the visible one. Might be worth a test case or at least a documented limitation.
- **Dead code**: `GetVisibleText` implements `Command<String>` just like `GetText`/`GetInnerText`/`GetOwnText`, but unlike those it's never registered in `Commands.java` — only the static `getVisibleText(...)` helper is actually called by the new conditions. The `Command` implementation looks unreachable/unnecessary.
- **Perf**: each condition check does a binary search that mutates `clone.textContent` and calls `getBoundingClientRect()` per iteration (~9-12 forced synchronous reflows per check), on top of a DOM clone/append/remove — worth being aware of given Selenide's polling/retry loop, though not blocking.

Happy to take a look again once these are addressed.

PR #3372, opened 2026-07-08
Selenide MCP: added capabilities set support

sidelnikovmike:
## Proposed changes
In previous version there was a problem with set of capabilities. Due to this problem - there can be problem with start browser: for example new chrome need --no-sandbox in google capability.

Now there is 2 changes:

1. --caps parameter moved to --tools, because previous implementation of this parameter was not about browser capabilities, more about code tools connection
2. added --cap parameters to specify browser capabilities. It now implemented as multi parameter with specific format `name=value`

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Updated configuration options to use separate flags for enabling tools and setting browser capabilities.
  * Added support for enabling the `codegen` tool via `--tools=<list>`.
  * Added support for custom browser capabilities via repeated `--cap=<name>=<value>` entries.

* **Bug Fixes**
  * Improved parsing to correctly preserve capability values that include `=`.
  * Added validation that errors when `--cap` entries don’t include a valid `name=value` pair.

* **Documentation**
  * Refreshed configuration documentation to reflect the new `Tools` and `Browser Capabilities` flags.

* **Tests**
  * Updated and expanded test coverage for tool and capability flag parsing.

### DIFF
--- a/modules/mcp/README.md
+++ b/modules/mcp/README.md
@@ -104,14 +104,29 @@ Add to `.claude/settings.json` (project) or `~/.claude/settings.json` (global):
 | `--selector-mode=<mode>` | Selector mode: `CSS`, `Sizzle` | `CSS` |
 | `--assertion-mode=<mode>` | Assertion mode: `STRICT`, `SOFT` | `STRICT` |
 
-### Capabilities
+### Tools
 
 | Parameter | Description |
 |---|---|
-| `--caps=<list>` | Comma-separated list of capabilities to enable |
+| `--tools=<list>` | Comma-separated list of tools to enable |
 
-Available capabilities:
+Available tools:
 
 - `codegen` — enables test code generation tools
 
-Example: `--caps=codegen`
+Example: `--tools=codegen`
+
+### Browser Capabilities
+
+| Parameter | Description |
+|---|---|
+| `--cap=<name>=<value>` | Add a custom browser capability |
+
+You can pass multiple `--cap=` parameters to add several capabilities.
+
+Examples:
+```bash
+--cap=browserName=chrome
+--cap=selenoid:options={"enableVNC":true}
+--cap=custom:option=value
+```
--- a/modules/mcp/src/main/java/com/codeborne/selenide/mcp/SelenideMcpServer.java
+++ b/modules/mcp/src/main/java/com/codeborne/selenide/mcp/SelenideMcpServer.java
@@ -17,6 +17,7 @@ import io.modelcontextprotocol.server.McpSyncServer;
 import io.modelcontextprotocol.server.transport.StdioServerTransportProvider;
 
 import java.util.Arrays;
+import java.util.List;
 import java.util.concurrent.CountDownLatch;
 
 /**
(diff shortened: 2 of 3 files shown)


asolntsev:
@sidelnikovmike Review by Jean Claude AI Damme:

### Overview

The PR touches the modules/mcp CLI arg parsing (SelenideMcpServer.java), its tests, and README. Two changes:
1. Renames the existing --caps=<list> flag (which enables MCP tool groups like codegen) to --tools=<list>.
2. Adds a new --cap=name=value flag (repeatable) to set arbitrary WebDriver browser capabilities, motivated by needing e.g. --no-sandbox for new Chrome versions.

CI is green (all checks pass), and tests were added/renamed to match.

### Critical issue: doesn't actually solve the motivating problem

The PR description's own example — "new chrome need --no-sandbox in google capability" — requires setting a value nested inside the goog:chromeOptions capability ({"goog:chromeOptions": {"args": ["--no-sandbox"]}}). But applyCapability only supports flat name=value pairs via MutableCapabilities.setCapability(String, String):

config.browserCapabilities().setCapability(cap[0], cap[1]);

There's no way to express --cap=goog:chromeOptions=--no-sandbox and get a nested args array out of it — it would just set the literal string "--no-sandbox" as the whole goog:chromeOptions value, which Chrome/Selenium won't interpret as an argument list. So the exact use case cited as the reason for this PR isn't actually achievable with the implementation as written.

The same limitation affects the README's own second example: --cap=selenoid:options={"enableVNC":true} sets selenoid:options to the string {"enableVNC":true}, not a JSON object/map. Selenoid expects selenoid:options to be a nested capability object, not a string — so this documented example is likely non-functional against a real Selenoid grid. Worth verifying against an actual Selenoid instance before merging, since it's presented as a working example.

### Backward-compatibility break (silent, undocumented)

--caps=<list> has been a public, documented flag since the MCP module first shipped (present in the v7.16.2 release tag, and earlier). This PR renames it to --tools=<list> with no fallback or deprecation warning. Anyone with --caps=codegen in their .claude/settings.json will silently stop getting the codegen tools after upgrading — no error, just missing functionality, which is the worst kind of breaking change (silent). At minimum this should be called out in release notes; ideally --caps= could still be accepted (with a warning) for one release cycle.

### Other observations

- The split("=", 2) fix (limiting to 2 parts) is a real, correct fix for capability values that themselves contain = (e.g. base64-ish or JSON values) — good.
- Only string values are supported (setCapability(String, String)); there's no way to set boolean/numeric capabilities in their native type (e.g. acceptInsecureCerts=true becomes the string "true", not the boolean). Some drivers are strict about capability types — worth at least documenting this constraint.
- Test coverage for the new --cap parsing is solid (basic, =-in-value, multiple caps, missing-value error).
- Minor style nit: applyCapability breaks the naming convention of sibling methods (applyBrowserArg, applyConnectionArg, applyProxyArg, applyBehaviorArg, applyModeArg) — should probably be applyCapabilityArg for consistency.
- README updates are clear and match the new flags.

### Verdict

Reasonable direction (browser capabilities are a legitimate gap), but I'd not merge as-is:
- It doesn't solve the stated real-world problem (nested capabilities like goog:chromeOptions.args) — the flat name=value format is too limited for the motivating case.
- It silently breaks a previously-released, documented flag (--caps=codegen → --tools=codegen) with no migration path.

I'd ask the author to either support nested/JSON capability values (or clearly scope this PR to flat string capabilities only and open a follow-up for the nested case), and to address the --caps rename as a breaking change — either keep backward compatibility or flag it prominently in the changelog.

@sidelnikovmike

PR #3407, opened 2026-08-12
Fix/unwrap config in setWebDriver to prevent infinite recursion

Francisco-Gonzalezz:
## Proposed changes

Fixes a `StackOverflowError` that occurred after calling `WebDriverRunner.using(driver, proxy, folder, lambda)` when a previous `WebDriver` was already bound to the current thread.

**Root cause:** `setWebDriver(WebDriver, SelenideProxyServer, DownloadsFolder)` was storing the `ThreadLocalSelenideConfig` *wrapper* object as the `Config` inside the `WebDriverInstance`, instead of the concrete leaf config it delegates to. When `using()` finished and its `finally` block ran:

```java
config.set(prev.config());  // prev.config() was the wrapper itself
```

…the `ThreadLocal` slot ended up holding the wrapper pointing at itself. Every subsequent config read (`config.browser()`, etc.) entered an infinite delegation loop and threw `StackOverflowError`.

**Fix:** Pass `config.unwrap()` — the concrete leaf `Config` — when constructing `WebDriverInstance` inside `setWebDriver`. This guarantees `prev.config()` in the restore path is always a real leaf, never the wrapper.

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed a `StackOverflowError` that could occur when switching to a previously thread-bound WebDriver.
  * Improved reliability when switching between external and previously used browsers.
  * Preserved browser configuration during nested driver switches and after returning to a previous driver.
  * Ensured subsequent browser operations continue working correctly after multiple driver changes.
  * Improved stability when repeatedly switching between previously used browser sessions.


asolntsev:
@Francisco-Gonzalezz Can you please describe how the actual Selenide user can reproduce the problem?
In other words, can you write an integration test, not unit-test? For example, you can append a case to `CustomWebdriverTest`.

I've tried to use `using` inside of another `using` - but couldn't reproduce the problem:

```java
@Test
void userCanSwitchToCustomWebdriverAndBackToSelenideWebdriver() {
  openFile("page_with_big_divs.html");
  $("h1").shouldBe(visible).shouldHave(text("Some big divs"));

  using(browser1, () -> {
    openFile("page_with_selects_without_jquery.html");
    $("h1").shouldBe(visible).shouldHave(text("Page with selects"));

    using(browser2, () -> {
      openFile("file_upload_form.html");
      $("h1").shouldBe(visible).shouldHave(text("File upload form"));
    });

    assertThat(WebDriverRunner.getWebDriver()).isSameAs(browser1);
  });
}
```

Francisco-Gonzalezz:
Hi @asolntsev , apologies for not including this upfront — I've added using_canBeCalledAgain_afterPreviousUsingWithExternalDriver to CustomWebdriverTest.

The key precondition I missed explaining is that the bug requires setWebDriver() to be called first (i.e. the user is managing an externally created driver). The sequence is:

1. setWebDriver(browser1) — binds an external driver; setWebDriver was storing the ThreadLocalSelenideConfig wrapper instead of the unwrapped leaf config inside WebDriverInstance
2. using(browser2, lambda) — on unwind, config.set(prev.config()) fires where prev.config() returns the wrapper, making the ThreadLocal point at itself
3. Then trying to interact with the original browser bound to WebDriverRunner

This is why nesting using() inside using() doesn't trigger it — when Selenide creates its own driver, createDriver() already calls config.unwrap(), so the self-reference never forms.

asolntsev (approved):

asolntsev:
@Francisco-Gonzalezz Thank you for the explanation.

PR #3415, opened 2026-08-18
fix appium long press and double tap (#3413)

### LINKED ISSUE #3413: Appium long press tap should be on the center of the element
Is it developed on purpose? 
Regular tap is on the center of element:
https://github.com/selenide/selenide/blob/757d987dba5a3a6eac65da3b7fb3e0cd0822f17d/modules/appium/src/main/java/com/codeborne/selenide/appium/commands/AppiumClick.java#L92

 and long press is on the corner (location of the element)
https://github.com/selenide/selenide/blob/757d987dba5a3a6eac65da3b7fb3e0cd0822f17d/modules/appium/src/main/java/com/codeborne/selenide/appium/commands/AppiumClick.java#L79

I can fix long press, but decided to ask first. Maybe there is a reason.

qwez:
## Proposed changes

- fix https://github.com/selenide/selenide/issues/3413
- added offset option to long press and double tap

## Checklist
- [x] Tests work `gradlew :modules:appium:android --tests "it.mobile.android.AndroidClickOptionsTest"`

## Summary by CodeRabbit

- **New Features**
  - Added support for offset-based double-tap and long-press gestures.
  - Mobile tap, double-tap, and long-press actions now use the element center as the default location.
  - Double-tap gestures now support custom horizontal and vertical offsets.

- **Bug Fixes**
  - Improved consistency between standard and offset-based tap actions while preserving existing default behavior.

### DIFF
--- a/modules/appium/src/main/java/com/codeborne/selenide/appium/AppiumClickOptions.java
+++ b/modules/appium/src/main/java/com/codeborne/selenide/appium/AppiumClickOptions.java
@@ -32,15 +32,23 @@ public class AppiumClickOptions extends ClickOptions {
   }
 
   public static AppiumClickOptions doubleTap() {
-    return new AppiumClickOptions(AppiumClickMethod.DOUBLE_TAP, 0, 0, ZERO);
+    return doubleTapWithOffset(0, 0);
+  }
+
+  public static AppiumClickOptions doubleTapWithOffset(int xOffset, int yOffset) {
+    return new AppiumClickOptions(AppiumClickMethod.DOUBLE_TAP, xOffset, yOffset, ZERO);
   }
 
   public static AppiumClickOptions longPress() {
-    return new AppiumClickOptions(AppiumClickMethod.LONG_PRESS, 0, 0, DEFAULT_LONG_PRESS_DURATION);
+    return longPressFor(DEFAULT_LONG_PRESS_DURATION);
   }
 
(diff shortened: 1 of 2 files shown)


asolntsev (changes requested):
Thanks for the fix — using `getCenter(webElement)` in `performDoubleTap` and `performLongPress` is the right change and confirms the bug also affected double-tap, not just long-press.

A few things I'd like addressed before merging:

1. **Tests are missing.** The checklist has "I have added tests that prove my fix is effective" unchecked. `AndroidClickOptionsTest` already has `androidLongPress`/`androidDoubleTap`, but nothing covers the new `doubleTapWithOffset` / `longPressWithOffsetFor` methods, and nothing regression-tests the center-vs-corner fix itself (the existing tests likely passed even with the bug, since the tapped elements are probably large enough to tolerate a corner tap). Please add integration tests for the new offset variants, ideally against a small enough element that a corner-tap would visibly miss.
2. **README.md not updated.** `modules/appium/README.md` documents the click options and was updated when `longPressFor` was originally added (see `b3a1a46c`). Please document `doubleTapWithOffset` and `longPressWithOffsetFor` there too.
3. **Minor naming nit:** the local variable is now named `location`, but after the fix it holds the element's *center* point, not its location. `center` would be more accurate — this was actually the same naming confusion (`size` used for a location) that prompted the original bug report, so let's get it right this time.

The core bug fix itself looks correct — just want tests + docs before merging.

PR #3416, opened 2026-08-18
#3413 Add tests and docs for appium tap-offset fix

### LINKED ISSUE #3413: Appium long press tap should be on the center of the element

asolntsev:
## Proposed changes

Follow-up to #3415 (which fixed #3413 by making `performDoubleTap`/`performLongPress` use the element's center instead of its top-left corner). That PR was merged without the checklist items for tests and docs, so this PR fills those in:

- Add `androidDoubleTapWithOffset` / `androidLongPressWithOffset` integration tests. Each computes an offset close to the target element's edge, so the tap only lands on the element if it's based on the element's **center** — with the old corner-based bug, the same offset would miss the element entirely. This is a regression test for the #3413 fix itself, not just the new offset methods.
- Document `doubleTapWithOffset` and `longPressWithOffsetFor` in `modules/appium/README.md`, matching how `tapWithOffset` was already documented.
- Rename the local variable `location` (holding the element's center point since #3415) to `center` in `AppiumClick`, to avoid the same naming confusion that partly motivated the original bug report.

## Checklist
- [x] Checkstyle and compilation pass locally (`./gradlew :modules:appium:checkstyleMain :modules:appium:checkstyleTest :modules:appium:compileTestJava`)
- [ ] New integration tests require a running Android emulator to execute — not run in this environment, please verify with `./gradlew :modules:appium:android --tests "it.mobile.android.AndroidClickOptionsTest"` before merging
- [x] Documentation updated (README)

## Summary by CodeRabbit

* **New Features**
  * Added examples for double-tap and long-press actions with positional offsets in the Appium documentation.
  * Added coverage for offset-based Android interactions, including long-press and double-tap actions.

* **Tests**
  * Verified that offset interactions produce the expected results.

### DIFF
--- a/modules/appium/README.md
+++ b/modules/appium/README.md
@@ -72,9 +72,11 @@ SelenideAppium.openIOSDeepLink("mydemoapprn://product-details/1");
    $(AppiumBy.xpath(".//*[@text='Views']")).tap(); //perform native event tap
    $(AppiumBy.xpath(".//*[@text='Views']")).doubleTap(); //perform native event double tap
    $(AppiumBy.xpath(".//*[@text='Views']")).click(tapWithOffset(100, -60)) //perform tap with offset from center of the element
+   $(AppiumBy.xpath(".//*[@text='Views']")).click(doubleTapWithOffset(100, -60)) //perform double tap with offset from center of the element
    $(AppiumBy.xpath(".//*[@text='People Names']")).click(longPress());
    $(AppiumBy.xpath(".//*[@text='People Names']")).click(longPressFor(ofSeconds(5)));
    $(AppiumBy.xpath(".//*[@text='People Names']")).tap(longPressFor(ofSeconds(4)));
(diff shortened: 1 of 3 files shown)


qwez:
Thank you for finishing this up. I’ll be more mindful of tests and documentation next time.

PR #3439, opened 2026-09-19
#3438 allow SelenideAppiumCollection subclasses to specify their element class

### LINKED ISSUE #3438: Allow subclasses of SelenideAppiumCollection to specify their own element class
### Use case

We maintain a mobile e2e framework layer on top of selenide-appium. It declares its own element interface:

```java
public interface MyElement extends SelenideAppiumElement {
  MyElement longPressCenter();
  // … project-specific commands, registered via Commands.getInstance().add(...)
}
```

Since 7.17.0 (#2761, PR #3362) `SelenideAppiumElement.$$` / `findAll` return `SelenideAppiumCollection`, which is great. To make them return a collection of `MyElement`, our collection has to be a subclass of `SelenideAppiumCollection` — otherwise a covariant override of `$$` does not compile.

### What blocks it

`SelenideAppiumCollection` has no constructor a subclass can use:

```java
SelenideAppiumCollection(CollectionSource collection)                    // package-private
SelenideAppiumCollection(Driver, Collection<? extends WebElement>)       // package-private
SelenideAppiumCollection(Driver, By)                                     // package-private
public SelenideAppiumCollection(WebElementSource parent, By selector)    // public, but hardcodes
                                                                         // SelenideAppiumElement
```

Two consequences:

1. there is no way to pass a different element class;
2. without a `CollectionSource` constructor a subclass cannot override `create(CollectionSource)`, so `filter()`, `first(int)`, `snapshot()` and friends cannot keep the subclass type either.

### Suggested fix

`ElementsCollection` already solves this one level up — 7.17.0 added:

```java
/**
 * Lets a subclass produce elements of a more specific class than {@link SelenideElement},
 * e.g. a plugin module's own element interface …
 */
protected ElementsCollection(CollectionSource collection, Class<? extends SelenideElement> elementClass)
```

The same constructor in `SelenideAppiumCollection` would be enough:

```java
protected SelenideAppiumCollection(CollectionSource collection,
                                   Class<? extends SelenideAppiumElement> elementClass) {
  super(collection, elementClass);
}
```

Purely additive: no existing constructor, signature or behaviour changes.

**Note on the bound** (updated after review feedback on PR #3439, commit `ea451a9`): the parameter is
`Class<? extends SelenideAppiumElement>`, deliberately narrower than the `Class<? extends SelenideElement>`
of the `ElementsCollection` counterpart. Mirroring the parent exactly would compile but is unsound here,
because `SelenideAppiumCollection.get()` / `first()` / `last()` / `find()` cast to `SelenideAppiumElement`.
A subclass passing `SelenideElement.class` compiles and then fails at runtime:

```
ClassCastException: class jdk.proxy2.$Proxy57 cannot be cast to
class com.codeborne.selenide.appium.SelenideAppiumElement
```

With the narrower bound that same subclass no longer compiles, which is where the error belongs.

### Status

PR #3439 is open with the constructor, javadoc and a unit test covering both element type and
derived-collection type (`filter`, `first(int)`, `snapshot`).

aquahitt:
## Proposed changes

Fixes #3438.

Adds a `protected` constructor to `SelenideAppiumCollection` mirroring `ElementsCollection(CollectionSource, Class<? extends SelenideElement>)`, which was added in 7.17.0 for exactly the same purpose one level up in the hierarchy.

This lets a downstream framework layer declare its own element interface extending `SelenideAppiumElement` together with a matching collection, so that a covariant override of `$$` / `findAll` can return it — the natural continuation of #2761 / #3362, which made those methods return `SelenideAppiumElement`s.

The change is purely additive: no existing constructor, signature or behaviour is modified.

The added test covers both halves of the problem: that elements come back as the subclass' own element interface, and that derived collections (`first(int)`, `snapshot()`) keep the subclass — the latter is only possible because the new constructor also makes `create(CollectionSource)` overridable.

## Checklist

  I could not run the Gradle build itself: in my environment the JVM has no outbound network access (Gradle fails with `Read timed out` on `repo.maven.apache.org`), and there is no browser available. I verified the change by hand instead:

  - compiled the changed class and the new test against the published 7.18.2 artifacts (`selenide-core`, `selenide`, `selenide-appium`) with Selenium 4.49.0 and java-client 10.1.1 — OK;
  - ran the new test on the JUnit Platform Console with the patched class shadowing the released one — **2 passed, 0 failed**;
  - ran Checkstyle 10.23.0 (the `toolVersion` from `gradle/code-analysis.gradle`) with this repository's `config/checkstyle/checkstyle.xml` over both files — **no violations**.

  SpotBugs and the browser suites were not run. Happy to adjust if CI finds anything.

- [x] I have added tests that prove my fix is effective or that my feature works (`SelenideAppiumCollectionSubclassTest`)
- [x] I have added necessary documentation (javadoc on the new constructor, pointing at its `ElementsCollection` counterpart)

I did not touch `CHANGELOG.md`: its entries reference PR numbers and appear to be written at release time, and there is no "unreleased" section. Happy to add a line if you prefer.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved type safety for Appium collection subclasses, preventing incompatible element types from being used and avoiding potential runtime casting errors.
  * Custom element and collection types are now preserved consistently across item access, filtering, subsets, and snapshots.

* **Tests**
  * Added coverage for custom element types returned by indexed, first, and last item access.
  * Added coverage confirming custom collection types are preserved in derived operations.


aquahitt:
Thanks — both review comments addressed in ea451a9.

**1. `elementClass` bound — accepted, this was a real hole.**

I had mirrored the parent's signature (`Class<? extends SelenideElement>`), but the parent's type parameter *is* `SelenideElement`, while here `get()` / `first()` / `last()` / `find()` cast to `SelenideAppiumElement` — so mirroring was wrong. Reproduced it before fixing: a subclass passing `SelenideElement.class` compiled fine and then failed at runtime with

```
ClassCastException: class jdk.proxy2.$Proxy57 cannot be cast to
class com.codeborne.selenide.appium.SelenideAppiumElement
```

With `Class<? extends SelenideAppiumElement>` that same subclass no longer compiles, which is where the error belongs. The now-unused `SelenideElement` import is removed as well.

**2. `filter()` assertion — added.**

For the record, `filter()`, `first(int)` and `snapshot()` all funnel through the same `create(CollectionSource)`, so the existing assertions already covered the mechanism; the new one only guards against `filter()` being reimplemented around `create()`. It is one line and `filter()` is the most used of the three in real code, so it seems worth having.

Verification after the change (same manual setup as described in the PR body, since Gradle has no network access in my environment): compiles, `SelenideAppiumCollectionSubclassTest` passes 2/2, Checkstyle 10.23.0 with this repository's config reports no violations.

Issue #3440, opened 2026-09-19
SelenideAppiumElement loses its type on 38 of 56 chainable methods


aquahitt:
### Problem

`#2761` / PR `#3362` made `find`, `$`, `$x`, `findAll`, `$$`, `$$x` return Appium types. That fixed
navigation, but the same type loss remains on most other chainable methods, so this still does not
compile:

```java
SelenideAppium.$(By.id("amount")).setValue("100").tap();
//                                              ^ setValue returns SelenideElement, tap() is not there
```

`setValue` here **is** executed by `AppiumSetValue` — the behaviour is Appium-specific, only the
declared return type is not.

### Numbers

Counted against the `v7.18.2` sources:

| | |
|---|---|
| methods on `SelenideElement` returning `SelenideElement` | 56 |
| covariantly redeclared in `SelenideAppiumElement` | 18 |
| **losing the Appium type** | **38** |

Of those 38, **12 already have an Appium-specific command** registered in `SelenideAppiumCommands`
(`click(ClickOptions)`, `doubleClick()`, `doubleClick(ClickOptions)`, `dragAndDrop`, `setValue` ×2,
`val` ×2, `type` ×2, `unfocus`, `scroll(ScrollOptions)`).

### The 38, split by what a fix needs

**Group A — the command returns the same proxy, so a covariant declaration is enough and no new
command is needed** (verified from each command's supertype: `FluentCommand`, or `Val` delegating to
`SetValue`, or `ScrollIntoCenter` delegating to `scrollIntoView`):

`setValue(String)`, `setValue(SetValueOptions)`, `val(String)`, `val(SetValueOptions)`,
`type(CharSequence)`, `type(TypeOptions)`, `append(String)`, `press(CharSequence...)`, `pressEnter()`,
`pressEscape()`, `pressTab()`, `hover()`, `hover(HoverOptions)`, `click(ClickOptions)`,
`doubleClick()`, `doubleClick(ClickOptions)`, `contextClick()`, `dragAndDrop(DragAndDropOptions)`,
`scroll(ScrollOptions)`, `scrollIntoView(String)`, `scrollIntoView(boolean)`,
`scrollIntoView(ScrollIntoViewOptions)`, `scrollIntoCenter()`, `setSelected(boolean)`, `paste()`,
`unfocus()`, `highlight()`, `highlight(HighlightOptions)`, `unhighlight()` — 29 methods.

**Group B — the command builds a *new* element, so a covariant declaration alone would throw
[...]

PR #3441, opened 2026-09-19
#3440 return SelenideAppiumElement from chainable methods that keep the same proxy

aquahitt:
## Proposed changes

Part of #3440 — the first of the two groups described there.

(Deliberately not `Fixes`: this PR covers only the methods that need no new command, so the
issue should stay open until the second group is decided.)

`#2761` / `#3362` made `find`, `$`, `$x`, `findAll`, `$$`, `$$x` return Appium types. The other
chainable methods still return `SelenideElement`, so this does not compile even though `setValue` is
executed by `AppiumSetValue`:

```java
SelenideAppium.$(By.id("amount")).setValue("100").tap();
//                                              ^ setValue returns SelenideElement
```

This PR adds covariant declarations for the **27** chainable methods whose command returns the
**same proxy**, so no new command and no behaviour change is involved:

`setValue` ×2, `val` ×2, `type` ×2, `append`, `paste`, `press`, `pressEnter`, `pressEscape`,
`pressTab`, `unfocus`, `setSelected`, `click(ClickOptions)`, `contextClick`, `doubleClick` ×2,
`hover` ×2, `dragAndDrop`, `scroll(ScrollOptions)`, `scrollIntoView(ScrollIntoViewOptions)`,
`scrollIntoCenter`, `highlight` ×2, `unhighlight`.

Each one was classified by reading its command's supertype: `FluentCommand` returns `proxy`
directly; `Val` delegates to `SetValue`; `ScrollIntoCenter` delegates to `scrollIntoView`. Twelve of
them already run Appium-specific implementations (`AppiumSetValue`, `AppiumType`, `AppiumVal`,
`AppiumClick`, `AppiumDoubleClick`, `AppiumDragAndDrop`, `AppiumUnfocus`, `AppiumScrollTo`) — only
the declared type was missing.

**Deliberately excluded:**

- `scrollIntoView(boolean)` and `scrollIntoView(String)` — both `@Deprecated` in `SelenideElement`;
  pulling them into the Appium API seemed wrong.
- The 9 methods whose command builds a *new* element — `parent`, `sibling`, `preceding`, `lastChild`,
  `ancestor` ×2, `closest`, `selectRadio`, `cached`. A covariant declaration alone would throw
  `ClassCastException` there, because the new proxy is created with `SelenideElement.class`. #3440
  describes the two ways to fix that and asks which you prefer; happy to send it as a follow-up.

## Checklist

  I could not run the Gradle build: in my environment the JVM has no outbound network access (Gradle
  fails with `Read timed out` on `repo.maven.apache.org`) and there is no browser. Verified by hand
  instead: compiled against the published 7.18.2 artifacts with `-Xlint:deprecation` (no warnings),
  ran the new test — **27 passed, 0 failed** — and ran Checkstyle 10.23.0 with this repository's
  `config/checkstyle/checkstyle.xml` over both files — **no violations**. SpotBugs and the browser
  suites were not run.

      (`SelenideAppiumElementReturnTypesTest` — a parameterized reflection test pinning the declared
      return type of every method this PR touches, so a future signature regression fails the build)
- [x] I have added necessary documentation (the declarations are `@Override`s; no new behaviour to document)

## Summary by CodeRabbit

* **New Features**
  * Appium element interactions now consistently return an Appium-specific element, enabling continued chaining across typing, clicking, hovering, scrolling, dragging, selection, and highlighting actions.
  * Added fluent support for interaction options and overloads, including nullable value handling.

* **Tests**
  * Added coverage verifying that supported Appium interaction methods preserve the expected chainable return type.


aquahitt:
On the review comment suggesting a covariant `click()` override — I checked it and it cannot be done,
so I have left the PR as is.

`SelenideElement.click()` returns **void**, not `SelenideElement`:

```java
// SelenideElement.java, v7.18.2
SelenideElement click(ClickOptions clickOption);   // line 1304
void click();                                      // line 1322
```

It comes from Selenium's `WebElement` (`public abstract void click()`), so there is nothing to make
covariant, and `$(...).click().tap()` cannot compile no matter what this interface declares — the
chain has no value to continue from. Declaring it is itself a compile error:

```
error: click() in ClickProbe clashes with click() in SelenideElement
  SelenideAppiumElement click();
                        ^
  return type SelenideAppiumElement is not compatible with void
```

The chainable click is `click(ClickOptions)`, which this PR does cover.

For what it is worth, the closest thing to the suggestion that *would* work is a new name returning
the element (`clickAndReturn()` or similar), but that is an API addition rather than a covariant
override, and it belongs to #3440 if you want it at all.

